PR #530, opened 2014-10-15
Sig capture: 

gpazo:


gilmoskowitz:
We've decided to go through signature capture in a different way entirely.

PR #531, opened 2014-10-15
re-add pocock's code from 10/13, with enhancements/holes for future impr...

gpazo:
...ovement


gilmoskowitz:
Recorded as issue #25664 in xTuple's issue tracker. We'll get back to this in a cleaner form and on the current dev branch in the near future.

PR #573, opened 2014-12-29
Issue #21298 Remove project type null text

andersdd:
pull into 4.9 branch


csvimp
```diff
@@ -1 +1 @@
-Subproject commit 4e0f9aecee9da19a449c448fc2054c9721aa627d
+Subproject commit 5d60c50d1da422cdb1edc3eb824f3f07a682cfa2
```
gilmoskowitz: We shouldn't update csvimp as part of this PR.

openrpt
```diff
@@ -1 +1 @@
-Subproject commit e897833e1ece76b29b507961586354a1426495c6
+Subproject commit 5130a709ae904acf09f6daa056552315dae867da
```
gilmoskowitz: We shouldn't update openrpt as part of this PR

PR #575, opened 2014-12-31
Issue # 25140 Copy Quote/Sales Orders to alternate Customer/Prospect

andersdd:
Copy quote and sales orders document(s) to an alternate customer or prospect.

Replaces several Sales Order lookups with a single display to assist searching for open/closed sales orders.

Also requires PR in the xtuple repo.


guiclient/quotes.cpp
```diff
@@ -171,11 +175,11 @@ void quotes::sConvert(int pType)
       QList<XTreeWidgetItem*> selected = list()->selectedItems();
       QList<XTreeWidgetItem*> notConverted;
 
-      for (int i = 0; i < selected.size(); i++)
+      foreach (XTreeWidgetItem *item, list()->selectedItems())
```
gilmoskowitz: Iterating over `list()->selectedItems()` here looks like it conflicts with `selected = notConverted` below (~line 310).

andersdd: What it appears to be doing is passing all items not copied (function returns an error) into the "selected" list after it processes the items (for the user to review ??)

gilmoskowitz:
Revision history indicates this construct is several years old, so if it's buggy then it's an old bug.
+1 when we merge https://github.com/xtuple/xtuple/pull/2101

PR #576, opened 2015-01-05
Add Copy fiscal year (and corresponding fiscal periods) 

andersdd:
Added to Fiscal Yea screen to facilitate setting up new Financial years based on the last year.

prerequisite xtuple repo. Db. changes

### DIFF
--- a/guiclient/accountingYearPeriods.cpp
+++ b/guiclient/accountingYearPeriods.cpp
@@ -26,7 +26,6 @@ accountingYearPeriods::accountingYearPeriods(QWidget* parent, const char* name, 
     setupUi(this);
 
     connect(_period, SIGNAL(populateMenu(QMenu*,QTreeWidgetItem*)), this, SLOT(sPopulateMenu(QMenu*,QTreeWidgetItem*)));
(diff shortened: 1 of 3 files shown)


guiclient/accountingYearPeriods.cpp
```diff
@@ -92,6 +96,11 @@ void accountingYearPeriods::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem *pSelect
   }
 }
 
+void accountingYearPeriods::sToggleCopy()
+{
+  _copyPeriod->setEnabled(_period->currentItem()->text(3) == "Yes");  
+}
```
gilmoskowitz: The logic here looks odd, untranslatable (unadorned `Yes`), and a maintenance problem (the hard-coded `3`).

Is there a way to create the UI without adding this extra column? Does it really matter which year gets copied, as long as the resulting FY follows the latest that exists? The copy function would need to be a little smarter to handle an arbitrary input yearperiod and perhaps raise an exception if gaps were created.

gilmoskowitz:
requires https://github.com/xtuple/xtuple/pull/2102

PR #590, opened 2015-01-23
Issue #  25258 : recalculate line item pricing

discountscott:
Disconnect _shipto signal while populating... Then reconnect.


guiclient/salesOrder.cpp
```diff
@@ -2623,6 +2623,8 @@ void salesOrder::populate()
     qu.exec();
     if (qu.first())
     {
+	  //disconnect the shipTo signal while we populate... Fixes recalculate pricing popping up
+	  disconnect(_shipTo, SIGNAL(newId(int)), this, SLOT(sParseShipToNumber()));
```
garyhgohoos: i don't see how this change fixes the problem.  this change manipulates signals when populating a quote (not a sales order).  the probable cause of the problem is in the xwd salesOrder script.  this PR should be deleted

discountscott: We were experiencing the same issue on quotes... This change stopped our quotes from generating the pop-up message when they shouldn't.

garyhgohoos: the root cause of this problem is created by the xwd salesOrder script which is attempting to set the ship to based on the preferred selling site.  this change does not fix the root cause and does not fix the problem for Sales Orders and should be deleted.

gpazo:
@garyhgohoos I agree, can you look at the real fix and link it to this pull request so @discountscott can see the solution?

guiclient/salesOrder.cpp
garyhgohoos: correct fix is private-extensions PR 561

gilmoskowitz:
https://github.com/xtuple/private-extensions/pull/561

PR #593, opened 2015-01-29
use git in-place to help generate report diffs

gilmoskowitz:
also simplified some of the logic that does the real work


gpazo:
+1 looks good, would like to test next release

PR #594, opened 2015-01-30
Issue #25327:add dropship argument to itemcost for xwdspa

garyhgohoos:


gilmoskowitz:
+1 but need to merge https://github.com/xtuple/xtuple/pull/2151 at the same time.

PR #595, opened 2015-02-03
issue #20480 add quantity, weight to (Purchase Order line items tab) totals

jcarlin:
screen shot 2015-02-03 at 1 36 53 pm
screen shot 2015-02-03 at 1 36 21 pm

### DIFF
--- a/guiclient/purchaseOrder.cpp
+++ b/guiclient/purchaseOrder.cpp
@@ -971,7 +971,7 @@ void purchaseOrder::sSave()
     _tax->clear();
     _freight->clear();
     _total->clear();
-    _totalProdWt->clear();
+    _totalWeight->clear();
     _totalQtyOrd->clear();
     _poitem->clear();
     _poCurrency->setEnabled(true);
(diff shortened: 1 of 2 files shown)


jcarlin:
@gilmoskowitz can you take a look at this? Thanks!

guiclient/purchaseOrder.cpp
```diff
@@ -152,6 +152,17 @@ purchaseOrder::purchaseOrder(QWidget* parent, const char* name, Qt::WFlags fl)
...
   _projectId = -1;
+
+XSqlQuery getWeightUOM;
+getWeightUOM.prepare("SELECT uom_name FROM uom WHERE (uom_item_weight);");
+getWeightUOM.exec();
+if (getWeightUOM.first())
+  {
+    QString newLabel (tr("Total Weight ("));
+    newLabel += getWeightUOM.value("uom_name").toString();
+    _totalWeightLit->setText(newLabel + "):");
```
gilmoskowitz: This won't translate well. Use Qt's `args` mechanism to insert the `uom_name`. That'll look something like this (using `newLabel` as a holding variable is fine):
```c++
_totalWeightLit->setText(tr("Total Weight (%1):").arg(getWeightUOM.value("uom_name").toString());
```

guiclient/purchaseOrder.ui
```diff
@@ -667,111 +667,31 @@ to be bound by its terms.</comment>
        <item row="1" column="0">
         <layout class="QHBoxLayout">
```
gilmoskowitz: The layout simplification below is very nice. It could be made one level shallower by moving the spacer at 670 into the grid layout and eliminating this QHBoxLayout entirely. Give it a try and see how it works but don't struggle with it. If the results don't look right, just say so and we'll take this version.

jcarlin: It extended the middle column too much and I wasn't able to adjust the width of the spacer/column.

PR #601, opened 2015-02-15
Issue #25311: Better client mismatch message

andersdd:
Also tweaks license expiry message to refer to obtaining a new license rather than payment (Issue #25350).

### DIFF
--- a/guiclient/main.cpp
+++ b/guiclient/main.cpp
@@ -399,7 +399,7 @@ int main(int argc, char *argv[])
             checkPassReason = QObject::tr("<p>Your xTuple license expired over 30 days ago, and this software will no longer function. Please contact xTuple immediately to reinstate your software.");
           }
           else
-            checkPassReason = QObject::tr("<p>Attention:  Your xTuple license has expired, and in %1 days this software will cease to function.  Please make arrangements for immediate payment").arg(30 - daysTo);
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
+1 for the text changes so far. Much nicer.
The bug report also refers to the auto-update message, which is in https://github.com/xtuple/qt-client/blob/4_9_x/common/checkForUpdates.cpp

andersdd:
I looked at checkforUpdates and the text was already pretty simple so made the call no changes were required.

guiclient/main.cpp
```diff
@@ -530,21 +532,14 @@ int main(int argc, char *argv[])
...
                   QMessageBox::Ok | QMessageBox::Escape | QMessageBox::Default );
       return 0;
     }
     else {
      _splash->hide();
      result = QMessageBox::warning( 0, QObject::tr("Version Mismatch"),
-     QObject::tr("<p>The version of the database you are connecting to is "
-                 "not the version this client was designed to work against. "
-                 "This client was designed to work against the database "
+     QObject::tr("<p>This client was designed to work against the database "
```
davidbeauchamp: Should we put the server version in this message as well? It could be that the mismatch was unintentional and if they say they don't want to continue they won't know what version they need.

xtpclark: Yes - say what version the db is.

andersdd: Look at the full main.cpp file and you will see that the Db version is included in the message

PR #602, opened 2015-02-15
Issue #25345: Resolve Characteristic sort order

andersdd:


gilmoskowitz:
This is great for characteristics that don't have a defined order but some do, like `small`/`medium`/`large`. I'm not sure how to deal with this but the `char_order` field (that's where the old `17` came from) and `charopt` table seem like good places to start.

andersdd:
Good point (as always).  The underlying issue appears that the setSort is not cumulative.  What we really need to enable is sorting by "order" (17) *then* sort by "code" (1).   I will dig a bit deeper.

andersdd:
Had to switch from QTableModel as this only accepts a single column sort order.  Switched to QSqlQueryModel which allows sql syntax.

Tested and the revised sql honors the characteristic sort order THEN the characteristic name order which looks like it will suit all scenarios.

PR #609, opened 2015-02-20
Issue #17257:correct currency conversion for allocated credits

garyhgohoos:


guiclient/salesOrder.cpp
```diff
@@ -4515,7 +4513,7 @@ void salesOrder::sEnterCashPayment()
 {
   XSqlQuery cashsave;
 
-  if (_cashReceived->localValue() >  _balance->localValue() &&
+  if (_cashReceived->baseValue() >  _balance->baseValue() &&
```
gilmoskowitz: Are there cases where the cash received and the balance might be in different currencies? If not then the original comparison looks correct.

gilmoskowitz:
Paired with https://github.com/xtuple/xtuple/pull/2221

guiclient/salesOrder.cpp
garyhgohoos: yes, different currencies are possible

PR #618, opened 2015-03-01
Record individual payments without printing

dpocock:
If a check has been written by hand, or if an electronic payment sent from an account, this allows the user to fill in the check number or electronic transaction number as the document number without actually printing any check.

### DIFF
--- a/guiclient/printCheck.cpp
+++ b/guiclient/printCheck.cpp
@@ -38,6 +38,7 @@ printCheck::printCheck(QWidget* parent, const char* name, Qt::WFlags fl)
   connect(_check,     SIGNAL(newID(int)), this, SLOT(sEnableCreateEFT()));
   connect(_createEFT, SIGNAL(clicked()),  this, SLOT(sCreateEFT()));
   connect(_print,     SIGNAL(clicked()),  this, SLOT(sPrint()));
+  connect(_printed,   SIGNAL(clicked()),  this, SLOT(sPrintedAlready()));
 
   _captive = FALSE;
   _setCheckNumber = -1;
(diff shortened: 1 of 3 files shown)


andersdd:
Not sure if this functionality is entirely necessary.  There was a setting introduced on the Bank Setup in 4.8.0 that indicated whether the bank account required check printing.  If not set the Payment Run screen amends itself and allows you to post the payments without having to physically print a check.

Although this change does allow the same bank account to have both printed checks and electronic payments.

dpocock:
Dave, thanks for the feedback, I had explored the possibilities with that new setting you refer to and I think this is a useful complement to that.  In the UK, the faster payments, BACS and direct debit payments are free of charge and used for most payments but most accounts also come with chequebooks too and people occasionally write a cheque.  I probably write maybe one or two cheques a year on my UK account.  In Switzerland, I've never actually seen a chequebook so a Swiss user would just use the new option to turn off cheque printing and they would be quite satisfied.  There is one other catch though: when you turn off cheque printing in the settings, the document numbers are auto-generated and can't be manually inserted anywhere.  If you are writing the cheques by hand or if you make wire payments and you want to record the reference numbers then this new option I've added allows you to do that.

andersdd:
Daniel, coming from NZ,  I also have not seen a check for a long time which is the reason I added the ability to have "Electronic" banking accounts which do not require printing.  And yes the automatic "check number" becomes more of a payment reference number.

Agreed, your functionality complements the existing and I know many people are happy to see xTuple support electronic banking.

dpocock:
There are some more possibilities here too: people sending or receiving money with systems like Paypal or other payment processors may also want to log the transaction IDs in the check number field.

PR #625, opened 2015-03-04
Issue #25072: Purchased xTuple licenses are consumed by non xTuple DB connections.

mkhubbard:
This pull request is to present the work that has been started and to begin discussion for the changes necessary to completely resolve [Issue #25072](http://www.xtuple.org/xtincident/view/bugs/25072).

Requires PR xtuple/xtuple#2242.

This patch implements database connection naming using the "application_name" connection parameter to help identify the client connections. Once that was in place I was able to simplify the connection counting routine to focus only on activity reported in "pg_stat_activity". The previous process was going through the extra effort to explicitly count the advisory locks acquired by the application, at this point I'm not sure which one is preferred so I am starting with the simplest.

I have verified that the license enforcement works properly based on the following rules:

 * No more than the licensed number of users may be connected to one xTuple database at a time using the xTuple client.

 * No more than twice the number of licensed users may be connected to one xTuple database at a time. This includes xTuple clients, pgAdmin, and any other external connection sources.

From my understanding of the initial report from the customer this should address any counting errors which was causing the licensing to be enforced for 4 client and 2 pgAdmin connections when using a 5 user license. 

However the error report also contained the following statement: 

> A purchased 5 user license should allow all 5 users to each open one xTuple client connection, regardless of other background server tasks.

I didn't see any comments on the ticket to indicate that this type of change should be made but I see no problem in making the adjustment if that is the new desired behavior.

Once this has been finalized I can address any mismatch in the client / server connection counts that are being reported on the System -> Setup screen.

### DIFF
--- a/common/login2.cpp
+++ b/common/login2.cpp
@@ -171,6 +171,10 @@ int login2::set(const ParameterList &pParams, QSplashScreen *pSplash)
   if (valid)
     _setSearchPath = true;
 
+  param = pParams.value("applicationName", &valid);
+  if (valid)
+    _connAppName = param.toString().trimmed();
+
   if(pParams.inList("login"))
     sLogin();
 
@@ -329,7 +333,7 @@ void login2::sLogin()
   int methodidx; // not declared in for () because we'll need it later
   for (methodidx = 0; methodidx < method.size(); methodidx++)
   {
-    db.setConnectOptions(method.at(methodidx).first);
+    db.setConnectOptions(QString("application_name='%1';%2").arg(_connAppName).arg(method.at(methodidx).first));
     db.setPassword(method.at(methodidx).second);
     if (db.open())
       break;  // break instead of for-loop condition to preserve methodidx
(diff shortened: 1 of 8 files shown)


mkhubbard:
I thought about this some more and decided to return to counting the explicit advisory locks acquired by the qt client to determine the number of client connections. After reviewing the application this should be accurate and is also more forgiving if multiple connections need to be opened by the same client.

These changes are present in commit 98dcdf4.

gilmoskowitz:
These changes bring us closer to our [published policy](http://www.xtuple.com/faq#t23n2272). Updating **System** > **Setup** > **Database** and its underlying code should lead to further simplifications in the login count enforcement. It should also help with the details of xTuple-but-not-qt-client connections (e.g. xTuple Connect's _Batch Manager_).

mkhubbard:
Required stored function changes have been submitted by PR xtuple/xtuple#2242.

Connection counting in the qt-client now uses the database stored functions which ensures the license enforcement process and usage reporting in the application are working on the same set of rules.

The changes for supplying the "application_name" database connection parameter have also been cleaned up. All instances of "login2" in the qt-client have been updated to pass the qt-client application name identifier. Additionally the "login2" class has been updated to default the "applicationName" parameter to "xTuple ERP (unknown)" if not provider by the calling process.

At this point I feel these changes are complete and resolve the concerns raised by Issue #25072.

PR #632, opened 2015-03-10
Issue #25493:restrict item list on existence in bomitem

garyhgohoos:


guiclient/displays/dspIndentedWhereUsed.cpp
```diff
@@ -30,9 +30,20 @@ dspIndentedWhereUsed::dspIndentedWhereUsed(QWidget* parent, const char*, Qt::WFl
   _worksetid = 0;
 
   if (_metrics->boolean("AllowInactiveBomItems"))
-    _item->setType(ItemLineEdit::cGeneralComponents);
+    _item->setQuery( QString( "SELECT DISTINCT item_id, item_number, item_descrip1, item_descrip2,"
```
gilmoskowitz: Since `_item` supports different queries via these flags, it make sense to add a new option `cHasBOM` to the widget instead of cloning the query N times.

PR #633, opened 2015-03-10
first pass at exposing QApplication singleton

gilmoskowitz:
This change is intended to aid in the current UI/UX push. It should now be possible to script something like
```JavaScript
QApplication.styleSheet = "QPushButton { background-color: red; }";
QApplication.styleSheet += "QComboBox { border-left-style: dotted; border-right-style: dotted; }";
```
and other such silliness.

### DIFF
--- a/scriptapi/qapplicationproto.cpp
+++ b/scriptapi/qapplicationproto.cpp
@@ -10,11 +10,9 @@
 
 #include "qapplicationproto.h"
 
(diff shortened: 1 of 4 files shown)


gpazo:
very cool, does this compile in qt5?

gilmoskowitz:
@gpazo - it does now

PR #635, opened 2015-03-11
Issue #22800:add customer defaults for ship charge and preferred selling...

garyhgohoos:
... warehouse


bendiy:
@garyhgohoos Can you modify these queries to support these new defaults?
https://github.com/xtuple/xtuple/blob/4_9_x/database/source/xm/javascript/customer.sql#L239-L241

bendiy:
The way this works, I think we can just change these lines:
https://github.com/xtuple/xtuple/blob/4_9_x/database/source/xm/javascript/customer.sql#L247
`"shipCharge": shipCharge,`
to
`"shipCharge": settings.DefaultShipChrgId,`

https://github.com/xtuple/xtuple/blob/4_9_x/database/source/xm/javascript/customer.sql#L261
`"preferredSite": preferredSite`
to
`"preferredSite": settings.DefaultSellingWarehouseId`

Now that they are in the metrics table.

PR #639, opened 2015-03-16
Issue #25510: Set Documents to Read only when in View mode

andersdd:
Not sure why setting Documents to read only was commented out on several documents.  But added the read-only back to the highlighted business objects


gilmoskowitz:
[This](https://github.com/andersdd/qt-client/commit/94db1752ac5418e902450bd97b0138198d1805da) proposed fix conflicts with https://github.com/xtuple/qt-client/commit/e744ad8901c97d0fb540d5bef1b68c1ff3371763
@andersdd and @garyhgohoos - Let's try to figure out what the real complaint was with 20996 so we can find a solution that addresses both issues.

andersdd:
@gilmoskowitz @garyhgohoos  so we have one user requesting the ability to add documents in View mode as a feature, and other considering this as a bug.  Seems to me we just make a call based on business logic and run with one or the other.  On reading 20996 I can see some merit in being able to attach documents after the fact, but it is inconsistent with the view mode elsewhere in the app.

gilmoskowitz:
After discussing with Tech Support, we've marked 25510 as "won't fix" because of 20996, and created issue #25607 - create an audit trail of document associations.

PR #640, opened 2015-03-17
Issue #21596: Post Documents in a single step

andersdd:
Add ability to post from within Misc Invoice and Misc Voucher documents.  Significant time saving for users who wish to utilise this feature.

### DIFF
--- a/guiclient/invoice.cpp
+++ b/guiclient/invoice.cpp
@@ -19,6 +19,7 @@
 #include <QDebug>
 
 #include "characteristicAssignment.h"
+#include "distributeInventory.h"
 #include "invoiceItem.h"
 #include "storedProcErrorLookup.h"
 #include "taxBreakdown.h"
@@ -124,6 +125,8 @@ invoice::invoice(QWidget* parent, const char* name, Qt::WFlags fl)
   _recurring->setParent(-1, "I");
 
   _miscChargeAccount->setType(GLCluster::cRevenue | GLCluster::cExpense);
+
+  _postInvoice->setEnabled(_privileges->check("PostMiscInvoices"));
 }
 
 invoice::~invoice()
@@ -275,6 +278,7 @@ enum SetResponse invoice::set(const ParameterList &pParams)
       _saleType->setEnabled(FALSE);
 //      _documents->setReadOnly(TRUE);
       _newCharacteristic->setEnabled(FALSE);
+      _postInvoice->setVisible(FALSE);
 
       disconnect(_invcitem, SIGNAL(valid(bool)), _edit, SLOT(setEnabled(bool)));
       disconnect(_invcitem, SIGNAL(valid(bool)), _delete, SLOT(setEnabled(bool)));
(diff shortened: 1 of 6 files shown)


guiclient/invoice.ui
```diff
@@ -44,6 +44,13 @@ to be bound by its terms.</comment>
      </item>
      <item>
       <layout class="QHBoxLayout">
+     <item>
+      <widget class="QCheckBox" name="_postInvoice">
+       <property name="text">
```
gilmoskowitz: This should be an XCheckBox

guiclient/invoice.cpp
```diff
@@ -728,6 +736,104 @@ bool invoice::save()
...
+                __FILE__, __LINE__);
+     }
+     else if (!xrate.first() || xrate.value("curr_rate").isNull())
+     {
+       systemError(this, tr("Could not post Invoice #%1 because of a missing exchange rate.")
+				.arg(_invoiceNumber->text()));
+     }
+  }
+
+  unpostedPost.exec("BEGIN;");	// because of possible lot, serial, or location distribution cancelations
```
gilmoskowitz: The transaction appears to start too late - the current structure allows posting the invoice without distributing inventory. With the current separation, how & when will the inventory get distributed if there's an error or the user cancels the distribution?

guiclient/invoice.ui
```diff
@@ -44,6 +44,13 @@ to be bound by its terms.</comment>
      </item>
      <item>
       <layout class="QHBoxLayout">
+     <item>
+      <widget class="QCheckBox" name="_postInvoice">
```
gilmoskowitz: This should be an `XCheckBox`.

guiclient/invoice.ui
```diff
@@ -1472,8 +1481,6 @@ to be bound by its terms.</comment>
   <tabstop>_salesrep</tabstop>
   <tabstop>_commission</tabstop>
   <tabstop>_taxzone</tabstop>
-  <tabstop>_cust</tabstop>
-  <tabstop>_copyToShipto</tabstop>
```
gilmoskowitz: Was this change in tab order made consciously?

guiclient/miscVoucher.ui
```diff
@@ -40,6 +40,13 @@ to be bound by its terms.</comment>
       </spacer>
      </item>
      <item>
+      <widget class="QCheckBox" name="_postVoucher">
```
gilmoskowitz: This should be an `XCheckBox`.

guiclient/invoice.ui
andersdd: Yes it was - the first thing required on an invoice is the customer, but that was not the starting widget

guiclient/invoice.cpp
andersdd: Interesting given I copied this logic from the post invoice function in the Unposted Invoices screen.  I will check further. May have to apply the logic elsewhere.

gilmoskowitz: hmmm... Then I might be wrong about the structure of the transaction. @garyhgohoos - do you have an opinion here?

garyhgohoos: the sum and xrate queries are performed outside of the transaction.  the post and distribution are inside of the begin/commit of the transaction.  looks ok to me.

garyhgohoos: post.prepare is outside, but post.exec is inside

gilmoskowitz: Ahhh... Thanks @garyhgohoos !

PR #648, opened 2015-03-31
Issue #21642: Add Sale Type to Sales Reports

andersdd:
Add parameter filter and display in list

relies on PR xtuple/xtuple#2261


gpazo:
+1 looks good, waiting on 2261 merge

PR #649, opened 2015-03-31
Issue #25612: Add Sale Type default checkbox to set up to fix SO bug

andersdd:
SO defaulted to first alphabetically listed Sale Type when not manually set. The default could change if Sale Types were updated.  This change allows a default Sale Type to be set and used in SO automatically.

Requires by xtuple/xtuple#2262

### DIFF
--- a/guiclient/saleType.ui
+++ b/guiclient/saleType.ui
@@ -106,7 +106,7 @@ to be bound by its terms.</comment>
               <string>Default</string>
              </property>
              <property name="checked">
-              <bool>true</bool>
+              <bool>false</bool>
              </property>
             </widget>
            </item>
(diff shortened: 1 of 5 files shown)


guiclient/saleType.ui
```diff
@@ -101,6 +101,16 @@ to be bound by its terms.</comment>
             </widget>
            </item>
            <item>
+            <widget class="QCheckBox" name="_default">
+             <property name="text">
+              <string>Default</string>
+             </property>
+             <property name="checked">
+              <bool>true</bool>
```
gilmoskowitz: There will be fewer default values than non-defaults so the `checked` property should be `false` (or unspecified in the `.ui`).

andersdd: Yes - missed that in my cut and paste.  Will fix.

PR #656, opened 2015-04-10
Issue #24940: Add user locking to Customer Edit

andersdd:
Add user locking to Customer, Vendor, Item, & CRM Account

Requires #659 - which has already been merged into this branch

### DIFF
--- a/guiclient/crmaccount.cpp
+++ b/guiclient/crmaccount.cpp
@@ -225,15 +225,7 @@ enum SetResponse crmaccount::set(const ParameterList &pParams)
     else if (param.toString() == "view")
     {
       _mode = cView;
-
-      _save->hide();
-      disconnect(_reg, SIGNAL(valid(bool)), _editReg, SLOT(setEnabled(bool)));
-      disconnect(_reg, SIGNAL(valid(bool)), _deleteReg, SLOT(setEnabled(bool)));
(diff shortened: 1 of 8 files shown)


gilmoskowitz:
@andersdd - This is a great start. Note that https://github.com/andersdd/qt-client/blob/24940/guiclient/customer.cpp#L752 changes the id, so the applock will have to release the old id and acquire the new id. There might be other ways to change the record being worked on as well.

Question: How good is the Applock API? What about it should be changed to make it better? It's better to figure that out now than after we propagate it further.

andersdd:
Additional logic added to release/re-acquire lock when customer number edited.

Regarding the Applock API, it appears to work fairly well and fairly simply.  My suggestions would be to encapsulate more of the logic in the API rather than just passing a true/false result back to the calling screen.  The messaging should be included in the API instead of the calling screen, to save effort, and ensure consistency.  

Also it would be useful, and typical to display which user was editing and had locked the screen, so resolving the lock could be followed up.

gilmoskowitz:
API change request recorded as dogfood issue 25676

gilmoskowitz:
The revised API is in https://github.com/xtuple/qt-client/pull/659 and can be used to finish work on issue #24940 with the CRM Account, Vendor, and Item windows.

andersdd:
@gilmoskowitz I have reviewed #659 and it is looking much more concise.  I will merge your changes into this branch and finish off the locks on the remaining objects.

PR #657, opened 2015-04-12
Issue #25652: Resize XComboBox after populating 

andersdd:
Ensures lengthy combobox entry options can be read.

also resolves issues #25653 #25654

NOTE: this change causes overlapping widgets on some screens (Sales Order) on the Linux client.  Needs to be tested on more screens on the Windows and Mac desktop.


gpazo:
@andersdd +1 change looks simple enough, can you elaborate on the overlapping? this widget is used all over the app so this may cause all kinds of odd behavior

gilmoskowitz:
@gpazo - when the text get really wide, the combobox gets wider than the available space. it then draws either over or behind adjacent widgets. this happens on both linux and mac osx. i've submitted a modification - https://github.com/andersdd/qt-client/pull/1 - to see if it solves the customer problem without creating more.

andersdd:
This PR (with @gilmoskowitz mod) reverts behaviour back to pre-change on a Linux system.  The issue was raised by a customer using Windows, so we really need to test this change on a Windows build.

gilmoskowitz:
Working on it... My Windows build just finished...

PR #659, opened 2015-04-13
issue #25676 - make applock optionally interactive

gilmoskowitz:
issue #25676 - extend the API for AppLock so it can either report its own errors or let the caller do it, then change existing callers to use the revised version


andersdd:
+1 code review done.  Plus also unit tested on several locking objects

PR #664, opened 2015-04-22
Issue #25702 Add Zone filtering in Location and Inventory Distribution screens

andersdd:
Used where significant warehouses are broken into several zones

### DIFF
--- a/guiclient/distributeInventory.ui
+++ b/guiclient/distributeInventory.ui
@@ -13,8 +13,8 @@ to be bound by its terms.</comment>
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>849</width>
-    <height>572</height>
+    <width>783</width>
+    <height>527</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -176,7 +176,7 @@ to be bound by its terms.</comment>
             </item>
            </layout>
           </item>
-          <item row="1" column="0">
+          <item row="2" column="0">
            <widget class="QLabel" name="_locationDefaultLit">
             <property name="text">
              <string>Location Default: </string>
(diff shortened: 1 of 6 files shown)


guiclient/distributeInventory.ui
```diff
@@ -176,7 +176,7 @@ to be bound by its terms.</comment>
             </item>
            </layout>
           </item>
-          <item row="1" column="0">
+          <item row="2" column="0">
```
gilmoskowitz: These changes to `row` don't look right.

guiclient/locations.cpp
```diff
@@ -167,22 +171,47 @@ void locations::sFillList()
...
-               "WHERE ( (location_warehous_id=warehous_id)" );
+               "FROM location  "
+               " JOIN whsinfo ON (location_warehous_id=warehous_id) "
+               " LEFT OUTER JOIN whsezone ON (location_whsezone_id=whsezone_id) "
+               " WHERE ( (true)" );
 
   if (_warehouse->isSelected())
     sql += " AND (warehous_id=:warehous_id)";
 
+  if (_zone->id() > 0)
```
gilmoskowitz: The query should be [converted to MetaSQL](https://github.com/xtuple/qt-client/wiki/xTuple-Desktop-Client-Coding-Standards#sql-and-metasql) instead of being build by string concatenation.

guiclient/distributeInventory.ui
andersdd: Not sure how these came about, but they are inconsequential.  I have manually reverted to the original row settings.

guiclient/locations.cpp
andersdd: Converted to metaSQL and tested.   Unfortunately pushing to git from China is a hit and miss affair, and I appear to be missing today.

gilmoskowitz:
@andersdd - I merged because it's better to have the fix in the beta than not. When you have more success pushing to github, please open a new pull request with the metasql and .ui tweaks and I'll merge those, too.

PR #665, opened 2015-04-22
Issue 25734

discountscott:
join shipto_name on purchase orders instead of using the CRMAcct_Name

### DIFF
--- a/guiclient/purchaseOrder.cpp
+++ b/guiclient/purchaseOrder.cpp
@@ -1870,13 +1870,13 @@ void purchaseOrder::sHandleShipTo()
 void purchaseOrder::sHandleShipToName()
 {
   XSqlQuery purchaseHandleShipTo;
-  purchaseHandleShipTo.prepare( "SELECT * "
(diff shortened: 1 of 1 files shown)


guiclient/purchaseOrder.cpp
```diff
@@ -1871,12 +1871,12 @@ void purchaseOrder::sHandleShipToName()
...
   purchaseHandleShipTo.prepare( "SELECT * "
-                               "FROM address "
+	  "FROM address left outer join shiptoinfo on (shipto_addr_id=addr_id) "
                                "WHERE (addr_id=:addr_id);" );
   purchaseHandleShipTo.bindValue(":addr_id", _shiptoAddr->id());
   purchaseHandleShipTo.exec();
   if (purchaseHandleShipTo.first())
   {
-    _shiptoName->setText(purchaseHandleShipTo.value("crmacct_name").toString());
+    _shiptoName->setText(purchaseHandleShipTo.value("shipto_name").toString());
```
garyhgohoos: not every address will be a valid shipto address.  change query to "SELECT COALESCE(ship to_name, crmacct_name) AS shiptoname" and use query value of "shiptoname" to set _shiptoName.

PR #669, opened 2015-04-27
Fix Distribute layout.  Convert locations sql to metaSQL

andersdd:
Fixes to Item Distribution Layout.  Update sql in locations screen to metaSQL

### DIFF
--- a/guiclient/locations.cpp
+++ b/guiclient/locations.cpp
@@ -180,11 +180,11 @@ void locations::sFillList()
                " JOIN whsinfo ON (location_warehous_id=warehous_id) "
                " LEFT OUTER JOIN whsezone ON (location_whsezone_id=whsezone_id) "
                " WHERE ( (true)" 
-               " <? if exists(\"warehous_id\") ?>"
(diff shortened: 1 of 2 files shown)


guiclient/locations.cpp
```diff
@@ -179,39 +179,44 @@ void locations::sFillList()
...
 {
-  XSqlQuery zoneFillList;
   QString zoneSql( "SELECT whsezone_id, whsezone_name||'-'||whsezone_descrip "
              " FROM whsezone  "
-             " WHERE ((true)  " );
-  if (_warehouse->isSelected())
-    zoneSql += " AND (whsezone_warehous_id=:warehous_id)";
+             " <? if exists(\"warehous_id\") ?> "
+             " WHERE (whsezone_warehous_id = <? value(\"warehous_id\") ?>) "
+             " ORDER BY whsezone_name;");
```
gilmoskowitz: Missing `endif`?
Also, FYI - MetaSQL can handle either single- or double-quotes so there's no need to backslash-".

PR #672, opened 2015-05-01
Issue #24663: Prevent addition of jobcosted items to BOMs

andersdd:
Add check to BOM item addition to prevent if the item is job costed.  Apparently causes all sorts of costing issues and orphaned elements.

### DIFF
--- a/guiclient/bomItem.cpp
+++ b/guiclient/bomItem.cpp
@@ -623,7 +623,7 @@ void bomItem::sItemIdChanged()
   }
   else if (qry.lastError().type() != QSqlError::NoError)
       ErrorReporter::error(QtCriticalMsg, this, tr("Item Job cost check"),
-                           _item->id(), __FILE__, __LINE__);
+                           qry, __FILE__, __LINE__);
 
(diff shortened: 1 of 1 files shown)


guiclient/bomItem.cpp
```diff
@@ -609,6 +609,21 @@ void bomItem::sFillSubstituteList()
...
+  if (qry.first() && qry.value("exists") == TRUE)
+  {
+    QMessageBox::question(this, tr("Job Costed Item"),
+                            tr("You may cannot add a Job Costed item to a Bill of Material"));
+    _item->setId(-1);
+    return;
+  }
+  else if (qry.lastError().type() != QSqlError::NoError)
+    systemError(this, qry.lastError().databaseText(), __FILE__, __LINE__);
+
```
gilmoskowitz: - `SELECT 1 FROM itemsite...` so we don't grab the whole row
- `qry.value("exists").toBool()` instead of `[QVariant] == TRUE` (simplifies the boolean check, doesn't depend on implicit casting, and `TRUE` doesn't work with Qt5)
- `ErrorReporter::error()`

PR #674, opened 2015-05-03
Issue #25665: Fix WO Material Return for Lot Items with different qtys

andersdd:
When returning Lot Qty greater than issued Qty an error occurred (when the lot was issued multiple times).  This fix sums the qty issued to check the correct value.

### DIFF
--- a/guiclient/createLotSerial.cpp
+++ b/guiclient/createLotSerial.cpp
@@ -340,14 +340,17 @@ void createLotSerial::sAssign()
   
   if (_preassigned)
   {
-    createAssign.prepare("SELECT lsdetail_qtytoassign "
-              "FROM lsdetail "
-              "WHERE (lsdetail_id=:lsdetail_id);");
+    createAssign.prepare("SELECT SUM(lsd.lsdetail_qtytoassign) AS qtytoassign "
+              "FROM lsdetail lsd "
+              "JOIN lsdetail lsd2 "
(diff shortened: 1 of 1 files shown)


guiclient/createLotSerial.cpp
```diff
@@ -340,14 +340,17 @@ void createLotSerial::sAssign()
...
   {
-    createAssign.prepare("SELECT lsdetail_qtytoassign "
-              "FROM lsdetail "
-              "WHERE (lsdetail_id=:lsdetail_id);");
+    createAssign.prepare("SELECT SUM(lsd.lsdetail_qtytoassign) AS qtytoassign "
+              "FROM lsdetail lsd "
+              "JOIN lsdetail lsd2 "
+              "  ON (lsd.lsdetail_source_id=lsd2.lsdetail_source_id "
+              "    AND lsd.lsdetail_ls_id = lsd2.lsdetail_ls_id) "
+              " WHERE (lsd2.lsdetail_id=:lsdetail_id);");
```
gilmoskowitz: This should probably include `lsdetail_source_type` and `lsdetail_source_number` in the join detail to ensure we're talking about the same transaction.

andersdd: Not sure source_number is essential as the source_id is included, but I have taken your suggestions on board and added to the JOIN.

PR #676, opened 2015-05-03
Issue #25727: Sales Order was not resetting when Ship To cleared

andersdd:
When Ship To widget is cleared SO is now reset back to customer defaults


gilmoskowitz:
+1
@garyhgohoos - could you look at this as well? If you approve then let's merge it.

PR #679, opened 2015-05-05
Issue #25729: Various UOM handling improvements and consistency

andersdd:
Build in ability to deactivate Item UoM conversions

Screens previously running individual, duplicated SQL now reference single metaSQL and take account of active status of conversion.

Screens also consistently take account of the conversion "type" (e.g Selling) which they were not previously with potentially confusing results.

requires xtuple/xtuple#2299

### DIFF
--- a/guiclient/creditMemoItem.cpp
+++ b/guiclient/creditMemoItem.cpp
@@ -15,6 +15,10 @@
 #include <QValidator>
 #include <QVariant>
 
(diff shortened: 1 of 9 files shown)


gilmoskowitz:
bleh... it feels like there's a better way to handle this but i don't see it.

andersdd:
I agree.  Open to suggestions

PR #681, opened 2015-05-07
Five fixd

gpazo:
DO NOT MERGE!


common/format.cpp
```diff
@@ -128,7 +128,7 @@ int decimalPlaces(QString pName)
   if (!loadedLocales)
     loadLocale();
 
-  QByteArray  tba  = pName.toAscii();
+  QByteArray  tba  = pName.toLatin1();
```
andersdd: Not sure about using Latin1 here.  This is only a single byte format and precludes double-byte languages like Japanese / Chinese.  Should we be using UTF-8 here?  @gilmoskowitz might know better.

andersdd:
Looks like all the Copyright notices in the scriptapi folder have reverted from 2014 to 2012.  Probably need to sed all the files up to 2015.

andersdd:
In scriptapi/qprinterproto.cpp the constructor.setProperty("PostScriptFormat" is commented out.  Is this intentional?

BTW I've gone past the Github file limit so can't find a way to do line notes on the subsequent file diffs

andersdd:
Do we want to be consistent on Qt5 version checks?

#if QT_VERSION >= 0x050000   versus  #if QT_VERSION < QT_VERSION_CHECK(5,0,0)

also 
Q_OS_MAC  versus Q_OS_DARWIN   (I understand these are synonyms)

andersdd:
+1 code reviewed.  I can't confess to reading _every_ line of the rather large raw diff, but I did look for patterns and other than my minor comments above could see no glaring issues.  Going to compile and test locally.

common/format.cpp
gilmoskowitz: `toLatin1()` looks like the right change in general. We were already shifting to single-byte data so this doesn't introduce new problems.

gilmoskowitz:
Noticed the following in the widgets directory:

- widgets/documents.cpp, ~line 70 - removing the `_x_privileges` check will cause problems in Qt Designer, ranging from spurious warnings to Designer crashes
- widgets/docAttach.cpp, ~line 69 - removing the `_x_privileges` check will cause problems in Qt Designer
- widgets/xtupleplugin.h calls `Q_PLUGIN_METADATA` without a `FILE` param. Is this a bug?
- widgets/filemoveselector.cpp script exposure of the constructor won't behave properly if the widget creating the FileMoveSelector isn't a Qt::Window, which may lock up the application in some circumstances
- There are changes of `TRUE` to `true` inside strings (e.g. `widgets/shipmentCluster.cpp`). I didn't see any bugs this might cause but we should be ready for 'em.

gpazo:
@gilmoskowitz the x_privileges code in documents.cpp and docAttach.cpp looked to be removed because five-fixd wasn't up-to-date with 4_9_x; upmerge resolves this.

gilmoskowitz:
Noticed the following in the scriptapi and utility directories:

- Doxyfile.public has an improper `PROJECT_NUMBER` change
- why was xsqltablnodeproto.cpp removed?
- qurlproto.h has a lot of unnecessary churn - the QT_VERSION_CHECK block for `addEncodedQueryItem()` through `setEncodedUserName()` looks like it'll remove functionality - and doesn't quite seem to match the .cpp. We use URLs in several critical extensions, such as xwd and xtdesktop, so Qt4 compatibility is important. The concern raised in code comments over return and param type changes can be ameliorated using stupid C-preprocessor tricks.
- qtsetup.h and .cpp handling of AnchorAttribute, UniteClip, LogText could be `ifdef`ed

PR #688, opened 2015-05-13
issue #25477 check to see if curl is actually at the file path, retur…

gpazo:
…n error if no curl exists

### DIFF
--- a/guiclient/creditcardprocessor.cpp
+++ b/guiclient/creditcardprocessor.cpp
@@ -25,6 +25,7 @@
 #include <QUrl>
 #include <QBuffer>
 #include <QDebug>
+#include <QFileInfo>
 
 #include <currcluster.h>
 #include <metasql.h>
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
fails CI with
```
creditcardprocessor.cpp:1933:49: error: cannot pass objects of non-trivially-copyable type ‘class QString’ through ‘...’
creditcardprocessor.cpp:1933:49: warning: format ‘%s’ expects argument of type ‘char*’, but argument 2 has type ‘QString’ [-Wformat]
creditcardprocessor.cpp:1937:26: error: no matching function for call to ‘QString::arg(QFileInfo&)’
```

guiclient/creditcardprocessor.cpp
```diff
@@ -1924,6 +1925,20 @@ int CreditCardProcessor::sendViaHTTP(const QString &prequest,
...
+     {
+       qDebug("%s", checkCurl.absoluteFilePath());
+     }
+     QApplication::restoreOverrideCursor();
+     _errorMsg = errorMsg(-18)
+           .arg(checkCurl)
+           .arg("")
+           .arg(QString(proc.readAllStandardError()));
+     return -18;
+    }
```
gilmoskowitz: Why use `QFileInfo::exists()` instead of `[QProcess::error()](http://doc.qt.io/qt-4.8/qprocess.html#error)`?

PR #692, opened 2015-05-14
issue #25741 - extension char assoc

gilmoskowitz:
This is the basic client-side code to support characteristic association with arbitrary document types. Merge with the corresponding server-side pull request xtuple/xtuple#2215

This does not include a generic widget that can be added to a scripted window. Like the C++ core, extensions currently have to build an entire characteristics tab with xtreewidget and new/edit/delete buttons.


andersdd:
+1 reviewed.   Looks good.

PR #699, opened 2015-05-18
Issue #23905: Additional checks on app startup

andersdd:
Add warnings for missing (open) fiscal periods and missing exchange rates as these cause errors inside the app if not set properly - as well as causing lockups in the mobile web client.

### DIFF
--- a/guiclient/main.cpp
+++ b/guiclient/main.cpp
@@ -796,6 +796,38 @@ int main(int argc, char *argv[])
                     "transactions in the system.") );
   }
 
+//  Check for valid current Fiscal period
+  XSqlQuery periodCheck;
+  periodCheck.prepare("SELECT EXISTS(SELECT * FROM period "
+            "     WHERE ((current_date BETWEEN period_start AND period_end) "
+            "       AND (NOT period_closed))) AS result; ");
+  periodCheck.exec();
+  if(periodCheck.first() && periodCheck.value("result").toBool() != true)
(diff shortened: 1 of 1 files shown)


davidbeauchamp:
I feel like these error messages should be directed at a user who can do something about it. I often field questions about error messages like these from users who don't know why they are receiving them... anyone else have an opinion on this?

gilmoskowitz:
@davidbeauchamp - i agree that messages like these are best directed to a person who is able to do something about them. it's also better to catch them early, rather than after people have started to do work that they cannot save. i wasn't expecting us to start addressing this problem yet so hadn't discussed my preferred strategy:

- give **OK** and **Quit** buttons to casual users
- give **Continue**, **Open _RelevantWindow_**, and **Quit** buttons to people who have sufficient privs to fix the problem

davidbeauchamp:
Does the OK button for the user do something different than the Continue button for the person with privs?

gilmoskowitz:
No, the OK and Continue buttons would be equivalent & could just have the same text.

andersdd:
There are other (existing) checks upon startup/login.  I assume we should be applying the same logic back to those as well?

gilmoskowitz:
@andersdd and @davidbeauchamp - After thinking through the consequences of adding these checks and *Open* buttons and ..., it's probably best to just add the checks as Dave has done and leave the complexity for another day.

PR #700, opened 2015-05-18
issue #25741 - build a characteristics widget and use it once

gilmoskowitz:

PR #701, opened 2015-05-18
issue #25741 - extend the use of the new characteristics widget

gilmoskowitz:
**Wait to merge?** - This builds on #700 by using the new CharacteristicsWidget in most of the application. This is pretty invasive, so we should either wait until after 4.9.0 or test it very carefully before merging.


andersdd:
Initial Unit Testing:
*> Missing primary key on charuse table*
> Able to create new characteristics of all type (text|date|list) and assign to one or more document types
> Able to use above chars in the assigned documents.  Create new, edit, delete
*> Lot/Serial char assignment is broken (as you are aware).  Only displays the first characteristic (QA-RESULTS)*
> Item screen not migrated to new widget but works as normal 
*> Need to watch migration of existing types to the new document/characteristic method.  It appears the Quality (mobile) additions to the _source_ table is missing data.
> Added new characteristic/document type to _source_ table
> New type was immediately available in the Characteristic creation list as a checkbox type and I could create a new Char for that type.
> Scripted characteristics not fully tested, but was able to add characteristic widget to the scripted screen and assign the type.  Upon opening the screen the characteristic corresponding to the new type was correctly displayed.

_Awesome job_ @gilmoskowitz - other than the Lot/Serial everything looked to be working seamlessly, and existing Db characteristics appear to have migrated without issue.

I would suggest some instructions (or a .org post) at some stage so Partners can make best advantage of these new features.

gilmoskowitz:
@andersdd - thanks for the thorough testing. a few notes:

- primary key on charuse added
- I couldn't reproduce the only-one-characteristic problem with lot/serial characteristic assignment. Could you please log a bug with full steps to reproduce? Meanwhile I tweaked the lot/serial registration to autopopulate the characteristics widget when a lot/serial is selected.
- What's missing from the `source` table for the `quality` extension?

andersdd:
- Lot/Serial:  Basically only the first characteristic was displayed on the Create Lot/Serial screen when receipting an item.  I had created a second Lot/Serial characteristic that did not appear on this screen.  I will raise an issue with full steps.
- I did more research after the comment about Quality data in the source table, and realised the source table is used for Comments, Documents and Characteristics.  Quality is only using the Comments side - so any actions are mine if/when I extend the Quality Tests to include documents/characteristics (it is on the list).

PR #706, opened 2015-05-19
Update RPM spec file, use desktop-file-install

jakeholcombe:


jakeholcombe:
This will need to be tagged 4.8.2 so the desktop file changes and appdata file will be in the 4.8.2 tarball download

PR #709, opened 2015-05-21
Issue #25701: Modify reference to check on Vendor screen

andersdd:
Missed in previous PR.

Question though - why do we have a Payments radio on the Vendor ui and then immediately upon selection have a checkbox enabling EFT payments?  Could just be the one option?


gilmoskowitz:
The radio button exists because xTuple Connect adds another radio button and page to the (invisible) QStackWidget for Vendor EDI. We could redesign the base UI definition but would have to change xTuple Connect in sync.

PR #713, opened 2015-05-28
switch from enum to string based document types

garyhgohoos:
Please give this change a thorough code review

### DIFF
--- a/widgets/comment.cpp
+++ b/widgets/comment.cpp
@@ -135,19 +135,17 @@ comment::comment( QWidget* parent, const char* name, bool modal, Qt::WindowFlags
   resize( QSize(524, 270).expandedTo(minimumSizeHint()) );
   //clearWState( WState_Polished );
 
-// signals and slots connections
   connect(buttonBox, SIGNAL(accepted()), this, SLOT(sSave()));
   connect(buttonBox, SIGNAL(rejected()), this, SLOT(reject()));
   connect(_next, SIGNAL(clicked()), this, SLOT(sNextComment()));
   connect(_prev, SIGNAL(clicked()), this, SLOT(sPrevComment()));
   connect(_more, SIGNAL(toggled(bool)), _comments, SLOT(setVisible(bool)));
 
-// tab order
   setTabOrder( _cmnttype, _comment );
   setTabOrder( _comment, _save );
   setTabOrder( _save, _close );
 
-  _source = Comments::Uninitialized;
+  _sourcetype = "";
   _cmnttype->setAllowNull(true);
 
   shortcuts::setStandardKeys(this);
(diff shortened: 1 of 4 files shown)


widgets/comment.cpp
```diff
@@ -388,120 +218,81 @@ void comment::set(const ParameterList &pParams)
   param = pParams.value("sourceType", &valid);
   if (valid)
   {
-    _source = (enum Comments::CommentSources)param.toInt();
+    _sourcetype = param.toString();
     if(!(_mode == cEdit || _mode == cView))
     {
-      switch (_source)
```
gpazo: Not sure I agree with removing this but if it's easier to read, carry on

widgets/comments.cpp
```diff
@@ -159,9 +203,21 @@ Comments::Comments(QWidget *pParent, const char *name) :
...
 
-void Comments::setType(enum CommentSources pSource)
+int Comments::type() const
 {
-  _source = pSource;
+  CommentMap *elem = _strMap.value(_sourcetype);
+  return elem ? elem->doctypeId : Uninitialized;
+}
+
+void Comments::setType(int sourceType)
```
gpazo: 2 functions, same name but different input types?

davidbeauchamp: Also known as an overload

widgets/comment.cpp
gilmoskowitz: If we're going to use string comparisons, the cascading else/if is required in C++. It happens to be much slower than an integer-based switch.

A more extensible way, faster than the else/if but slower than the switch, would be to do a lookup from the `QMap<QString, struct CommentMap*> _strMap` from comments.cpp below.

widgets/comments.cpp
```diff
@@ -38,55 +38,99 @@ void Comments::showEvent(QShowEvent *event)
...
+//    addToMap(TimeExpense,       "TE",    tr("Time Expense")                                     );
+    addToMap(TodoItem,          "TD",    tr("To-Do"),           "todoitem_id","todoItem",     "MaintainPersonalToDoItems MaintainAllTodoItems" );
+    addToMap(TransferOrder,     "TO",    tr("Transfer Order"),  "tohead_id",  "transferOrder"   );
+    addToMap(TransferOrderItem, "TI",    tr("Transfer Order Item")                              );
+    addToMap(Vendor,            "V",     tr("Vendor"),          "vend_id",    "vendor"          );
+//    addToMap(Voucher,           "VCH",   tr("Voucher"),         "vohead_id",  "voucher"         );
+    addToMap(Warehouse,         "WH",    tr("Site")                                             );
+    addToMap(WorkOrder,         "W",     tr("Work Order"),      "wo_id",      "workOrder"       );
+    addToMap(Task,              "TA",    tr("Project Task"),    "prjtask_id", "projectTask"     );
+  }
```
gilmoskowitz: Looks like `TimeAttendance , "TATC"` is missing from the map.

PR #716, opened 2015-05-29
add starter translation files

gilmoskowitz:


piercetyler:
verified correct number of translation files. looks good.

PR #727, opened 2015-06-05
update to 4.9.0

gpazo:


gilmoskowitz:
safer to record as a git merge - #728

PR #734, opened 2015-06-10
issue #25694 fix longstanding menubar bug

gpazo:


guiclient/guiclient.h
```diff
@@ -83,6 +83,11 @@ class XWidget;
 
 // Possible return values from submitReport
 #define cNoReportDefinition   -2
+#ifdef __APPLE__
+#define SETUPMENUITEM "&xSetup"
```
gilmoskowitz: - Why `__APPLE__` instead of `Q_OS_MAC`?
- There's no reason to change the accelerator key, so this could be `x&Setup` or `&Setup Module`

guiclient/menuProducts.cpp
```diff
@@ -222,9 +218,6 @@ menuProducts::menuProducts(GUIClient *Pparent) :
   { "pd.reassignClassCodeByClassCode", tr("Reassign &Class Codes..."), SLOT(sReassignClassCodeByClassCode()), utilitiesMenu, "MaintainItemMasters", NULL, NULL, true , NULL },
   { "pd.reassignProductCategoryByProductCategory", tr("&Reassign Product Categories..."), SLOT(sReassignProductCategoryByProductCategory()), utilitiesMenu, "MaintainItemMasters", NULL, NULL, true , NULL },
 
-  // Setup
-  { "pd.setup",	    tr("&Setup..."),	  SLOT(sSetup()),     mainMenu,	"true",	NULL, NULL,  true, NULL}
-
```
gilmoskowitz: This will remove the Setup item from the Products menu.

guiclient/guiclient.h
gpazo: for some reason Q_MAC_OS was not defined in my qt creator and caused problems, __APPLE__ is defined on the compiler level for GCC, we could possibly consider replacing all instances of Q_OS_MAC with __APPLE__ in the future, note: did not see this behavior with Q_WS instances we previously used in 4.8

guiclient/guiclient.h
```diff
@@ -83,6 +83,11 @@ class XWidget;
 
 // Possible return values from submitReport
 #define cNoReportDefinition   -2
+#ifdef Q_MAC_OS
+#define SETUPMENUITEM "&xSetup"
+#else
+#define SETUPMENUITEM "x&Setup"
```
gilmoskowitz: The string `Setup` is only magic on Macs. This should be plain old `&Setup` on Windows and Linux.

PR #756, opened 2015-06-25
Issue #25127:add saletype

garyhgohoos:


gilmoskowitz:
The code changes here and in xtuple/xtuple#2380 look fine but how do they relate to issue #25125 (problem with temporary items during copy operation)?

garyhgohoos:
sorry, should be issue #25127

PR #759, opened 2015-06-29
Issue #25706: Replace ordercluster lock with AppLock

andersdd:


widgets/ordercluster.cpp
```diff
@@ -294,7 +295,6 @@ OrderLineEdit::OrderLineEdit(QWidget *pParent, const char *pName) :
   _toPrivs=false;
   _fromPrivs=false;
   _custid = -1;
-  _lock = false;
```
gilmoskowitz: `_locked` isn't initialized. For that matter, what does the `_locked` member really accomplish?

andersdd: Good question.  I was not sure what the lockSelected() and setLockSelected() functions are used for which references _locked.

gilmoskowitz: Found it. The clue is in the 3 places `setLockSelected` is called, e.g. [`issueToShipping`](https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/issueToShipping.cpp#L48). The `_lock` member tells the widget whether it should or should not obtain a lock on the record. Therefore, we should keep `_lock` but we don't need `_locked`.

andersdd: _lock was a boolean whereas the new _lock is type AppLock.  This throws an error when compiling.  So we probably need both but need to correctly set _locked to true/false

gilmoskowitz: ...and change its name - `_lockOnSelect`?

andersdd: That makes sense

andersdd:
@gilmoskowitz Changes made to orderCluster widget as discussed.  Appears to be working in different UAT scenarios

widgets/ordercluster.cpp
```diff
@@ -723,28 +724,20 @@ void OrderLineEdit::setId(const int pId, const QString &pType)
       setAllowedType(pType);
     silentSetId(pId);
     // Attempt to obtain an application lock on the order
-    if (_lock && _id != -1)
+    if (_id != -1)
```
gilmoskowitz: This should honor `_lockOnSelect`.

widgets/ordercluster.cpp
```diff
@@ -889,23 +882,9 @@ void OrderLineEdit::silentSetId(const int pId)
...
-      lock.bindValue(":table","pohead");
-    else if (isSO())
-      lock.bindValue(":table","cohead");
-    else if (isRA())
-      lock.bindValue(":table","rahead");
-    else if (isTO())
-      lock.bindValue(":table","tohead");
-    lock.exec();
-  }
+  if (! _lock.release())
```
gilmoskowitz: This should honor `_lockOnSelect`.

PR #766, opened 2015-07-02
add back removed single slash, caused compile error on mac build

gpazo:
do not merge yet, still testing.


gpazo:
mergable, compiled on all 3 OS, tested on mac and windows.

PR #768, opened 2015-07-06
Issue #25963: Add AP Open Item report and hooks into application

andersdd:
Requires xtuple/xtuple#2395

TODO Add in printing hooks for xtConnect

### DIFF
--- a/guiclient/apOpenItem.cpp
+++ b/guiclient/apOpenItem.cpp
@@ -14,6 +14,7 @@
 #include <QSqlError>
 #include <QVariant>
 
+#include "errorReporter.h"
 #include "printApOpenItem.h"
 #include "taxDetail.h"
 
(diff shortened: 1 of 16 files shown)


andersdd:
AP Credit Memo function returns different value to the AP Debit Memo function.  Added handling for this scenario.

guiclient/apOpenItem.cpp
```diff
@@ -317,9 +322,44 @@ void apOpenItem::sSave()
...
+  {
+    if(_printOnPost->isChecked())
+      sPrintOnPost(_apopenid);
     done(_apopenid);
+  }
   else
+  {
+    if(_printOnPost->isChecked())
+    {
+      if (_docType->currentIndex() == 0)
```
gilmoskowitz: This is inherited from the prior version but is also extremely fragile. It would be better to replace the hard-coded entries in `_docType` with a populate-by-select in the constructor. Then this line could use `code()` and `setCode()` to manipulate the combobox symbolically instead of depending on magic numbers.

guiclient/displays/dspAPOpenItemsByVendor.cpp
```diff
@@ -155,6 +156,10 @@ void dspAPOpenItemsByVendor::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem *select
       menuItem->setEnabled(_privileges->check("EditAPOpenItem"));
     }	
   }
+
+  menuItem = pMenu->addAction(tr("Print..."), this, SLOT(sPrintItem()));
+  if (!_privileges->check("EditAPOpenItem"))
+    menuItem->setEnabled(false);
```
gilmoskowitz: `EditAPOpenItem`, `ViewAPOpenItem`, or either? If "either", note that one call to `check()` will suffice in 4.9.0 and later.

guiclient/printApOpenItem.h
```diff
@@ -0,0 +1,36 @@
+/*
+ * This file is part of the xTuple ERP: PostBooks Edition, a free and
+ * open source Enterprise Resource Planning software suite,
+ * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
```
gilmoskowitz: `2015`

andersdd:
@gilmoskowitz Converted both AR/AR Open Items away from combobox index to code()

PR #769, opened 2015-07-06
Issue #26038: Customer selector for cash rcpt

bartmaj:


guiclient/cashReceipt.cpp
```diff
@@ -507,15 +483,18 @@ void cashReceipt::sApplyToBalance()
   XSqlQuery applyToBal;
 
   applyToBal.prepare( "UPDATE cashrcpt "
-             "SET cashrcpt_cust_id=:cust_id, "
+             "SET cashrcpt_=:cust_id, "
```
gilmoskowitz: Incomplete column name `cashrcpt_`

guiclient/cashReceipt.ui
```diff
@@ -1313,22 +1229,6 @@ to be bound by its terms.</comment>
    </hints>
   </connection>
   <connection>
-   <sender>_cust</sender>
-   <signal>valid(bool)</signal>
-   <receiver>_newCC</receiver>
-   <slot>setEnabled(bool)</slot>
```
gilmoskowitz: Is `_newCC` enabled in the .cpp when a specific customer gets selected?

widgets/customerselector.cpp
```diff
@@ -22,6 +24,9 @@ CustomerSelector::CustomerSelector(QWidget *pParent, const char *pName) : QWidge
 
   setSizePolicy(QSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed));
 
+  _allowedStates = 0;
+  populate(All + Selected + SelectedGroup + SelectedType + TypePattern);
```
gilmoskowitz: For backwards compatibility this should probably be just `Selected | SelectedType | TypePattern`.

widgets/customerselector.h
```diff
@@ -22,18 +22,26 @@ class XTUPLEWIDGETS_EXPORT CustomerSelector : public QWidget, public Ui::Custome
...
   Q_ENUMS(CustomerSelectorState)
-
+  Q_FLAGS(CustomerSelectorState)
   Q_PROPERTY(enum CustomerSelectorState state READ state WRITE setState)
 
+
   public:
+    enum CustomerSelectorState
+    {
+      All = 0x1 , Selected = 0x2, SelectedGroup = 0x4, SelectedType = 0x8,  TypePattern = 0x10
```
gilmoskowitz: Is there an advantage to `All = 0x1` as opposed to either 0 or `0x2 + 0x4 + 0x8 + 0x10`?

widgets/customerselector.cpp
```diff
@@ -29,8 +34,11 @@ CustomerSelector::CustomerSelector(QWidget *pParent, const char *pName) : QWidge
   _customerGroup->setId(-1);
   _customerType->setText("");
 
+
+
   connect(_select,SIGNAL(currentIndexChanged(int)), this, SIGNAL(updated()));
   connect(_select,SIGNAL(currentIndexChanged(int)), this, SIGNAL(newState(int)));
+  connect(_select, SIGNAL(currentIndexChanged(int)), this, SLOT(setStackElement()));
```
gilmoskowitz: Is `setStackElement()` redundant? Can this work be handled in `newState()`?

widgets/customerselector.cpp
bartmaj: But then there wouldn't be anything referring to "Customer Group"...

widgets/customerselector.cpp
bartmaj: `setStackElement()` controls `_selectStack`, while `newState()` controls `_select`. I can certainly merge them...

gilmoskowitz: Merge them if `_selectStack` and `_select` must be kept in sync. If there are cases where they should not be synchronized, keep the two methods separate.

widgets/customerselector.cpp
gilmoskowitz: Right. The default state of the revised widget should be backwards-compatible with the old version of the widget, which doesn't know about Customer Groups. We need to distinguish between the full range of capability this widget is capable of providing and what subset it'll allow in any particular use case.

PR #774, opened 2015-07-07
Issue #18550:make getDate script invokable

garyhgohoos:


guiclient/xdateinputdialog.h
```diff
@@ -25,7 +25,8 @@ class XDateInputDialog : public XDialog, public Ui::XDateInputDialog
 public:
     XDateInputDialog(QWidget* parent = 0, const char* name = 0, bool modal = false, Qt::WindowFlags fl = 0);
     ~XDateInputDialog();
-    QDate getDate();
+//    QDate getDate();
```
gilmoskowitz: We don't need the old version in the comment. We can use git for that.

PR #777, opened 2015-07-14
Issue #20282:combine vendor and vendorworkbench

garyhgohoos:


widgets/vendorcluster.cpp
```diff
@@ -29,26 +49,239 @@ VendorLineEdit::VendorLineEdit(QWidget *pParent, const char *pName) :
...
+
+  emit editable(p);
+  return p;
+}
+
+void VendorLineEdit::sParse()
+{
+// TODO - don't understand why
+//  if (_editMode)
+//    return;
```
gilmoskowitz: Re: `TODO` - could it be that `return` here allows entering a previously unused vendor number?

widgets/vendorcluster.h
```diff
@@ -11,29 +11,107 @@
...
+class QDropEvent;
+class QMouseEvent;
+
+//class VendInfo;
+
+#define __allVendors    0x01
+#define __activeVendorss 0x02
+
+#define CRMACCT_ID      6
+#define ISVENDOR        7
```
gilmoskowitz: These are pretty far removed from where they're used and they only appear to be used once each. There's got to be a better way.

widgets/vendorcluster.cpp
```diff
@@ -29,26 +49,239 @@ VendorLineEdit::VendorLineEdit(QWidget *pParent, const char *pName) :
...
+  if (model() && _id != -1)
+  {
+    setUiName("vendor");
+    setEditPriv("MaintainVendors");
+    setViewPriv("ViewVendors");
+    setNewPriv("MaintainVendors");
+    _idColName="vend_id";
+    sUpdateMenu();
+
+    _crmacctId = model()->data(model()->index(0,CRMACCT_ID)).toInt();
```
gilmoskowitz: There should be a way to derive the column number from the model and the column _name_. Using hard-coded constants like this is fragile and caused us trouble with characteristics handling.

garyhgohoos: used custCluster as a template.  don't know how to improve.

PR #781, opened 2015-07-15
Issue #25576: Reconnect automatically after lost db connection

bartmaj:


guiclient/guiclient.cpp
```diff
@@ -1023,21 +1023,31 @@ void GUIClient::sTick()
...
-    systemError(this, tr("<p>You have been disconnected from the database server.  "
-                          "This is usually caused by an interruption in your "
-                          "network.  Please exit the application and restart."
-                          "<br><pre>%1</pre>" )
-                      .arg(tickle.lastError().databaseText()));
+    if (!QSqlDatabase::database().isOpen())
+    {
+      if (QSqlDatabase::database().open())
+      {
+        qDebug() << "Reconnected to the database";
```
davidbeauchamp: You can see an example of how we currently run the login() database function [here](https://github.com/xtuple/qt-client/blob/4_10_x/common/login2.cpp#L479-L486)

bendiy: This needs to run through most of the logic in https://github.com/xtuple/qt-client/blob/4_10_x/common/login2.cpp to call `SELECT getEffectiveXtUser() AS user;` and setup search paths, etc.

davidbeauchamp: That stuff should happen fine in login() (the search path, etc). The call to getEffectiveXtuser() is to set some C++ variables which should still be set, it is using the same QSqlDatabase it had before.. unless you think getEffectiveXtUser() will change upon reconnection?

bendiy: It will need to know if `_setSearchPath` is `true` or `false`. `geteffectivextuser()` needs to be called so `XT.username` is set on the connection for any plv8 functions.
https://github.com/xtuple/xtuple/blob/4_10_x/database/source/public/functions/geteffectivextuser.sql#L8

bendiy: `login()` actually calls `geteffectivextuser()` here, so it's not needed.
https://github.com/xtuple/xtuple/blob/4_10_x/foundation-database/public/functions/login.sql#L48

I'm not sure why we do it twice in `login2.cpp`.

PR #785, opened 2015-07-22
Issue #17543: Add "copy column" to XTreeWidget

bartmaj:


widgets/xtreewidget.cpp
```diff
@@ -2278,6 +2279,38 @@ void XTreeWidget::sCopyVisibleToClipboard()
...
+      }
+      opText = opText + line + "\r\n";
+      idx    = indexBelow(idx);
+    }
+  }
+  text.setText(opText);
+  if (_x_preferences->boolean("CopyListsPlainText"))
+    mime->setText(text.toPlainText());
+  else
+    mime->setHtml(text.toHtml());
```
davidbeauchamp: Should we be setText() and setHtml()' in BOTH sections of this conditional? Wondering if it will stem the confusion people have about having the preference set to Rich text and wondering why they can't paste into notepad. I am unsure of the intricacies of doing this though I know I've seen places where clipboard modification will set both. @gilmoskowitz thoughts?

gilmoskowitz: If we wrote both MIME types to the clipboard, we could probably improve predictability, reduce code redundancy, and get rid of the metric. It sounds like a good idea to try in this function first, then in the others that handle cell, row, and all-visible.

In case it helps with testing, I can paste HTML content copied from an XTreeWidget into MS Word, Apple's TextEdit, and OpenOffice Text and Spreadsheet documents, but not Apple's Terminal or mvim (Mac port of gvim).

widgets/xtreewidget.cpp
```diff
@@ -2278,6 +2279,38 @@ void XTreeWidget::sCopyVisibleToClipboard()
   clipboard->setMimeData(mime);
 }
 
+void XTreeWidget::sCopyColumnToClipboard()
+{
+  QTextEdit       text;
+  QMimeData       *mime      = new QMimeData();
+  QClipboard      *clipboard = QApplication::clipboard();
+  QString line;
+  QString opText = line + "\r";
```
gilmoskowitz: Why start with `\r`?

widgets/xtreewidget.cpp
```diff
@@ -2278,6 +2279,38 @@ void XTreeWidget::sCopyVisibleToClipboard()
...
+  QMimeData       *mime      = new QMimeData();
+  QClipboard      *clipboard = QApplication::clipboard();
+  QString line;
+  QString opText = line + "\r";
+
+  XTreeWidgetItem *item = topLevelItem(0);
+  if (item)
+  {
+    QModelIndex idx = indexFromItem(item);
+    while (idx.isValid())
```
gilmoskowitz: Classic for-loop

widgets/xtreewidget.cpp
```diff
@@ -2278,6 +2279,38 @@ void XTreeWidget::sCopyVisibleToClipboard()
...
+    {
+      item = (XTreeWidgetItem *)itemFromIndex(idx);
+      if (item)
+      {
+        line = "";
+        line = line + item->text(currentColumn()) + "\t";
+      }
+      opText = opText + line + "\r\n";
```
gilmoskowitz: if `! item` in this iteration, `line` will still hold the value from the previous iteration and `optText` will get two copies of it.

PR #786, opened 2015-07-23
Issue #22413: Removes temporary update file

bartmaj:


bartmaj:
I moved the the code responsible for deleting the temporary download file so that it executes unconditionally.

common/checkForUpdates.cpp
```diff
@@ -252,45 +252,32 @@ void checkForUpdates::cancelDownload()
...
+  {
+    file->close();
+    file->remove();
     delete file;
     file = 0;
+  }
 
-  if (DEBUG) qDebug() << "downloadFinished() returning";
+  if (downloadRequestAborted)
+    return;
```
gilmoskowitz: Please leave the `if (DEBUG)` here. Also, the conditional `return` is not necessary at the end of the function.

PR #788, opened 2015-07-24
Mac dock menu window list

mboman28:


guiclient/xdialog.cpp
```diff
@@ -186,3 +190,12 @@ ParameterList XDialog::get() const
...
 
+int XDialog::exec()
+{
+	#ifdef Q_OS_MAC
+		omfgThis->updateMacDockMenu(this);
+	#endif
+
+	QDialog::exec();
+}
+
```
gilmoskowitz: Do we need similar code in XMainWindow and XWidget?

mboman28: XWidgets are opened with handleNewWindow(), so they are handled there.

I didn't change anything in XMainWindow because I couldn't find anywhere where it is actually used to create a window, but if that is something that is going to be added, then yes, the updateMacDockMenu() function will need to be called from there when the window is being opened.

PR #793, opened 2015-08-03
Issue #26258:allow entering of customer number in workbench mode

garyhgohoos:


guiclient/customer.cpp
```diff
@@ -364,7 +364,8 @@ void customer::setViewMode()
 {
   _mode = cView;
 
-  _number->setEnabled(false);
+//  _number->setEnabled(false);
+  _number->setEditMode(true);
```
gilmoskowitz: In b6061c49f this was changed from `_number->setCanEdit(FALSE)` to `_number->setEnabled(FALSE)`. Apparently `setEnabled()` caused the regression, but is the fix to call `setCanEdit()` or `setEditMode()`?

PR #795, opened 2015-08-03
Issue #26148:add refresh order checkbox

garyhgohoos:


guiclient/purchaseOrderItem.ui
```diff
@@ -1206,7 +1222,7 @@ to be bound by its terms.</comment>
                 <enum>Qt::ClickFocus</enum>
                </property>
                <property name="type">
-                <enum>Comments::PurchaseOrderItem</enum>
+                <number>0</number>
```
gilmoskowitz: This change has to be undone by hand; Qt Designer will break it. There's a known bug in the Comments widget causing this now that we can associate comments with arbitrary document types.

PR #801, opened 2015-08-07
Issue #23152: Allow to edit/create shared filters based on a privilege

bartmaj:
Database changes at: xtuple/xtuple#2439

### DIFF
--- a/widgets/parameterwidget.cpp
+++ b/widgets/parameterwidget.cpp
@@ -1466,7 +1466,8 @@ void ParameterWidget::save()
   params.append("classname", classname);
   if (_shared)
     params.append("shared", true);
-
+  if (!_x_privileges->check("AllowSharedFilterEdit"))
+    params.append("disableshare", true);
   filterSave newdlg(this);
   newdlg.set(params);
   filter_id = newdlg.exec();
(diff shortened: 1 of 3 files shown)


widgets/parameterwidget.cpp
```diff
@@ -115,6 +116,7 @@ ParameterWidget::ParameterWidget(QWidget *pParent, const char *pName)  :
   connect(_manageButton,              SIGNAL(clicked()), this, SLOT(sManageFilters()));
   connect(_saveButton,                SIGNAL(clicked()), this, SLOT(save()));
   connect(this,                       SIGNAL(updated()), this, SLOT(toggleSave()));
+  connect(_manageButton,              SIGNAL(clicked()), this, SLOT(sDebug()));
```
gilmoskowitz: There is no `sDebug()` slot so this `connect()` will fail (and write spurious debug output:-)

bartmaj: Oopsie daisies!

widgets/parameterwidget.cpp
```diff
@@ -41,6 +41,7 @@
 #include "shiptocluster.h"
 #include "ordercluster.h"
 #include "wocluster.h"
+#include "../guiclient/guiclient.h"
```
gilmoskowitz: A forward reference to `guiclient.h` is not good and doesn't appear necessary here.

PR #804, opened 2015-08-11
Issue #17067: Allow Shipping Forms to print T/O shipments

andersdd:


guiclient/printShippingForms.cpp
```diff
@@ -39,19 +38,38 @@ printShippingForms::printShippingForms(QWidget* parent, const char* name, bool m
              "       shipform_report_name      AS reportname,"
              "       shiphead_id, shiphead_shipchrg_id"
              "  FROM shiphead"
-             "     JOIN cohead   ON (shiphead_order_id=cohead_id"
-	     "                   AND shiphead_order_type='SO')"
+             "     JOIN cohead   ON (shiphead_order_id=cohead_id)"
```
gilmoskowitz: Each half of the UNION must restrict its join based on the order type. Otherwise overlap in cohead_id and tohead_id values may lead to spurious rows in the output.

andersdd: The Github screen looks a bit confusing, but I have updated the file.

PR #805, opened 2015-08-11
Issue #26302:save new comment type before granting sources

garyhgohoos:


gilmoskowitz:
@garyhgohoos - could you please backport this to 4_9_x?

PR #808, opened 2015-08-12
Issue #26319:add POSupply and WOSupply warehouse types and enforce wh…

garyhgohoos:
…en creating PO and WO


andersdd:
+1 code review.  Is this worth backporting to 4.9.1?

PR #815, opened 2015-08-17
Issue #26224:add qty reserved

garyhgohoos:


gilmoskowitz:
requires xtuple/xtuple#2453 and see xtuple/private-extensions#682

PR #846, opened 2015-08-28
Issue #24158: Add RMA field to PO Returns

andersdd:
Add Vendor Authorisation #  (RMA) to PO Returns

requires xtuple/xtuple#2479

### DIFF
--- a/guiclient/enterPoitemReturn.cpp
+++ b/guiclient/enterPoitemReturn.cpp
@@ -156,7 +156,7 @@ enum SetResponse enterPoitemReturn::set(const ParameterList &pParams)
 
   param = pParams.value("poreject_rma", &valid);
   if (valid)
-    _rma = param.toString();
+    _rmAuthority = param.toString();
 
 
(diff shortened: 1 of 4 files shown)


guiclient/enterPoitemReturn.cpp
```diff
@@ -154,6 +154,11 @@ enum SetResponse enterPoitemReturn::set(const ParameterList &pParams)
     }
   }
 
+  param = pParams.value("poreject_rma", &valid);
+  if (valid)
+    _rma = param.toString();
+
```
gilmoskowitz: There's a conflict between `_rma` the XLineEdit widget in the .ui file and `_rma` the QString as declared in the .h and used in the .cpp files. Use the widget everywhere.

andersdd: Interesting.  I did not get any compile or runtime errors with this.  The variable and widget are on separate screens.  Easy enough to resolve.

PR #852, opened 2015-08-31
Issue #26346:clear _location and _zone before filling

garyhgohoos:


guiclient/locations.cpp
```diff
@@ -170,6 +170,8 @@ void locations::sPrint()
 
 void locations::sFillList()
 {
+  _location->clear();
+
```
gilmoskowitz: Instead of adding `clear()` here, remove the `if (query.first())` later. See [XTreeWidget::populate()](https://github.com/xtuple/qt-client/blob/4_10_x/widgets/xtreewidget.cpp#L250-L254) and its [defaults](https://github.com/xtuple/qt-client/blob/4_10_x/widgets/xtreewidget.h#L179).

PR #854, opened 2015-08-31
Issue #17846:include child crmaccts when listing comments and contacts

garyhgohoos:


gilmoskowitz:
accompanies xtuple/xtuple#2482

PR #865, opened 2015-09-09
Issue #26376:disable new/next button in new mode, save starts new lin…

garyhgohoos:
…e item


gilmoskowitz:
This looks incorrect. It changes the workflow and fails to handle quote line items. Is there a reason not to unwind partially saved work in `salesOrderItem::sNext()`?

garyhgohoos:
I fixed the problem with quote line items.  I think this new workflow reduces redundancy.  Why have two routines to unwind cancellations?

PR #872, opened 2015-09-13
Issue #20135: VAT Taxation enhancements

andersdd:
Initial commit of VAT taxation enhancements.  These changes stand on their own but further changes required for PO Freight Taxation.   I've added a Tax Return report - but not entirely convinced they are sufficiently different from Tax History to warrant the additional screen / report.

requires xtuple/xtuple#2495


guiclient/dspTaxReturn.cpp
```diff
@@ -0,0 +1,246 @@
...
+  if (valid)
+    _purchases->setChecked(param.toBool());
+  
+  param = pParams.value("startDate", &valid);
+  if (valid)
+    _dates->setStartDate(param.toDate());
+    
+  param = pParams.value("endDate", &valid);
+  if (valid)
+    _dates->setStartDate(param.toDate());
```
gilmoskowitz: `_dates->setEndDate()`?

guiclient/miscCheck.cpp
```diff
@@ -394,3 +439,30 @@ void miscCheck::sCreditMemoSelected()
...
+  calcTax.exec();
+  if (ErrorReporter::error(QtCriticalMsg, this,
+                                tr("Tax Calculation"),
+                                calcTax, __FILE__, __LINE__))
+    return; 
+  else
+  {
+    if (calcTax.first())
+      _tax->setLocalValue(calcTax.value("tax").toDouble());
+  }
```
gilmoskowitz: What's the rationale for inverting the normal flow of
```
if (q.first()) {
  doStuff;
} else if (ErrorReporter::error(...)) {
  return;
}
```
?

andersdd:
@gilmoskowitz had an epiphany yesterday about the Misc Voucher and our discussions about tax being applicable or not depending on the tax type.  I have made some assumptions in this current code, but think I can improve the logic and flexibility with some minor changes.  These changes also work with the freight tax subsequent work - which is where the idea came from.

gilmoskowitz:
@andersdd - OK. Let me know when I should review this again.

PR #875, opened 2015-09-15
issue #26462 - update postgres version checks

gilmoskowitz:
Merge after testing with PG 9.4. If that goes well, we can also mark issue #24342 resolved.


andersdd:
+1 code review

PR #876, opened 2015-09-16
Another batch of programs for errorReporterRefactor

jrosengarden:


guiclient/apOpenItem.cpp
```diff
@@ -490,12 +490,12 @@ void apOpenItem::populate()
     populateOpenItem.bindValue(":apopen_id", _apopenid);
     populateOpenItem.bindValue(":other", tr("Other"));
     populateOpenItem.exec();
-    if (populateOpenItem.lastError().type() != QSqlError::NoError)
-	systemError(this, populateOpenItem.lastError().databaseText(), __FILE__, __LINE__);
+    if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Information"),
+                                  populateOpenItem, __FILE__, __LINE__))
```
gilmoskowitz: The `if(ErrorReporter::error()` will cause the `_apapply->populate()` to be called _only_ if there was an error running the query. The old code apparently had a redundant error check, so just remove 493-494 instead of updating them.

guiclient/apOpenItem.cpp
```diff
@@ -584,10 +584,10 @@ void apOpenItem::sTaxDetail()
     ap.bindValue(":docNumber", _docNumber->text());
     ap.bindValue(":currId", _amount->id());
     ap.exec();
-    if (ap.lastError().type() != QSqlError::NoError)
+    if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Setting Tax Amounts"),
+                                  ap, __FILE__, __LINE__))
```
gilmoskowitz: This is an exceptional case - the `ap` query is part of a multistep transaction. Please leave the `if` as it was and simply replace `systemError()` with `ErrorReporter::error()`. That'll avoid holding the transaction open while waiting for user acknowledgement.

guiclient/applyARCreditMemo.cpp
```diff
@@ -110,21 +111,21 @@ void applyARCreditMemo::sPost()
...
-      systemError(this, storedProcErrorLookup("postARCreditMemoApplication", result),
-		  __FILE__, __LINE__);
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting A/R CM"),
+                                      applyPost, __FILE__, __LINE__);
       return;
     }
   }
-  else if (applyPost.lastError().type() != QSqlError::NoError)
+  else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting A/R CM"),
+                                applyPost, __FILE__, __LINE__))
```
gilmoskowitz: _Here's_ a user dialog inside a transaction... (see the rollback 2 lines down?-) Please leave the `lastError()` check in the `if` statement and move the `ErrorReporter::error()` to follow the rollback.

PR #877, opened 2015-09-16
Issue #26528:synchronize inventory and price uom

garyhgohoos:


guiclient/item.cpp
```diff
@@ -845,9 +845,9 @@ void item::sSave()
   itemSave.exec();
   if (itemSave.lastError().type() != QSqlError::NoError)
   {
+    systemError(this, itemSave.lastError().databaseText(), __FILE__, __LINE__);
     itemSave.exec("ROLLBACK;");
     _inTransaction = false;
-    systemError(this, itemSave.lastError().databaseText(), __FILE__, __LINE__);
```
gilmoskowitz: This will hold the transaction open until the user response to the `systemError()`.

gilmoskowitz:
More generally, this might not even be a bug. See the comments in the issue tracker.

garyhgohoos:
I agree with Zenq that this is a bug

PR #880, opened 2015-09-21
issue #26334 - fix redirect

gpazo:


guiclient/main.cpp
```diff
@@ -459,24 +466,31 @@ int main(int argc, char *argv[])
       {
         db = metric.value("db").toString();
       }
-
-#if QT_VERSION < 0x050000  // below removed in qt5, needs to be ported
-      QHttp *http = new QHttp();
-      
+#if QT_VERSION >= 0x050000
```
gilmoskowitz: Is there a reason these changes are specific to Qt 5? Just use the new code unless Qt4 won't support it.

guiclient/main.cpp
```diff
@@ -99,8 +99,14 @@
...
 #if QT_VERSION < 0x050000
 #include <QHttp>
+#else
+#include <QUrlQuery>
 #endif
 #include <QUrl>
+#include <QDebug>
+#include <QNetworkAccessManager>
+#include <QNetworkRequest>
+#include <QNetworkReply>
```
gilmoskowitz: We don't use all of these classes below so don't need the includes here.

guiclient/main.cpp
gpazo: right, qt4 won't support QUrlQuery but it's needed for qt5 support. Once it's passed into the xtNetworkRequest engine it is supported by both qt4 and qt5

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+void xtNetworkRequestManager::sslErrors(QNetworkReply*, const QList<QSslError> &errors) {
+#if QT_VERSION >= 0x050000
+    QString errorString;
+       foreach (const QSslError &error, errors) {
+           if (!errorString.isEmpty())
+               errorString += ", ";
+           errorString += error.errorString();
+       }
+   if(DEBUG)
```
gilmoskowitz: Always `qDebug()` this so we know things went wrong by looking at the database log window. We might need a way to tell the user but not in this pass.

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+#if QT_VERSION < 0x050000 //this feels hackish, ignore error or provide root cert, we'll ignore for now
+    connect(nwam, SIGNAL(sslErrors(QNetworkReply*,QList<QSslError>)), _nwrep, SLOT(ignoreSslErrors()));
+#endif
+    connect(nwam, SIGNAL(finished(QNetworkReply*)), _loop, SLOT(quit()));
+    _loop->exec();
+}
+void xtNetworkRequestManager::requestCompleted() {
+  _response = _nwrep->readAll(); //we don't really care here but store it anyways
+  _nwrep->close();
+  QVariant possibleRedirect = _nwrep->attribute(QNetworkRequest::RedirectionTargetAttribute);
```
gilmoskowitz: http://doc.qt.io/qt-5/qnetworkrequest.html -
"...The returned URL might be relative. Use QUrl::resolved() to create an absolute URL out of it."

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if(DEBUG){
+      qDebug() << "redirect=" << possibleRedirect;
+      qDebug() << "replyError=" << _nwrep->errorString();
+      qDebug() << "replyErrorCode=" << _nwrep->error();
+  }
+  if(_nwrep->error() == QNetworkReply::NoError){
+      _nwrep->deleteLater();
+      _mutex->unlock();
+  }
+  else if(!possibleRedirect.isNull()){
```
gilmoskowitz: http://doc.qt.io/qt-5/qnetworkreply.html -
"Note: When the HTTP protocol returns a redirect no error will be reported. You can check if there is a redirect with the QNetworkRequest::RedirectionTargetAttribute attribute."

gilmoskowitz: The order of the `if-else-if` still looks wrong. If there's a redirect, no error will be reported so we'll `deleteLater` the QNetworkReply and unlock the mutex. We won't fall into the redirection handling code.

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  _mutex->lock();
+  _loop = new QEventLoop;
+  startRequest(_url);
+}
+void xtNetworkRequestManager::startRequest(const QUrl & url) {
+    _nwrep = nwam->get(QNetworkRequest(url));
+    connect(_nwrep, SIGNAL(finished()), SLOT(requestCompleted()));
+    connect(nwam, SIGNAL(sslErrors(QNetworkReply*,QList<QSslError>)), this, SLOT(sslErrors(QNetworkReply*,QList<QSslError>)));
+#if QT_VERSION < 0x050000 //this feels hackish, ignore error or provide root cert, we'll ignore for now
+    connect(nwam, SIGNAL(sslErrors(QNetworkReply*,QList<QSslError>)), _nwrep, SLOT(ignoreSslErrors()));
```
gilmoskowitz: `ignoreSslErrors()` isn't defined anywhere.

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  if(_nwrep->error() == QNetworkReply::NoError && !possibleRedirect.isValid()){
+      //success and no redirect
+      _nwrep->deleteLater();
+      _mutex->unlock();
+  }
+  else {
+      QUrl newUrl = _url.resolved(possibleRedirect.toUrl());
+      _nwrep->deleteLater();
+      startRequest(newUrl);
+  }
```
gilmoskowitz: The revisions still don't handle `_newrep->error() != QNetworkReply::NoError`. There are 2 booleans here, so 4 possible cases, but only two branches in the code - _warning_!

common/xtNetworkRequestManager.cpp
```diff
@@ -0,0 +1,77 @@
...
+void xtNetworkRequestManager::sslErrors(QNetworkReply*, const QList<QSslError> &errors) {
+#if QT_VERSION >= 0x050000
+    QString errorString;
+       foreach (const QSslError &error, errors) {
+           if (!errorString.isEmpty())
+               errorString += ", ";
+           errorString += error.errorString();
+       }
+   qDebug() << "errorString= " << errorString;
+#endif
```
gilmoskowitz: This doesn't look specific to Qt5. Why `#if`/`#endif`?

common/xtNetworkRequestManager.cpp
gpazo: It's a hack-ish way around the SSL error encountered on the https version of the site, ignore the error and don't stop execution, rather than providing a proper cert for the connection(until we implement a lasting solution for qt5)

gilmoskowitz: Oh, I see: http://doc.qt.io/qt-4.8/qnetworkaccessmanager.html#sslErrors

PR #881, opened 2015-09-21
issue #26515 - don't bug the user as often about CVV

gilmoskowitz:


guiclient/creditcardprocessor.cpp
```diff
@@ -1595,14 +1595,13 @@ int CreditCardProcessor::checkCreditCard(const int pccid, const QString &pcvv, Q
     _errorMsg = errorMsg(-98);
     return -98;
   }
-  else if (pcvv.isEmpty())
+  else if (pcvv.isEmpty() && _metrics->value("CCCVVCheck").contains(QRegExp("^[FW]$")))
```
andersdd: How does this global metric check work if there are multiple users processing CCs simultaneously?

gilmoskowitz: There are no conflicts, as the metrics are loaded at application startup. Like most metrics it doesn't change often. See https://github.com/xtuple/qt-client/blob/4_9_x/guiclient/configureCC.cpp#L115-L122 and https://github.com/xtuple/qt-client/blob/4_9_x/guiclient/configureCC.ui#L424-L463 for context.

Is there something deeper to the question?

PR #883, opened 2015-09-22
replace macro with working __APPLE__ macro

gpazo:


guiclient/guiclient.h
```diff
@@ -83,7 +83,7 @@ class XWidget;
 
 // Possible return values from submitReport
 #define cNoReportDefinition   -2
-#ifdef Q_MAC_OS
+#ifdef __APPLE__
```
gilmoskowitz: Please use the Qt macro, which is mistyped here (should be `#ifdef Q_OS_MAC`).

PR #902, opened 2015-10-08
errorReporterRefactor:  10/08/15 Batch

jrosengarden:


guiclient/closeWo.cpp
```diff
@@ -176,25 +177,27 @@ bool closeWo::okToSave()
...
-    systemError(this, tr("Cannot close Work Order %1 because the Item does not "
-                         "appear to have an Item Type! Please check the Item "
-                         "definition. You may need to reset the Type and save "
-                         "the Item record.").arg(_wo->woNumber()));
+    ErrorReporter::error(QtCriticalMsg, this, tr("Cannot close Work Order %1 because the item does not "
+                         "appear to have an Item Type!  Please check the Item "
+                         "definition.  You may need to reset the Type and save "
+                         "the Item record.")
+                         .arg(_wo->woNumber()),
+                         type, __FILE__, __LINE__);
```
gilmoskowitz: This `::error()` call won't work because the `type` query didn't have an error (that case would have been caught at line 188). Instead, get rid of the `type` argument and insert a short `tr()` title argument between `this` and the long `tr()`. That'll force display of the long message with a short title for the dialog.

openrpt
```diff
@@ -1 +1 @@
-Subproject commit 32fdfe9a2d063d356656478a06886ff52a4d4dd7
+Subproject commit ff241369764b1639f1b28d3e45b15ac8cdf328e2
```
gilmoskowitz: This change should not be committed. We need to update with an xtuple/openrpt commit hash (PR for this is coming soon).

PR #904, opened 2015-10-09
issue #23207 - adds additional filters to gltrans

mccjim:


davidbeauchamp:
Requires https://github.com/xtuple/xtuple/pull/2520

guiclient/displays/dspGLTransactions.cpp
```diff
@@ -265,8 +278,16 @@ bool dspGLTransactions::setParams(ParameterList &params)
...
       XSqlQuery begq;
-      begq.prepare("SELECT "
+
+      MetaSQLQuery mql = mqlLoad("gltransactions", "begq");
+  //    ParameterList params;
+  //    params.append("accnt_id", )
+
+      begq = mql.toQuery(params);
+
+/*      begq.prepare("SELECT "
```
gilmoskowitz: The `select` has been moved to MetaSQL so just remove it from here. We can always go back in the revision history to see it if we need to.

PR #905, opened 2015-10-09
issue #26655 add missing line, per feedback from mike

gpazo:


guiclient/voucher.cpp
```diff
@@ -592,6 +592,7 @@ void voucher::sFillList()
                "         FROM recv"
                "         WHERE ( (recv_posted)"
                "           AND (recv_invoiced)"
+               "           AND (recv_order_type='PO')"
```
gilmoskowitz: The same change probably needs to be made to the `qtyreceived` subquery as well.

PR #908, opened 2015-10-19
update openrpt to include qt5 build fix

gilmoskowitz:


andersdd:
Assuming this is the new compiled hash - then +1 code review

PR #915, opened 2015-10-21
issue #26687 fix uom conversion for reserving stock on sales order screen

gpazo:


gilmoskowitz:
@gpazo - this looks like the wrong change.

- `qtyAvailable()` returns the quantity currently at the itemsite, presumably in inventory UOM
- `qtyReserved()` returns the reserved quantity in all coitems for the current itemsite, presumably in the selling UOM. However, there is no check that the various coitem records use the same UOM.

 `qtyAvailable() - qtyReserved()` can mix units while `qtyUnreserved()` tries to account for UOM differences between coitems. Therefore the original code looks closer to correct.

Did you try changing `qtyReserved()` to normalize quantities to the same UOM? If so, _which_ UOM?

gpazo:
After looking around in the database and doing a bit of trial/error I think this fix still stands true, the UOM conversion inside of qtyUnreserved is the source of the bug, the ```coitem_qtyreserved``` is returning the correct value but when multiplying it by the UOM the conversion is essentially done twice. 

ex. TWHEEL with UOM WSET 1: 4 ratio, 1 WSET should reserve a quantity of 4 TWHEELS from stock.

when reserving 1 TWHEEL with WSET UOM you'd expect on this screen to see a quantity of 4 but instead you see 16.  The UOM conversion is already present in the ```coitem_qtyreserved``` value but is multiplied again incorrectly by qtyUnreserved function.

Adjusting the ```qtyUnreserved``` function could return unintended consequences as it is used to calculate inventory in these two database functions:
```xtuple/foundation-database/public/functions/sufficientinventorytoshipitem.sql```
```xtuple/foundation-database/public/functions/relocateinventory.sql```

PR #917, opened 2015-10-24
Issue #16649: Expired Registration key gui issue

andersdd:


andersdd:
@gilmoskowitz you might want to check the logic with this.  Previously if you entered an invalid key the system still let you log in.  With this change the Registration key dialog is now opened and a valid key is required otherwise the system will exit.

PR #918, opened 2015-10-25
errorReporterRefactor 10/25/15 Batch

jrosengarden:


guiclient/correctProductionPosting.cpp
```diff
@@ -169,10 +171,10 @@ void correctProductionPosting::sCorrect()
     correctCorrect.exec("COMMIT;");
     omfgThis->sWorkOrdersUpdated(_wo->id(), true);
   }
-  else if (correctCorrect.lastError().type() != QSqlError::NoError)
+  else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Production Correction"),
+                                correctCorrect, __FILE__, __LINE__))
```
gilmoskowitz: This instance is best left with the error check in the `if` and `ErrorReporter::error()` call inside the block after the `rollback.exec()`.

guiclient/countTag.cpp
```diff
@@ -160,7 +161,7 @@ void countTag::sEnter()
   {
 
 //  Make sure that there aren't any unposted Slips for this Tag
-    countEnter.prepare( "SELECT cntslip_id "
+    countEnter.prepare( "=SELECT cntslip_id "
```
gilmoskowitz: `=`!

guiclient/countries.cpp
```diff
@@ -116,10 +117,11 @@ void countries::sDelete()
...
     countriesDelete.exec();
-    if (countriesDelete.lastError().type() != QSqlError::NoError)
-      systemError(this, countriesDelete.lastError().databaseText(), __FILE__, __LINE__);
-  
-    sFillList();
+    ErrorReporter::error(QtCriticalMsg, this, tr("Error Deleting Country"),
+                         countriesDelete, __FILE__, __LINE__);
+    {
+      sFillList();
+    }
```
gilmoskowitz: The `{` and `}` aren't necessary because there's no `if` (a rarity).

PR #919, opened 2015-10-27
issue #26546 - show existing windows instead of creating new ones

gilmoskowitz:
for msgs that are already open.

This is an experiment. If it works then we should consider propagating the idea throughout the application. I think this is what https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/salesOrder.cpp#L3647-L3740 is trying to accomplish, for example.


davidbeauchamp:
+1

PR #920, opened 2015-10-28
issue #26833 - simple sales order entry

garyhgohoos:


guiclient/salesOrderSimple.ui
```diff
@@ -0,0 +1,1270 @@
...
+                 <property name="sizeHint" stdset="0">
+                  <size>
+                   <width>20</width>
+                   <height>20</height>
+                  </size>
+                 </property>
+                </spacer>
+               </item>
+               <item row="6" column="0">
+                <widget class="QPushButton" name="_authorize">
```
gilmoskowitz: We don't need preauthorizations here (double-check with Ciara).

guiclient/salesOrderSimple.ui
```diff
@@ -0,0 +1,1270 @@
...
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="2">
+           <widget class="ShiptoCluster" name="_shipTo">
```
gilmoskowitz: We don't need ship-tos here (double-check with Ciara).

guiclient/salesOrderSimple.ui
```diff
@@ -0,0 +1,1270 @@
...
+                 <property name="focusPolicy">
+                  <enum>Qt::NoFocus</enum>
+                 </property>
+                 <property name="text">
+                  <string>View</string>
+                 </property>
+                </widget>
+               </item>
+               <item row="3" column="0">
+                <widget class="QPushButton" name="_upCC">
```
gilmoskowitz: We don't need to edit credit card priorities here. We probably don't need to edit or view them, either. Double-check with Ciara.

gpazo:
The file is truncated and contains a majority of the changes but possible bugs in guiclient/salesOrderSimple.cpp:

line 229, 778 wrap if/else in ```{ }```

line 1160, 1392, no SQL error check

line 1455 missing allocateSales.first()?

line 202, 213, 363, 452, 473, 503, 542, 600, 715, 753, 784, 830, 850, 868, 892, 993, 1002, 1056, 1087, 1243, 1451, 1479, 1507, 1540, 1573, 1604, 1630, 1687, 1721, 1734, 1773, 1797, 1832, 1878, 1909, 1944, 1957, 1977, 2007, 2034, 2061,   use ```ErrorReporter::error``` subclass

gilmoskowitz:
The .cpp is too long to show in the web diff:-(

~line 700 - the customer P/O field should be highlighted if it's required and empty. The highlighting should be turned off if it's (required and non-empty) or not required. The highlight color should be the warning or error color, not `red`. It would be nice to explore using [Qt Stylesheets](http://doc.qt.io/qt-4.8/stylesheet-examples.html#customizing-qlineedit) rather than procedural code to selectively apply color. That might require adding some properties to XLineEdit. We can merge without this change.

gilmoskowitz:
~line 1024 - `_orderNumber->setEnabled(true);` should use the number-generation mode instead of hard-code `true`.

guiclient/salesOrderSimple.ui
garyhgohoos: I did not remove ship-to, but I changed it to optional (not required)

PR #928, opened 2015-11-05
Issue 26849: Extract Item Groups populate query to MetaSQL, provide checkbox to show only top level item groups

davidbeauchamp:


davidbeauchamp:
Requires https://github.com/xtuple/xtuple/pull/2539

andersdd:
Not 100% sure about the extra sFillList - personally think we should remove the int parameter.    But rest of code looks fine +1 code review

davidbeauchamp:
Removed duplicate sFillList(), implemented getParams(), also wrapped delete group code in a QMessageBox to confirm deletion

andersdd:
+1 code review.   Like this feature - very useful

PR #935, opened 2015-11-07
issue #26815 - itemcost issues

jrosengarden:


guiclient/maintainItemCosts.cpp
```diff
@@ -262,13 +262,18 @@ void maintainItemCosts::sEnterActualCost()
...
 void maintainItemCosts::sCreateUserCost()
 {
-  ParameterList params;
-  params.append("item_id", _item->id());
-  params.append("mode", "new");
-
-  itemCost newdlg(this, "", true);
-  if (newdlg.set(params) == NoError && newdlg.exec())
-    sFillList();
+  if (_item->isValid()) {
```
gilmoskowitz: This is a good guard but doesn't fix the bug as reported. `maintainItemCosts` probably needs a new `sHandleButtons()` method. See the [printQuote constructor](https://github.com/xtuple/qt-client/blob/56f6d6a6e639fd8d6e87965bd0ca2caaa4a18d0d/guiclient/printQuote.cpp#L38) and [method](https://github.com/xtuple/qt-client/blob/56f6d6a6e639fd8d6e87965bd0ca2caaa4a18d0d/guiclient/printQuote.cpp#L66-L69) for an example. This should probably handle the **Edit** and **Delete** buttons, too.

PR #940, opened 2015-11-08
I26654 project type x combo box

jrosengarden:


jrosengarden:
This was replaced by #941 and reverted out

PR #947, opened 2015-11-09
Issue #26847:correctly report errors returned by changeWoQty

garyhgohoos:


guiclient/salesOrderItem.cpp
```diff
@@ -2664,7 +2664,7 @@ void salesOrderItem::sHandleSupplyOrder()
               if (ordq.first())
               {
                 int result = ordq.value("result").toInt();
-                if (result < 0)
+                if (result != 0)
```
gilmoskowitz: This is a big violation of standard behavior. Better to make `changeWoQty()` raise exception now than hope that future generations (hah!) will remember `!=` instead of `<` here.

PR #948, opened 2015-11-09
Issue #26846:correctly position to next line item after cancelling

garyhgohoos:


guiclient/salesOrderItem.cpp
```diff
@@ -4343,6 +4343,7 @@ void salesOrderItem::sCancel()
 
   _modified  = false;
   _canceling = false;
+  sNext();
```
gilmoskowitz: Call `sPrev()` if we're canceling the last item? Or does `sNext()` do the right thing regardless?

PR #951, opened 2015-11-10
Issue #26883:tighten up credit checking and on account payments

garyhgohoos:


guiclient/salesOrderSimple.cpp
```diff
@@ -687,8 +687,9 @@ void salesOrderSimple::sPopulateCustomerInfo(int pCustid)
                                       "Customer must be taken off of Credit Hold "
                                       "before you may create a new Sales Order "
                                       "for the Customer." ) );
+          // set to configured default cash customer
           _cust->setId(-1);
-          _shipTo->setCustid(-1);
+          _cust->setId(_metrics->value("SSOSDefaultCustId").toInt());
```
gilmoskowitz: `_cust->setId(-1)` above is now redundant.

guiclient/salesOrderSimple.cpp
```diff
@@ -1883,6 +1885,14 @@ void salesOrderSimple::sEnterCashPayment()
 {
   XSqlQuery cashsave;
 
+  // check for on account
+  if (_fundsType->code() == "A")
+  {
+    sCalculateTotal();
+    sCompleteOrder();
+    return;
+  }
```
gilmoskowitz: There's something missing here... Let's say the total for the order is $100 and the purchaser is a Customer with $50 in available credit. The purchaser hands the clerk three $20 bills. The flow defined here bypasses the "$10 change due" prompt (spec sections 3.3.3.7.2 and 3.3.3.8) unless I've misread something.

garyhgohoos: the available credit would be allocated to the order separate from payment processing by clicking on the "Available Credit" link and allocating.  the allocation of the $50 credit would change the balance due to $50 which would result in the correct "$10 change due" prompt when $60 is received.

PR #952, opened 2015-11-11
Issue # 26866: Add documents tab to Vendor master

andersdd:
Linked via CRM account in the same manner as Customer.

### DIFF
--- a/guiclient/vendor.cpp
+++ b/guiclient/vendor.cpp
@@ -1001,7 +1001,7 @@ bool vendor::sPopulate()
     sFillTaxregList();
 
     _comments->setId(_crmacctid);
-    _documents->setId(_crmacctid);
+    _documents->setId(_vendid);
     _address->setSearchAcct(_crmacctid);
     _charass->setId(_vendid);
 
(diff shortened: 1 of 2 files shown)


guiclient/vendor.cpp
```diff
@@ -1000,6 +1001,7 @@ bool vendor::sPopulate()
     sFillTaxregList();
 
     _comments->setId(_crmacctid);
+    _documents->setId(_crmacctid);
```
davidbeauchamp: Setting the id isn't enough for _documents, you also have to [setType](http://xtuple.github.io/desktop-programmer-ref/4.9/class_documents.html#a261dc7c55490b602ddab4344abf548fc) for the source document type, which IIRC is "V" for vendors.

davidbeauchamp: But now that I realize you're linking to the crmacctid you should use "CRMA"

gilmoskowitz: That's already done at line 1125 in the .ui.

davidbeauchamp: Ahh you're right about the type, I missed that. Completely agree on using the vendor specific type here.

PR #961, opened 2015-11-18
Issue #26811 Tab Order fix

szuke:

PR #962, opened 2015-11-18
I26810 tabsetting cm

szuke:


gilmoskowitz:
Like #961 there's a lot of .ui file churn that'll break the application. Could be a problem with loading the xTuple widget library in Qt Designer or Creator.

gilmoskowitz:
replaced by #1024 as this current pull request intermingles the fixes for 2 distinct issues.

PR #968, opened 2015-11-19
Issue #26935:fix sIssueStock when checking ordertype

garyhgohoos:


gilmoskowitz:
unfortunately phrased query

PR #971, opened 2015-11-20
Error reporter refactor

jrosengarden:


guiclient/enterPoReturn.cpp
```diff
@@ -181,16 +182,17 @@ void enterPoReturn::sPost()
     result = enterPost.value("result").toInt();
     if (result < 0)
     {
-      rollback.exec();
```
gilmoskowitz: We need to keep this `rollback.exec();`

PR #982, opened 2015-11-24
Issue #26832: Enforce user definable unique characteristic setting

andersdd:
requires xtuple/xtuple#2555

### DIFF
--- a/widgets/characteristicAssignment.cpp
+++ b/widgets/characteristicAssignment.cpp
@@ -162,6 +162,7 @@ int characteristicAssignment::set(const ParameterList &pParams)
     else if (param.toString() == "edit")
     {
       _mode = cEdit;
+      _originalchar = _charassid;
     }
     else if (param.toString() == "view")
     {
@@ -237,7 +238,10 @@ void characteristicAssignment::sSave()
       return;
     }
   }
-  if (_mode == cNew)
+  if (_mode == cNew || 
+      (_mode == cEdit && 
+       _originalchar != _char->model()->data(_char->model()->index(_char->currentIndex(), _d->idCol)) )
+     )
   {
     characteristicSave.prepare("SELECT EXISTS(SELECT true FROM charass "
                                "WHERE ((charass_char_id=:charass_char_id) "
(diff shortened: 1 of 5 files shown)


widgets/characteristicAssignment.cpp
```diff
@@ -237,6 +237,24 @@ void characteristicAssignment::sSave()
       return;
     }
   }
+  if (_mode == cNew)
```
davidbeauchamp: I think we need to do this check in edit mode too, since someone could just add a value then go in and edit it to be a duplicate otherwise.

andersdd: It is checking for the addition of the characteristic not the *value* entered of the char.  So we only need to check on new.

davidbeauchamp: you mean adding the characteristic to the target more than one times? I may be missing the goal here, but you can change both the value and the characteristic while editing a characteristic assignment

andersdd: @davidbeauchamp that is correct, and if you can change the char in edit mode then that does also require a check to be made.  Thanks!

andersdd: Change made to handle suggested scenario.

widgets/characteristicAssignment.cpp
```diff
@@ -237,6 +238,27 @@ void characteristicAssignment::sSave()
...
+  if (_mode == cNew || 
+      (_mode == cEdit && 
+       _originalchar != _char->model()->data(_char->model()->index(_char->currentIndex(), _d->idCol)) )
+     )
+  {
+    characteristicSave.prepare("SELECT EXISTS(SELECT true FROM charass "
+                               "WHERE ((charass_char_id=:charass_char_id) "
+                               "  AND (charass_target_id=:charass_target_id) "
+                               "  AND (charass_target_type=:charass_target_type))) AND "
+                               " (SELECT char_unique FROM char where char_id=:charass_char_id) as char_unique;");
```
gilmoskowitz: This rule should also be enforced in the charass trigger. Alternatively, be optimistic and always/only enforce it in the trigger. If the `RAISE EXCEPTION` is formatted properly, `ErrorReporter::error()` can give a friendly translated message as well as the SQL details.

PR #984, opened 2015-11-25
Fivetwo

gpazo:


common/xtNetworkRequestManager.cpp
```diff
@@ -19,23 +19,21 @@
...
   _url = url;
+  _params = params;
   _mutex = &mutex;
   _mutex->lock();
   _loop = new QEventLoop;
   startRequest(_url);
 }
 void xtNetworkRequestManager::startRequest(const QUrl & url) {
-    _nwrep = nwam->get(QNetworkRequest(url));
+    _nwrep = nwam->post(QNetworkRequest(url), _params.toUtf8());
```
gilmoskowitz: `get` should be the normal case, `put` optional.

guiclient/creditcardprocessor.cpp
```diff
@@ -342,16 +335,6 @@ CreditCardProcessor::CreditCardProcessor()
...
-#if QT_VERSION < 0x050000
-      if (cert && ! cert->isValid()) {
-        delete cert;
-        cert = new QSslCertificate(&certfile, QSsl::Der);
-      }
-      if (cert->isValid()) {
-        certs.append(*cert);
-        if (DEBUG) qDebug() << "adding certificate" << cert;
-      }
-#else
```
gilmoskowitz: Still need some `#if` because `QSslCertificate::isBlacklisted()` isn't in [Qt 4](http://doc.qt.io/qt-4.8/qsslcertificate.html).

guiclient/creditcardprocessor.cpp
```diff
@@ -1909,51 +1874,43 @@ int CreditCardProcessor::sendViaHTTP(const QString &prequest,
...
-        if (!waitForHTTP())
-        {
-          //Possible TODO: handle a timeout as indicated by a false return value from waitForHTTP()
-        }
+        QMutex mutex; //theoretically we create and pass the mutex, when processing is finished, code below will run
+        xtNetworkRequestManager request(ccurl, mutex, prequest.toUtf8());
         QApplication::restoreOverrideCursor();
-
-        if(reply->error() != QNetworkReply::NoError)
+        QByteArray data = request.response();
```
gilmoskowitz: Need to wait for `mutex` here.

gilmoskowitz:
We still need qt4 build compatibility for a few reasons. We can strip out the Qt 5 `#ifdef`s everywhere later.

PR #991, opened 2015-11-29
Error reporter refactor

jrosengarden:
These 3 commits are my "Turkey Weekend" work!  

The next batch will be a little while coming (maybe by the end of the week).  The next16 files "up-to-bat" have over 150 systemError calls, between them, to be replaced.

### DIFF
--- a/guiclient/group.cpp
+++ b/guiclient/group.cpp
@@ -113,10 +113,11 @@ enum SetResponse group::set(const ParameterList &pParams)
                  "VALUES( :grp_id, :grp_id, '' );" );
       groupet.bindValue(":grp_id", _grpid);
       groupet.exec();
-      if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Group Information"),
-                                    groupet, __FILE__, __LINE__))
+      if (groupet.lastError().type() != QSqlError::NoError)
       {
         groupet.exec("ROLLBACK;");
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Group Information"),
+                             groupet, __FILE__, __LINE__);
         _trapClose = false;
         return UndefinedError;
       }
--- a/guiclient/group.cpp
+++ b/guiclient/group.cpp
@@ -99,9 +99,9 @@ enum SetResponse group::set(const ParameterList &pParams)
       groupet.exec("SELECT NEXTVAL('grp_grp_id_seq') AS grp_id;");
       if (groupet.first())
         _grpid = groupet.value("grp_id").toInt();
-      else if (groupet.lastError().type() != QSqlError::NoError)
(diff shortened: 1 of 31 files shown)


guiclient/group.cpp
```diff
@@ -113,9 +113,9 @@ enum SetResponse group::set(const ParameterList &pParams)
                  "VALUES( :grp_id, :grp_id, '' );" );
       groupet.bindValue(":grp_id", _grpid);
       groupet.exec();
-      if (groupet.lastError().type() != QSqlError::NoError)
+      if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Group Information"),
+                                    groupet, __FILE__, __LINE__))
       {
-        systemError(this, groupet.lastError().databaseText(), __FILE__, __LINE__);
```
gilmoskowitz: Please put the `ErrorReporter::error()` call after the rollback. The original code had the wrong sequence and it would be nice to fix it. Beware of reusing `groupet`.

jrosengarden: Gil:  that 2nd comment you made didn't come to me in an email for some reason.  I was wondering why I hadn't seen a response to my fix.

Anyhow...I'm confused now.  Not sure what I'm supposed to do in guiclient/group.cpp.  Can you give me a hint?  Pretty please????  Thanks and.....sorry!

guiclient/group.cpp
```diff
@@ -115,8 +115,9 @@ enum SetResponse group::set(const ParameterList &pParams)
       groupet.exec();
       if (groupet.lastError().type() != QSqlError::NoError)
       {
-        systemError(this, groupet.lastError().databaseText(), __FILE__, __LINE__);
         groupet.exec("ROLLBACK;");
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Group Information"),
+                             groupet, __FILE__, __LINE__);
```
gilmoskowitz: @jrosengarden - the problem is that groupet at line 120 refers to the rollback at 118, not the original failed query. The fix would be to create a new XSqlQuery object to do the rollback instead of reusing groupet. Then the ErrorReporter::error() call will do the right thing.

jrosengarden: Gil

I get it but what confuses me is that I actually tested the code and the error:: method fired off properly.  From what you just explained ,... It shouldn't have.

gilmoskowitz: That is confusing. The message shouldn't have appeared. The [Qt docs](http://doc.qt.io/qt-5/qsqlquery.html#exec-1) are explicit: "the last error for this query is reset when exec() is called."

jrosengarden: Gil:

I think (know!) it was me.  I set the query equal to 0 to force the error message to pop.  I wasn’t thinking.   Sorry!

jrosengarden:
By Jove....I believe I have it this time!  (guiclient/group.cpp).

:-)

PR #993, opened 2015-11-30
Issue #26796: Fix Setup menu text (on Mac)

andersdd:
Reverted to same approach as the other menus.  Cannot test on Mac, but have tested on my build and changes are working.


gilmoskowitz:
This fix is incomplete - see [the Qt docs](http://doc.qt.io/qt-4.8/qmenubar.html) and search for `preferences`. The `xSetup` hack was an attempt to fix issues 25694 and 25695 but doesn't work properly. It looks like https://bugreports.qt.io/browse/QTBUG-43881 holds the answer to all three.

andersdd:
@gilmoskowitz not sure I understand why these two menus are different to the rest of the menus?  Are they broken on Mac as well?

gilmoskowitz:
Yes, they're broken on Macs. The assumption earlier was that Qt was treating the first occurrence of Setup in the menu system left-to-right, top-to-bottom as special. The problem is a little deeper than that. I was hoping to work on it this week.

PR #996, opened 2015-12-01
Issue #26705: ItemSite Lot/Serial control on Transfer Orders

andersdd:
Check for existing lot/serial setting when creating Item Site.  User can override if desired.

### DIFF
--- a/guiclient/itemSite.cpp
+++ b/guiclient/itemSite.cpp
@@ -586,6 +586,31 @@ bool itemSite::sSave()
 
   if (GuiErrorCheck::reportErrors(this, tr("Cannot Save Item Site"), errors))
     return false;
+
+  if (_controlMethod->currentIndex() < 2)
+  {
+    itemSave.prepare("SELECT EXISTS(SELECT 1 "
+                     "              FROM itemsite "
+                     "              WHERE ((itemsite_item_id = :itemsite_item_id) "
+                     "                 AND (itemsite_warehous_id <> :itemsite_id) "
+                     "                 AND (itemsite_controlmethod IN ('L','S'))));");
+    itemSave.bindValue(":itemsite_id", _itemsiteid);
+    itemSave.bindValue(":itemsite_item_id", _item->id());
+    itemSave.exec();
+    if (itemSave.first())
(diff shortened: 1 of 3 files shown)


guiclient/itemSite.cpp
```diff
@@ -586,6 +586,31 @@ bool itemSite::sSave()
...
     return false;
+
+  if (_controlMethod->currentIndex() < 2)
+  {
+    itemSave.prepare("SELECT EXISTS(SELECT 1 "
+                     "              FROM itemsite "
+                     "              WHERE ((itemsite_item_id = :itemsite_item_id) "
+                     "                 AND (itemsite_warehous_id <> :itemsite_id) "
+                     "                 AND (itemsite_controlmethod IN ('L','S'))));");
+    itemSave.bindValue(":itemsite_id", _itemsiteid);
```
gilmoskowitz: The binding (itemsite) doesn't match the usage (warehouse/site).

guiclient/itemSite.cpp
```diff
@@ -586,6 +586,31 @@ bool itemSite::sSave()
 
   if (GuiErrorCheck::reportErrors(this, tr("Cannot Save Item Site"), errors))
     return false;
+
+  if (_controlMethod->currentIndex() < 2)
```
gilmoskowitz: Depending on `currentIndex()` is fragile. It would be better to populate `_controlMethod` with [XComboBox::append()](https://github.com/xtuple/qt-client/blob/4_10_x/widgets/xcombobox.h#L136-L137) in the .cpp constructor (just before [this](https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/itemSite.cpp#L115-L116)) instead of in [the .ui](https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/itemSite.ui#L349-L368), then work with [code](https://github.com/xtuple/qt-client/blob/4_10_x/widgets/xcombobox.h#L121) values.

andersdd: @gilmoskowitz I have converted the entire itemSite to use code() rather than currentIndex()

PR #1001, opened 2015-12-03
issue #26994 - reduce redundancy in smart bar code handling

gilmoskowitz:
This is a first pass at replacing functions with data to scan smart bar codes. The next step is to drive this mechanism from a database table, allowing extensions to add their own smart bar codes.

### DIFF
--- a/guiclient/inputManager.cpp
+++ b/guiclient/inputManager.cpp
@@ -1,33 +1,54 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2015 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
  * version 1.0, the full text of which (including xTuple-specific Exhibits)
  * is available at www.xtuple.com/CPAL.  By using this software, you agree
  * to be bound by its terms.
  */
 
-#include <QObject>
-#include <QList>
-#include <QKeyEvent>
-#include <QEvent>
 #include <QDebug>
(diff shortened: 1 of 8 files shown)


guiclient/inputManager.cpp
```diff
@@ -205,9 +206,9 @@ bool InputManager::eventFilter(QObject *, QEvent *pEvent)
 #ifdef Q_OS_MAC
       case cPrologCtrl:
         // why does character come back as 0?
-	// on an Intel Mac with Qt 4 the key() came back as Key_PageUp
-	// but with PowerPC Mac with Qt 3 the key() came back as 'V'.
-	// Accept either for now.
+        // on an Intel Mac with Qt 4 the key() came back as Key_PageUp
+        // but with PowerPC Mac with Qt 3 the key() came back as 'V'.
```
davidbeauchamp: Is this still the case?

guiclient/inputManagerPrivate.h
```diff
@@ -0,0 +1,77 @@
...
+    int                 _state;
+    int                 _cursor;
+    ScanEvent          *_event;
+    int                 _length1;
+    int                 _length2;
+    int                 _length3;
+    QString             _buffer;
+
+    void dispatchScan(int type);
+    void dispatchWorkOrderOperation();
```
davidbeauchamp: Are the rest of these dispatch* functions still used? I think I may be missing something, but I only see dispatchScan()   being used, and the rest of the function definitions removed from inputManager.cpp

guiclient/inputManager.cpp
gilmoskowitz: Seems to be, at least with Qt 4.8. Didn't test with Qt 5.

guiclient/inputManagerPrivate.h
gilmoskowitz: Good catch.

PR #1003, opened 2015-12-04
Issue #26981: Transfer Order Item cancellation

andersdd:
Prevent cancelling an unreleased TO item

### DIFF
--- a/guiclient/transferOrderItem.cpp
+++ b/guiclient/transferOrderItem.cpp
@@ -164,6 +164,7 @@ enum SetResponse transferOrderItem::set(const ParameterList &pParams)
 
       _save->setEnabled(false);
       _next->setText(tr("New"));
+      _cancel->setEnabled(false);
       _comments->setReadOnly(true);
       _item->setReadOnly(false);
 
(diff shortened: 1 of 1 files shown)


guiclient/transferOrderItem.cpp
```diff
@@ -870,9 +871,8 @@ void transferOrderItem::populate()
     return;
   }
 
-  if( (cNew == _mode) || ((cEdit == _mode) &&
-			  (item.value("toitem_status").toString() == "O")) )
-    _cancel->setEnabled(_shippedToDate->toDouble()==0.0);
+  if( (cEdit == _mode) && (item.value("toitem_status").toString() == "O") )
```
gilmoskowitz: What's the rationale for preventing **Cancel** on new transfer order items? Seems like a step backwards in the UI.

andersdd: There are numerous issues cancelling an item on New
[1] System still asks to create an Item Site after cancelling the line
[2] In my system cancelling a new line was cancelling the line PLUS the previously entered line
[3] It does not make sense to enter a cancelled item especially when you can only cancel the item in EDIT mode after the T/O has been released.  

Preventing cancellation on entering a New line seems more consistent

PR #1009, opened 2015-12-06
Issue #26705: Fix Transfer Order transfer of Lot Serial to Receiving Site

andersdd:
Resolves passing of lot/serial information between sites on a T/O


davidbeauchamp:
+1

PR #1015, opened 2015-12-08
Issue #27014:new return authorizations by item display

garyhgohoos:


gilmoskowitz:
depends on https://github.com/xtuple/xtuple/pull/2570 or whatever replaces it in private-extensions

PR #1021, opened 2015-12-10
Issue #26233: Hide inactive Account Numbers from Financial Report 

andersdd:
Financial Report item screen


andersdd:
Closing PR based on feedback.   Was an irritant rather than an important issue

PR #1034, opened 2015-12-23
Issue #15171: Sales Order Credit  Limit check

andersdd:
requires xtuple/xtuple#2578

Configurable S/O credit check on save.  Also enhances Open SO screen to filter/display hold type.

### DIFF
--- a/guiclient/salesOrder.cpp
+++ b/guiclient/salesOrder.cpp
@@ -5502,8 +5502,6 @@ bool salesOrder::creditLimitCheck()
   {
     customerCurrent = creditCheck.value("creditcheck_bookings").toDouble() + 
                       creditCheck.value("creditcheck_aropen").toDouble();
-    if (_mode == cNew)
-      customerCurrent += _balance->localValue();
 
     // The Credit Check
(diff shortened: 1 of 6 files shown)


guiclient/salesOrder.cpp
```diff
@@ -152,7 +152,8 @@ salesOrder::salesOrder(QWidget *parent, const char *name, Qt::WindowFlags fl)
   connect(_shipToAddr,          SIGNAL(addressChanged(QString,QString,QString,QString,QString,QString, QString)),
           _shipToCntct, SLOT(setNewAddr(QString,QString,QString,QString,QString,QString, QString)));
 
-  _saved = false;
+  _saved  = false;
+  _saving = false;
```
gilmoskowitz: `_saving` seems unnecessary. It's only used in `sSave()` and `save()`, where we know it's true, and it's never reset to `false`.

andersdd: The credit check was being kicked in when opening the S/O so something calls save() on opening.  This superfluous variable ensures we only credit check after pressing save button.

PR #1044, opened 2016-01-04
issue #27282 - reduce the work done when saving customers

gilmoskowitz:
@garyhgohoos: Please review this carefully. The changes look innocuous which worries me - What essential work is _not_ going to happen?

Tested in both Customer and Workbench modes with no apparent problems.


garyhgohoos:
I agree that the innocuous change of removing populate() may cause a problem, but nothing jumps out at me

PR #1045, opened 2016-01-04
issue #26796 and issue #25694 - 'Setup' menu items should not be handled specially on Macs

gilmoskowitz:


gilmoskowitz:
follow-on changes to #993

andersdd:
@gilmoskowitz code reviewed.  That looks a much cleaner solution than before

PR #1067, opened 2016-01-18
issue #27365: Item images metasql

andersdd:
Convert Images List from embedded SQL to metaSQL.  Allows choosing to calculate image size or not.  Significant performance gain on Db with a lot of images if the size is not calculated.

Requires xtuple/xtuple#2589

### DIFF
--- a/guiclient/images.cpp
+++ b/guiclient/images.cpp
@@ -14,6 +14,8 @@
 #include <QSqlError>
 
 #include <parameter.h>
+#include <metasql.h>
+#include "mqlutil.h"
 
 #include "image.h"
 #include "guiclient.h"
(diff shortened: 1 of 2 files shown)


guiclient/images.cpp
```diff
@@ -108,14 +110,9 @@ void images::sDelete()
...
-  // Do not select image_data into the list
-  imagesFillList.exec("SELECT image_id, image_name, image_descrip, LENGTH(image_data) AS image_size, "
-         "       CASE WHEN nspname='public' THEN ''"
-         "            ELSE nspname END AS nspname"
-         "  FROM image, pg_class, pg_namespace "
-         " WHERE ((image.tableoid=pg_class.oid)"
-         "   AND  (relnamespace=pg_namespace.oid))"
-         "ORDER BY image_name;" );
+  MetaSQLQuery mql = mqlLoad("images", "list");
+  ParameterList params;
```
davidbeauchamp: How does doNotShowImageSize get set for the MetaSQL?

andersdd: I was doing a quick proof of concept to determine if removing the size calc resolved the performance issues (which it does dramatically).  The doNotShowImageSize is just a placeholder to demonstrate the alternate queries.   I will add a switch to the ui based on your feedback  :-)

davidbeauchamp:
+1

PR #1072, opened 2016-01-22
issue #26887 - use revised documents handling database code

gilmoskowitz:


gilmoskowitz:
This improves performance all by itself by an order of magnitude or so but still suffers from the data duplication problem. In conjunction with xtuple/xtuple#2594, performance is further improved and the data duplication problem is fixed.

PR #1074, opened 2016-01-22
More Qt Script exposure objects and fixes.

bendiy:


scriptapi/qurlproto.cpp
```diff
@@ -102,16 +103,24 @@ void QUrlProto::addQueryItem(const QString &key, const QString &value)
...
+  _query = new QUrlQuery(queryString);
+}
+#endif
+
 // RETURN AND PARAM TYPES CHANGED FOR Qt5
 // I did not bother with Qt4 backwards compatibility directives - true backwards compatibility
 // would require varying the return/param types of several methods since QStrings are now used over QByteArrays
 #if QT_VERSION >= 0x050000
-QStringList QUrlProto::allEncodedQueryItemValues(const QString &key) const
+QStringList QUrlProto::allEncodedQueryItemValues(const QString &key)
```
bendiy: Removing `const` here is unfortunate, but adding `setProtectedQuery(item->query());` to these methods means they are no longer pure.
I can't find any other way to set `_query` to the `QUrl->query()`.

scriptapi/scriptapi.pro
```diff
@@ -36,7 +36,7 @@ HEADERS += setupscriptapi.h \
     qapplicationproto.h \
     qboxlayoutproto.h \
     qbytearrayproto.h \
-    qcoreapplicationproto.h     \
+    qcoreapplicationproto.h \
```
bendiy: @gilmoskowitz We stomped on each other here, but I think this is the only thing you added. Note, `xnetworkaccessmanager` is removed and exposed as `qnetworkaccessmanagerproto` now.

scriptapi/qnetworkreplyproto.cpp
```diff
@@ -45,14 +63,15 @@ void QNetworkReplyProto::close()
...
 }
 
-int QNetworkReplyProto::error() const
+QNetworkReply::NetworkError QNetworkReplyProto::error() const
 {
   QNetworkReply *item = qscriptvalue_cast<QNetworkReply*>(thisObject());
   if (item)
     return item->error();
-  return (QNetworkReply::UnknownNetworkError);
+  return QNetworkReply::NetworkError();
```
gilmoskowitz: `UnknownNetworkError` is a better choice. What is the result of calling a constructor for an `enum`?

scriptapi/qnetworkreplyproto.h
```diff
@@ -85,6 +89,10 @@ class QNetworkReplyProto : public QObject, public QScriptable
     Q_INVOKABLE qint64      write(const char * data, qint64 maxSize);
     Q_INVOKABLE qint64      write(const QByteArray &byteArray);
     Q_INVOKABLE qint64      write(const QString &string);
+
+  public Q_SLOTS:
```
gilmoskowitz: `Q_SLOTS` should be simply `slots`.

scriptapi/qnetworkreplyproto.cpp
bendiy: This one is fun because it doesn't work in Qt Scripting in Qt 5 because the `void	error(QNetworkReply::NetworkError code)` signal stomps on it and `NetworkError	error() const` is not exposed. The use of `UnknownNetworkError` is misleading because calling `QNetworkReply->error()` when there is no instance of `QNetworkReply` isn't really a `Network` error. It's more likely a bug in the code.

I thought `QNetworkReply::NetworkError();` returned the default `enum`, `QNetworkReply::NoError` which `QNetworkReply->error()` returns when there is no error. 
> If no error was found, returns NoError.

Should I switch it back?

Either way, it won't work in Qt 5.5. Maybe they will fix this bug in a future release. The work around is to listen for the `void	error(QNetworkReply::NetworkError code)` signal.

gilmoskowitz: Please switch back the return value to `UnknownNetworkError` - at least then we know something went wrong, even if we don't know what.

To deal with the multiple meanings of `error()`, the two options I see are
- rely on the signal as you suggest and `#if` out the method prototype
- change the name in the prototype to `getError()` (awkward but functional)

I'm leaning toward relying on the signal but am open to discussion.

davidbeauchamp: I think relying on the signal is the best course of action here, Though I am not opposed at all to changing the prototype to getError() because sometimes it is convenient to check for errors inside the finished signal handler.

bendiy: Ideally, our exposed prototype should match the Qt object documentation. I don't like the idea of `QNetworkReply->getError()` because we can't just look at the `QNetworkReply` docs, we now have to look at our source that exposes it to see how it actually works instead of how it should work.

I'll switch back to `UnknownNetworkError`. I don't think we should `#if` out the method because it might be fixed in qt 5.6 or something in the future.

PR #1094, opened 2016-01-30
Expose QSsl objects

bendiy:


scriptapi/qbytearrayproto.cpp
```diff
@@ -9,27 +9,66 @@
...
-void QByteArrayfromScriptValue(const QScriptValue &obj, QByteArray* &item)
+void QByteArrayPointerfromScriptValue(const QScriptValue &obj, QByteArray* &item)
 {
   item = new QByteArray(obj.toVariant().toByteArray());
 }
 
+QScriptValue QByteArraytoScriptValue(QScriptEngine *engine, QByteArray const &item)
+{
+  QByteArray tmpBA = item;
+  QVariant   tmpVar = QVariant(tmpBA);      // TODO: memory leak?
```
gilmoskowitz: No memory leak here. `tmpVar` is on the stack and will be purged when the function returns.

scriptapi/qsslcertificateproto.cpp
```diff
@@ -0,0 +1,410 @@
...
+{
+  QString certificate = obj.property("_certificate").toString();
+  QSslCertificate newCert = QSslCertificate(certificate.toLocal8Bit(), QSsl::Pem);
+  item.swap(newCert);
+}
+
+QScriptValue QSslCertificatePointertoScriptValue(QScriptEngine *engine, QSslCertificate* const &item)
+{
+  QScriptValue obj = engine->newObject();
+  obj.setProperty("_certificate", qPrintable(QString(item->toPem())));
```
gilmoskowitz: Why `_certificate` as a property instead of using the string as the value of `obj` directly?

scriptapi/qsslconfigurationproto.cpp
```diff
@@ -0,0 +1,459 @@
...
+  QSslCertificate localCertificate = item->localCertificate();
+  obj.setProperty("_localCertificate", qPrintable(QString(localCertificate.toPem())));
+
+  QSslKey privateKey = item->privateKey();
+  obj.setProperty("_privateKey", qPrintable(QString(privateKey.toPem())));
+
+  obj.setProperty("_protocol", item->protocol());
+
+  return obj;
+}
```
gilmoskowitz: It would be good to document why the use of these `_` properties is required. Why doesn't exposing the methods suffice?

scriptapi/qsslcertificateproto.cpp
bendiy: I want `obj` to be a `QSslCertificate` with all it's functions. Having it be just a string would loose all of that.

scriptapi/qbytearrayproto.cpp
bendiy: This is removed in my next PR.

scriptapi/qsslconfigurationproto.cpp
bendiy: I'm on the fence about the need for the `_myprop`. I don't want it to be there, but it's the only way I can consistently get `QSslKey` and `QSslCertificate` to work.

I'll try to eliminate these to/from functions all together if I can.

bendiy: @gilmoskowitz I've fix the need for `_myProp` in this commit:
https://github.com/bendiy/qt-client/commit/2dedb376ab29c4ff6361d8939a1dcd996511bea4

On this PR:
https://github.com/xtuple/qt-client/pull/1117

You can ether merge both or just that one.

PR #1131, opened 2016-02-15
Issue #27378:avoid negative receipt greater than qoh

garyhgohoos:


guiclient/correctProductionPosting.cpp
```diff
@@ -98,19 +98,31 @@ bool correctProductionPosting::okToPost()
...
+    }
+
+    if (itemtypeq.value("itemsite_qtyonhand").toDouble() < _qty->toDouble())
+    {
+      QMessageBox::warning(this, tr("Cannot Post Correction"),
+                           tr("You may not post a correction to a Work Order for a "
+                              "Item Site with a quantity on hand less than the "
+                              "correction quantity."));
+      return false;
+    }
```
gilmoskowitz: This should be limited to average costed items. This check as written may prevent accurate production corrections when some other inaccurate transaction is actually the one driving QOH negative.

garyhgohoos: i intentionally applied this edit to all items.  agree that other inaccurate transactions (or accurate transactions) would be the cause of driving the QOH negative.  i think this is a good check which would prompt the fixing (or reversing) of other transactions.  for example, if the inventory that was created by a postProduction has been consumed, does it make sense to allow the correction?  why hide the fact that a negative inventory mess is being created?

PR #1137, opened 2016-02-18
Issue #26675:correctly set shipto name for non-drop ship orders

garyhgohoos:


gilmoskowitz:
@rkannadevara - when testing, please make sure that both 26675 and 25734 are fixed.

PR #1142, opened 2016-02-22
Issue #27055:disable QWheelEvent

garyhgohoos:


widgets/xcombobox.cpp
```diff
@@ -1598,6 +1598,10 @@ void XComboBox::mousePressEvent(QMouseEvent *event)
   QComboBox::mousePressEvent(event);
 }
 
+void XComboBox::wheelEvent(QWheelEvent *event)
+{
```
gilmoskowitz: This should be controlled by a user preference which defaults to "accept wheel events" for XComboBoxes. Some users rely on the original behavior.

garyhgohoos: user preference added which defaults to original behavior

PR #1143, opened 2016-02-22
Issue #26484:prevent locked quotes from being converted

garyhgohoos:


guiclient/quotes.h
```diff
@@ -42,6 +43,10 @@ public slots:
 
 signals:
     void finishedPrinting(int);
+  
+private:
+  AppLock _lock;
```
gilmoskowitz: The AppLock should be local to the method that uses it rather than part of the class definition.

guiclient/salesOrder.cpp
```diff
@@ -1292,7 +1292,7 @@ bool salesOrder::save(bool partial)
   }
 
   // TODO: should this be done before saveSales.exec()?
-  if ((cNew == _mode) && (!_saved)
+  if ((cNew == _mode || cNewQuote == _mode) && (!_saved)
```
gilmoskowitz: Or `if (ISNEW(_mode) && ! _saved...)`

guiclient/salesOrder.cpp
```diff
@@ -2448,7 +2448,7 @@ void salesOrder::populate()
   if ( (_mode == cNew) || (_mode == cEdit) || (_mode == cView) )
   {
     XSqlQuery so;
-    if (_mode == cEdit
+    if ((_mode == cEdit || cEditQuote == _mode)
```
gilmoskowitz: Or `if (ISEDIT(_mode) && ...)`

PR #1151, opened 2016-03-07
fix bad method call to allow compiling for upcoming fedora release

gilmoskowitz:
Unit tested the one code change and confirmed that the 4.9.4 application will connect to a 4.9.3 database without error or warning.

@dpocock - Could you please review and let me know if anything else is required at this time?

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+postbooks (4.9.4-1) experimental; urgency=medium
+
+  * Placeholder
+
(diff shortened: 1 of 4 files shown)


davidbeauchamp:
+1 if no response otherwise

dpocock:
Some other Fedora users have been discussing this here:

https://bugzilla.redhat.com/show_bug.cgi?id=1307871

Could you compare against the patch on the bug and ask them if they want to test 4.9.4?

I'm going out to Singapore next week for FOSSASIA and I probably won't be able to look at updating the packages until I return

PR #1157, opened 2016-03-13
Issue #24549: Resolve some currency conversion issues

andersdd:
Extend entry of Exchange Rate to 6 decimal places which is required to avoid 0.01 rounding issues in the Bank Reconciliation.

The scenario mentioned in the issue appeared resolved as a result of this change (no rounding issues in the Bank Rec.) but I cannot guarantee this is the entire solution.


gilmoskowitz:
See also xtuple/xtuple#2651

PR #1168, opened 2016-03-28
guiErrorCheckRefactor 03/28/16 - Have Gil check one more (that had so…

jrosengarden:
…me oddities) to be sure I'm doing this properly

Gil:  Forgive me.  This one had some 'oddities' so I want to make sure I handled them properly before I move on and start stacking up multiple files in a single commit/push.

### DIFF
--- a/guiclient/arOpenItem.cpp
+++ b/guiclient/arOpenItem.cpp
@@ -222,20 +222,18 @@ void arOpenItem::sSave()
           <<GuiErrorCheck(_tax->localValue() > _amount->localValue(), _tax,
                           tr("The tax amount may not be greater than the total Receivable Memo amount."));
 
-    if(GuiErrorCheck::reportErrors(this,tr("Cannot Save Receivable Memo"),errors))
-        return;
 
     if (_useAltPrepaid->isChecked())
     {
-      errors.clear();
       errors<<GuiErrorCheck(_altSalescatidSelected->isChecked() && !_altSalescatid->isValid(), _altSalescatid,
                            tr("You must choose a valid Alternate Sales Category for this Receivable Memo before you may save it."))
             <<GuiErrorCheck(_altAccntidSelected->isChecked() && !_altAccntid->isValid(), _altAccntid,
                           tr("You must choose a valid Alternate Prepaid Account Number for this Receivable Memo before you may save it."));
 
-      if(GuiErrorCheck::reportErrors(this,tr("Cannot Save Receivable Memo"),errors))
-          return;
     }
+
+    if(GuiErrorCheck::reportErrors(this,tr("Cannot Save Receivable Memo"),errors))
+        return;
 
     if (_docType->code() == "C")
     {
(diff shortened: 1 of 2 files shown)


guiclient/arOpenItem.cpp
```diff
@@ -210,52 +211,30 @@ void arOpenItem::sSave()
...
+    errors<<GuiErrorCheck(!_docDate->isValid(), _docDate,
+                            tr("You must enter a date for this Receivable Memo before you may save it."))
+          <<GuiErrorCheck(!_dueDate->isValid(), _dueDate,
+                           tr("You must enter a date for this Receivable Memo before you may save it."))
+          <<GuiErrorCheck(_amount->isZero(), _amount,
+                          tr("You must enter an amount for this Receivable Memo before you may save it."))
+          <<GuiErrorCheck(_tax->localValue() > _amount->localValue(), _tax,
+                          tr("The tax amount may not be greater than the total Receivable Memo amount."));
+
+    if(GuiErrorCheck::reportErrors(this,tr("Cannot Save Receivable Memo"),errors))
```
gilmoskowitz: Remove this instance of `GuiErrorCheck::reportErrors()`, then...

guiclient/arOpenItem.cpp
```diff
@@ -210,52 +211,30 @@ void arOpenItem::sSave()
...
-        return;
-      }
+      errors.clear();
+      errors<<GuiErrorCheck(_altSalescatidSelected->isChecked() && !_altSalescatid->isValid(), _altSalescatid,
+                           tr("You must choose a valid Alternate Sales Category for this Receivable Memo before you may save it."))
+            <<GuiErrorCheck(_altAccntidSelected->isChecked() && !_altAccntid->isValid(), _altAccntid,
+                          tr("You must choose a valid Alternate Prepaid Account Number for this Receivable Memo before you may save it."));
+
+      if(GuiErrorCheck::reportErrors(this,tr("Cannot Save Receivable Memo"),errors))
+          return;
```
gilmoskowitz: get rid of the `errors.clear()` and move _this_ `GuiErrorCheck::reportErrors()` call outside the `if` block. The code is a little simpler and will allow future versions of `reportErrors()` to be more helpful.

guiclient/arOpenItem.cpp
```diff
@@ -697,21 +676,14 @@ void arOpenItem::sTaxDetail()
...
-      return;
-    }
+      QList<GuiErrorCheck>errors;
+      errors<<GuiErrorCheck(!_docDate->isValid() || !_dueDate->isValid(), _docDate,
+                            tr("You must enter document and due dates for this Receivable Memo before you may set tax amounts."))
+           <<GuiErrorCheck(_amount->isZero(), _amount,
+                           tr("You must enter an amount for this Receivable Memo before you may set tax amounts."));
+
+      if(GuiErrorCheck::reportErrors(this,tr("Cannot Set Tax Amounts"),errors))
+          return;
```
gilmoskowitz: We haven't used `GuiErrorCheck` in places like this but it's a good idea. I'd suggest separating the two date checks so the reporting can be more precise.

jrosengarden:
I get it;  
1)  Add all necessary error checks to the QList variable as necessary (which made the errors.clear() stupid) and then 
2) ONLY call the GuiErrorCheck::reportErrors method once within the method

PR #1173, opened 2016-03-30
Fix QByteArray static method exposure. Add helper functions to emulate Node.js's Buffer.

bendiy:


scriptapi/qbytearrayproto.cpp
```diff
@@ -130,46 +133,43 @@ void setupQByteArrayProto(QScriptEngine *engine)
   engine->setDefaultPrototype(qMetaTypeId<QByteArray>(), proto);
 
   QScriptValue constructor = engine->newFunction(constructQByteArray, proto);
-  engine->globalObject().setProperty("QByteArray",  constructor);
 
 #if QT_VERSION >= 0x050000
   // Static Public Members:
-  // TODO: These are crashing the client.
```
bendiy: These were crashing because I missed the `return engine...`.

scriptapi/qbytearrayproto.cpp
```diff
@@ -185,45 +187,54 @@ QScriptValue constructQByteArray(QScriptContext *context,
...
+  if (context->argumentCount() == 0) {
     obj = QByteArray();
-
-  else if (context->argumentCount() == 1 &&
-             qscriptvalue_cast<QByteArray*>(context->argument(0)))
+  } else if (context->argumentCount() == 1 &&
+             qscriptvalue_cast<QByteArray*>(context->argument(0))) {
     obj = QByteArray(*(qscriptvalue_cast<QByteArray*>(context->argument(0))));
+  } else if (context->argumentCount() == 1 &&
+             context->argument(0).isArray()) {
```
bendiy: Adds support for passing an array in Qt Script. `var foo = new QByteArray([1,2,3,4]);`

scriptapi/qbytearrayproto.cpp
```diff
@@ -1039,3 +1051,321 @@ QString QByteArrayProto::toString() const
     return QString(*item);
   return QString();
 }
+
```
bendiy: Add support to do Node.js Buffer emulation in C++:
https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html

The other methods there are done in javascript.

gilmoskowitz: This code could be greatly simplified. Let's wait for the next go-round.

bendiy: Agreed. A helper function or two could simplify it.

PR #1177, opened 2016-04-06
issue #27691 Issuing multiple quantity of same serial number

ChristopherCotnoir:
Changed prompt to warning when issuing more items than available. Action is disallowed instead of simply discouraged.

### DIFF
--- a/guiclient/distributeToLocation.cpp
+++ b/guiclient/distributeToLocation.cpp
@@ -86,6 +86,12 @@ enum SetResponse distributeToLocation::set(const ParameterList &pParams)
     _locationQty->setDouble(locQty);
   }
 
+  param = pParams.value("itemsite_controlmethod", &valid);
+  if (valid)
+  {
+    _controlMethod = param.toString();
+  }
+
   param = pParams.value("distribute", &valid);
   if (valid)
   {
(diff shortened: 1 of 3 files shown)


guiclient/distributeToLocation.cpp
```diff
@@ -130,15 +130,11 @@ void distributeToLocation::sDistribute()
...
-  if (qty < 0 && _availToDistribute < qAbs(qty) &&
-      QMessageBox::question(this, tr("Distribute More Than Available?"),
-			    tr("<p>It appears you are trying to distribute "
-			       "more than is available to be distributed. "
-			       "Are you sure you want to distribute this "
-			       "quantity?"),
-			    QMessageBox::Yes,
-			    QMessageBox::No | QMessageBox::Default) == QMessageBox::No)
+
+  if (qty < 0 && _availToDistribute < qAbs(qty))
```
gilmoskowitz: This change is on the right track but not quite right. We should allow distributing arbitrary numbers of lot controlled items and items that are only location controlled. Items that are serial controlled should be restricted to `qAbs(qty) == 1` but still warn if the available quantity < 1. Does this sound right @garyhgohoos @andersdd @davidbeauchamp ?

andersdd: Personally, I dont think we should distribute more than available - that is how we end up with negative inventory Qtys.   Agree on the serial check should only allow distributing singles.

ChristopherCotnoir: I was going based off the last comment by the user that filed the bug report, which suggested that negative quantities should not be allowed in any circumstance. It's ultimately our judgement call, but I don't know of any instance where anyone would actually want a negative quantity at a location, so it seems better not to allow it, because it would mean someone is making a mistake, like in the cases above.

garyhgohoos: in the real world, negative quantities are required for timing issues.  the issue transaction for goods that just arrived may occur before the receipt transaction.  everybody does not perform transactions in real time.

andersdd: Whilst I agree with @garyhgohoos about real-world timing issues, some companies use the system as a control mechanism so want to prevent negative inventory and ensure transactions are processed in sequence.  
I have partially completed a development that makes the ability to allow negative inventory a configurable option, but not completed as yet

PR #1182, opened 2016-04-11
Issue #27779: Expose the embedded AddressCluster inside ContactWidget

davidbeauchamp:
Also exposes AddressCluster* to the script engine.

### DIFF
--- a/scriptapi/addressclustersetup.cpp
+++ b/scriptapi/addressclustersetup.cpp
@@ -1,18 +1,30 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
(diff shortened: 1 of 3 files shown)


scriptapi/addressclustersetup.cpp
```diff
@@ -1,18 +1,30 @@
...
+QScriptValue AddressClustertoScriptValue(QScriptEngine *engine, AddressCluster* const &item)
+{
+  return engine->newQObject(item);
+}
+
+void AddressClusterfromScriptValue(const QScriptValue &obj, AddressCluster* &item)
+{
+  item = qobject_cast<AddressCluster*>(obj.toQObject());
+}
+
```
bendiy: You might not need the to/from functions.

davidbeauchamp: I tried that first but Qt barked at me to register to/from's for the pointer. The non-pointer versions still work as expected without qScriptRegisterMetaType().

PR #1191, opened 2016-04-21
issue #27814 Reassign Lot/Serial # Itemsite passing

ChristopherCotnoir:
Itemsite is properly passed to the Reassign Lot/Serial # Window.


guiclient/displays/dspInventoryLocator.cpp
```diff
@@ -124,6 +124,16 @@ void dspInventoryLocator::sReassignLotSerial()
 {
   ParameterList params;
   params.append("itemloc_id", list()->id());
+  XSqlQuery warehouse; 
+  warehouse.prepare( "SELECT warehous_id "
+               "FROM warehous "
+               "WHERE (warehous_code=:warehous_code);" );
+  warehouse.bindValue(":warehous_code", list()->selectedItems().at(0)->text(0));
```
davidbeauchamp: Should use the column name instead of the index, ->text("column_name") will work even if the position changes.

guiclient/displays/dspInventoryLocator.cpp
```diff
@@ -124,6 +124,16 @@ void dspInventoryLocator::sReassignLotSerial()
...
+  XSqlQuery warehouse; 
+  warehouse.prepare( "SELECT warehous_id "
+               "FROM warehous "
+               "WHERE (warehous_code=:warehous_code);" );
+  warehouse.bindValue(":warehous_code", list()->selectedItems().at(0)->text("warehous_code"));
+  warehouse.exec();
+  if (warehouse.first())
+  {
+    params.append("warehous_id", warehouse.value("warehous_id").toInt());
+  }
```
gilmoskowitz: Turns out the fix is much simpler than this. No need to query the `warehous` view or `whsinfo` table here since that information is available indirectly in reassignLotSerial.cpp.

guiclient/reassignLotSerial.cpp
```diff
@@ -102,6 +101,13 @@ enum SetResponse reassignLotSerial::set(const ParameterList &pParams)
     }
   }
 
+  param = pParams.value("warehous_id", &valid);
+  if (valid)
+  {
+      _warehouse->setId(param.toInt());
+      _warehouse->setEnabled(false); 
+  }
+
```
gilmoskowitz: See https://github.com/xtuple/qt-client/blob/4_10_x/widgets/itemcluster.h#L285 
We can connect `_item`'s `warehouseIdChanged(int)` signal to `_warehouse`'s `setId(int)` slot instead of manually selecting the info in the parent window and working with it here.

PR #1194, opened 2016-04-25
Issue #27728,  issue #26654 - backport for 4.9.5

gilmoskowitz:
The default query to populate the Project Type should only use active types. However, when a Project with an inactive type is viewed or edited, that option must be available to the combobox. Otherwise it's not visible and, if saved after edit, removed from the project. Thus repopulate the project type combobox if "this" project's type isn't there already.


ChristopherCotnoir:
+1

PR #1200, opened 2016-04-26
issue #27751 Added missing error check

ChristopherCotnoir:


guiclient/uom.cpp
```diff
@@ -154,6 +154,11 @@ void uom::sSave()
   uomSave.bindValue(":uom_item_weight", QVariant(_weightUom->isChecked()));
   uomSave.exec();
 
+  if (uomSave.lastError().type() != QSqlError::NoError)
```
gilmoskowitz: This is a good candidate for `ErrorReporter::error()`. See line ~283 in this file for an example.

gilmoskowitz:
See xtuple/private-extensions#777 and xtuple/private-extensions#779 for the other two parts of the fix

PR #1210, opened 2016-05-03
issue #27585 Fixed work order number formatting

ChristopherCotnoir:


widgets/woCluster.cpp
```diff
@@ -68,7 +68,7 @@ void WoLineEdit::init()
   _qtyReceived = 0.0;
 
   _query =  QString("SELECT wo_id AS id, "
-                    "       wo_number::text || '-' || wo_subnumber::text AS number,"
+                    "       formatWoNumber(wo_id) AS number,"
```
gilmoskowitz: A similar change should be made to the VirtualCluster constructor calls for completeness. Check the woCluster's list and search options to make sure they behave properly.

PR #1213, opened 2016-05-04
issue #27592 - tweak code to build with gcc 6

gilmoskowitz:
Tested with gcc 6 using `-Werror` and Qt5, clang Qt5, and clang Qt4. Builds on a Mac in all 3 configurations.

### DIFF
--- a/guiclient/relocateInventory.cpp
+++ b/guiclient/relocateInventory.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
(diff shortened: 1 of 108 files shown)


guiclient/relocateInventory.cpp
```diff
@@ -327,20 +327,16 @@ void relocateInventory::sShowHideDefaultToTarget()
...
-         _defaultToTarget->show();
-        //Allow default location update with correct privileges
-        if (_privileges->check("MaintainItemSites"))
-            _defaultToTarget->setEnabled(true);
-        else
-            _defaultToTarget->setEnabled(false);
-            _defaultToTarget->setChecked(false);
+        _defaultToTarget->show();
+        _defaultToTarget->setEnabled(_privileges->check("MaintainItemSites"));
+        _defaultToTarget->setChecked(false);
```
bendiy: This `setChecked(false)` default change if you have `MaintainItemSites` _looks_ fine to me, but I wanted to call it out to double check.

bendiy:
+1

guiclient/relocateInventory.cpp
gilmoskowitz: There is actually not a change to existing behavior. Note that there were no braces around the `else` clause. The indentation implied that `setChecked()` would be conditionally called; the indentation lied.

PR #1215, opened 2016-05-06
issue #25125: Refactored copyItem into two steps

paladinlogic:
This commit addresses a race condition.  When selecting an available item number for the new item, the number was being selected within a transaction block.  Until it was committed, other users could select the same item number.  The attempt to find a number by appending an integer to it was not sufficient.

In this proposed solution, I split the copy process into two steps.  After identifying what item to copy, the user must first enter an item number and click "Next".  A database call immediately checks if the item number is available before creating the new item.  When the new item is created successfully, that item exists in the database, and the number can be found by other users.  After clicking "Next", the user can decide what children to copy (item sites, BOMs, etc.)

If after clicking "Next", the user decides to cancel, changes on the second page of the screen are rolled back, and the new item is removed from the database.  A click on "Copy" on this second page of the screen completes the copy operation.

### DIFF
--- a/guiclient/copyItem.cpp
+++ b/guiclient/copyItem.cpp
@@ -102,7 +102,7 @@ void copyItem::sNext()
 
     if (saveItem())
     {
-      _pages->setSelectedIndex(1);
+      _pages->setCurrentIndex(1);
     }
   }
 }
@@ -683,7 +683,7 @@ void copyItem::cancelCopy()
   { 
     XSqlQuery query;
     query.prepare("SELECT deleteItem(:itemid)");
-    query.bind(":itemid", _newitemid);
+    query.bindValue(":itemid", _newitemid);
     query.exec();
   }
 }
--- a/guiclient/copyItem.cpp
+++ b/guiclient/copyItem.cpp
@@ -156,8 +156,8 @@ bool copyItem::saveItem()
 
     // Set item inactive until copy is committed.
 
-    itemsave.prepare("UPDATE item SET item_active=false,"
-                     "                item_descrip1=:item_descrip1"
+    itemsave.prepare("UPDATE item SET item_active=false, "
+                     "                item_descrip1=:item_descrip1 "
                      "WHERE (item_id=:item_id);");
     itemsave.bindValue(":item_id", _newitemid);
     itemsave.bindValue(":item_descrip1", _targetItemDescrip->text());
@@ -563,8 +563,8 @@ void copyItem::sCopy()
   if (! okToSave())
     return;
 
-  copyCopy.prepare("UPDATE item SET item_active=:item_active,"
-                   "                item_listprice=:item_listprice,"
+  copyCopy.prepare("UPDATE item SET item_active=:item_active, "
+                   "                item_listprice=:item_listprice, "
                    "                item_listcost=:item_listcost "
                    "WHERE (item_id=:item_id);");
   copyCopy.bindValue(":item_id", _newitemid);
--- a/guiclient/copyItem.cpp
+++ b/guiclient/copyItem.cpp
@@ -103,11 +103,23 @@ void copyItem::sNext()
 
     if (saveItem())
     {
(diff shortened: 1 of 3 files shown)


guiclient/copyItem.cpp
```diff
@@ -88,72 +92,100 @@ enum SetResponse copyItem::set(const ParameterList &pParams)
...
+    // Set item inactive until copy is committed.
+
+    itemsave.prepare("UPDATE item SET item_active=false, "
+                     "                item_descrip1=:item_descrip1 "
+                     "WHERE (item_id=:item_id);");
+    itemsave.bindValue(":item_id", _newitemid);
+    itemsave.bindValue(":item_descrip1", _targetItemDescrip->text());
+    itemsave.exec();
+
+    begin();
```
gilmoskowitz: `begin()` here puts the `copyItem`, setting the new item inactive, and copying the BOM and Itemsite in separate transactions. This feels incorrect.

guiclient/copyItem.cpp
```diff
@@ -88,72 +92,100 @@ enum SetResponse copyItem::set(const ParameterList &pParams)
...
+
+    sCopyBom();
+    sCopyItemsite();
+
+    saved = true;
   }
-  if (itemsave.lastError().type() != QSqlError::NoError)
+
+  if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Copying Item"),
+                                  itemsave, __FILE__, __LINE__))
```
gilmoskowitz: The user interaction for acknowledging errors in `sCopyBom()`, in `sCopyItemsite()`, and here all occurs within the transaction. It would be much better to roll back the transaction before reporting errors to the user.

guiclient/copyItem.cpp
```diff
@@ -88,72 +92,100 @@ enum SetResponse copyItem::set(const ParameterList &pParams)
...
   {
-    itemsave.exec("ROLLBACK;");
-    _inTransaction = false;
-    return;
+    cancelCopy();
```
gilmoskowitz: In case of a database error, `cancelCopy()` here will fail if the transaction is still open.

guiclient/copyItem.cpp
```diff
@@ -635,8 +666,7 @@ void copyItem::sCopy()
     return;
   }
 
-  copyCopy.exec("COMMIT;");
-  _inTransaction = false;
+  commit();
```
gilmoskowitz: If we're inside a transaction and errors occurred before here, we need to roll back before reporting the errors. It's easiest to understand what's going on if the `begin`, `commit`, and any `rollback`s are in the same function(s). The old version of the code didn't do that; the new version should try.

guiclient/copyItem.cpp
```diff
@@ -665,13 +744,7 @@ void copyItem::clear()
...
 void copyItem::closeEvent(QCloseEvent *pEvent)
 {
-  XSqlQuery itemcloseEvent;
-  if(_inTransaction)
-  {
-    itemcloseEvent.exec("ROLLBACK;");
-    _inTransaction = false;
-  }
-  
+  cancelCopy();
```
gilmoskowitz: This call is not guarded so it looks like it'll cancel the copy even on success.

paladinlogic:
Since copyItem was designed to have user interaction within the context of a transaction block, I could not find a way to get the begin/rollback/commit logic all into one function.  It seemed that the transaction block itself was a cause for concern with database errors potentially aborting the transaction with any user interaction.
My latest commit removes the transaction block entirely.  If the user cancels or closes the screen without saving, the BOM, item sites, and item itself are explicitly removed rather than rolled back.  The up side is that the code is shorter, simpler, easier to understand.  I can see the potential for undesirable effects if, while working in the screen, the user's computer goes down.  With the database actions happening immediately, nothing would be rolled back in that case.  If this is a problem, I can go back to commit 61bb1c1, but I don't see a way to get begin/rollback/commit all into one function.

PR #1219, opened 2016-05-11
Network fixes

bendiy:
Found some issues when working on Address Verification.

1. Add `global.engineEvaluate()` to Qt Script so we can load javascript code from outside of `pkgscript` tables and have it still showup in Qt Script Debugger. This is for js library code that doesn't have anything to do with our window scripting. It avoids name collisions and prevent the loading of multiple scripts you get with `include('foo')`.
2. Expose all the `QDnsFoo` objects so we can do `dns.lookup`.
3. Expose `QNetworkReply::NetworkError` enum.
4. Fix `QNetworkReply.rawHeader()` bug so it actually returns.
5. Expose `QNetworkRequest::foo` enums.
6. Expose `QTimerProto::setTimerType`, `QTimerProto::timerType` and `Qt::TimerType` enum

### DIFF
--- a/scriptapi/engineevaluate.cpp
+++ b/scriptapi/engineevaluate.cpp
@@ -1,0 +1,47 @@
+/*
+ * This file is part of the xTuple ERP: PostBooks Edition, a free and
+ * open source Enterprise Resource Planning software suite,
+ * Copyright (c) 1999-2016 by OpenMFG LLC, d/b/a xTuple.
+ * It is licensed to you under the Common Public Attribution License
+ * version 1.0, the full text of which(including xTuple-specific Exhibits)
+ * is available at www.xtuple.com/CPAL.  By using this software, you agree
+ * to be bound by its terms.
+ */
+
+#include "engineevaluate.h"
+#include <QtScript>
+
+QScriptValue engineEvaluateForJS(QScriptContext* context, QScriptEngine* engine)
+{
+  if (context->argumentCount() == 2 &&
+      context->argument(0).isString() &&
+      context->argument(1).isString()) {
+
+    context->setActivationObject(context->parentContext()->activationObject());
+    context->setThisObject(context->parentContext()->thisObject());
+
+    QString script = context->argument(0).toString();
+    QString scriptname = context->argument(1).toString();
+    QScriptValue result = engine->evaluate(script, scriptname, 1);
+
+    if (engine->hasUncaughtException()) {
+      qWarning() << "Uncaught exception in script: " << scriptname
+                 << " at line: "
+                 << engine->uncaughtExceptionLineNumber() << ":"
+                 << result.toString();
+      return engine->toScriptValue(false);
+    } else {
+      return result;
+    }
+  } else {
+    context->throwError(QScriptContext::UnknownError,
+                          "The engineEvaluate() method takes two arguments.");
+    return engine->toScriptValue(false);
+  }
+}
+
+void setupEngineEvaluate(QScriptEngine *engine)
+{
+  QScriptValue engineEvaluate = engine->newFunction(engineEvaluateForJS);
+  engine->globalObject().setProperty("engineEvaluate", engineEvaluate);
+}
--- a/scriptapi/qdnsdomainnamerecordproto.cpp
+++ b/scriptapi/qdnsdomainnamerecordproto.cpp
@@ -28,14 +28,20 @@ QScriptValue QListQDnsDomainNameRecordToScriptValue(QScriptEngine *engine, const
 void QListQDnsDomainNameRecordFromScriptValue(const QScriptValue &obj, QList<QDnsDomainNameRecord> &list)
 {
   list = QList<QDnsDomainNameRecord>();
-  QScriptValueIterator it(obj);
 
-  while (it.hasNext()) {
-    it.next();
-    if (it.flags() & QScriptValue::SkipInEnumeration)
-      continue;
-    QDnsDomainNameRecord item = qscriptvalue_cast<QDnsDomainNameRecord>(it.value());
-    list.insert(it.name().toInt(), item);
+  if (obj.isArray()) {
+    QScriptValueIterator it(obj);
+
+    while (it.hasNext()) {
+      it.next();
+      if (it.flags() & QScriptValue::SkipInEnumeration)
+        continue;
+
+      if (it.value().isString()) {
+        QDnsDomainNameRecord item = qscriptvalue_cast<QDnsDomainNameRecord>(it.value());
+        list.insert(it.name().toInt(), item);
+      }
+    }
   }
 }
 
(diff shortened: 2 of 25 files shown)


scriptapi/engineevaluate.cpp
```diff
@@ -0,0 +1,47 @@
...
+                          "The engineEvaluate() method takes two arguments.");
+    return engine->toScriptValue(false);
+  }
+}
+
+void setupEngineEvaluate(QScriptEngine *engine)
+{
+  QScriptValue engineEvaluate = engine->newFunction(engineEvaluateForJS);
+  engine->globalObject().setProperty("engineEvaluate", engineEvaluate);
+}
```
gilmoskowitz: What are the benefits of this wrapper vs. calling `eval()`?

scriptapi/qdnsdomainnamerecordproto.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include "qdnsdomainnamerecordproto.h"
+
+#if QT_VERSION < 0x050000
+void setupQDnsDomainNameRecordProto(QScriptEngine *engine)
+{
+  Q_UNUSED(engine);
+}
+#else
+
+QScriptValue QListQDnsDomainNameRecordToScriptValue(QScriptEngine *engine, const QList<QDnsDomainNameRecord> &list)
```
gilmoskowitz: Do we really have to get this low level? Why should scripts have to work with DNS name records?

scriptapi/qdnslookupproto.cpp
```diff
@@ -0,0 +1,250 @@
...
+}
+
+QScriptValue TypeToScriptValue(QScriptEngine *engine, const QDnsLookup::Type &item)
+{
+  return engine->newVariant(item);
+}
+void TypeFromScriptValue(const QScriptValue &obj, QDnsLookup::Type &item)
+{
+  item = (QDnsLookup::Type)obj.toInt32();
+}
```
gilmoskowitz: These 4 conversion functions should have less ambiguous names to avoid potential future conflicts.

scriptapi/qdnsdomainnamerecordproto.cpp
```diff
@@ -0,0 +1,112 @@
...
+    newArray.setProperty(i, engine->toScriptValue(list.at(i)));
+  }
+  return newArray;
+}
+void QListQDnsDomainNameRecordFromScriptValue(const QScriptValue &obj, QList<QDnsDomainNameRecord> &list)
+{
+  list = QList<QDnsDomainNameRecord>();
+  QScriptValueIterator it(obj);
+
+  while (it.hasNext()) {
```
gilmoskowitz: This iterates through the properties of `obj`. What if `obj` isn't an array? What if `obj` is an array with additional properties added?

scriptapi/qdnshostaddressrecordproto.cpp
```diff
@@ -0,0 +1,112 @@
...
+    newArray.setProperty(i, engine->toScriptValue(list.at(i)));
+  }
+  return newArray;
+}
+void QListQDnsHostAddressRecordFromScriptValue(const QScriptValue &obj, QList<QDnsHostAddressRecord> &list)
+{
+  list = QList<QDnsHostAddressRecord>();
+  QScriptValueIterator it(obj);
+
+  while (it.hasNext()) {
```
gilmoskowitz: Same here as above - what if `obj` doesn't match expectations?

scriptapi/qdnsmailexchangerecordproto.cpp
```diff
@@ -0,0 +1,120 @@
...
+    newArray.setProperty(i, engine->toScriptValue(list.at(i)));
+  }
+  return newArray;
+}
+void QListQDnsMailExchangeRecordFromScriptValue(const QScriptValue &obj, QList<QDnsMailExchangeRecord> &list)
+{
+  list = QList<QDnsMailExchangeRecord>();
+  QScriptValueIterator it(obj);
+
+  while (it.hasNext()) {
```
gilmoskowitz: And here...

scriptapi/qnetworkrequestproto.cpp
```diff
@@ -18,15 +18,114 @@
...
 #define DEBUG false
 
+QScriptValue AttributeToScriptValue(QScriptEngine *engine, const QNetworkRequest::Attribute &item)
+{
+  return engine->newVariant(item);
+}
+void AttributeFromScriptValue(const QScriptValue &obj, QNetworkRequest::Attribute &item)
+{
+  item = (QNetworkRequest::Attribute)obj.toInt32();
+}
```
gilmoskowitz: Make these function names less ambiguous. Too many kinds of object have "attributes".

scriptapi/qnetworkrequestproto.cpp
```diff
@@ -18,15 +18,114 @@
...
+void LoadControlFromScriptValue(const QScriptValue &obj, QNetworkRequest::LoadControl &item)
+{
+  item = (QNetworkRequest::LoadControl)obj.toInt32();
+}
+
+QScriptValue PriorityToScriptValue(QScriptEngine *engine, const QNetworkRequest::Priority &item)
+{
+  return engine->newVariant(item);
+}
+void PriorityFromScriptValue(const QScriptValue &obj, QNetworkRequest::Priority &item)
```
gilmoskowitz: Make these function names less ambiguous. Too many kinds of object have "priority".

scriptapi/engineevaluate.cpp
bendiy: I tried `eval()` originally. The issue with `eval()` is that the code that is evaluated does not show up in Qt Script Debugger.  The Qt Script Engine runs the javascript, but it doesn't list the file in Qt Script Debugger, you can't do any step debugging on it and any errors the code might throw do not have traces so you can't tell what is causing the error. We get all of that back by passing the code through `QScriptEngine::evaluate()`.

Our `include()` function does the same thing, but has some downsides I ran into.

- It includes multiple script files from the `pckscript` table and evaluates each one if there is more than one with the same name.
- It tries to load scripts based on the window name. e.g. `item`. If I have a library to load, I can load it explicitly with this change and not risk a naming collision with other `pckscript` code.

scriptapi/qdnsdomainnamerecordproto.cpp
bendiy: The function I'm after here is `void QDnsLookup::lookup()`. All the other QDnsFoo objects are here to support exposing the `QDnsLookup` class. The lookup is to support these:
https://nodejs.org/dist/latest-v4.x/docs/api/net.html#net_event_lookup
https://nodejs.org/dist/latest-v4.x/docs/api/net.html#net_socket_remoteaddress

I had it as a `TODO` on my socket wrapper which I cleaned up and finalized last week when I put everything in an xTuple extension package.

PR #1233, opened 2016-05-23
issue #27953 - try to avoid crashes

gilmoskowitz:
Try to avoid crashes when the Qt 5 desktop client is run by Squish by removing unused code that appears near the top of the stack trace.


gilmoskowitz:
Both FrogLogic and ZenQ have verified that this code change eliminates the crashes when running the application with Squish.

PR #1235, opened 2016-05-25
Issue #22170: Work-around for copy of large number of rows of rich data from XTreeWidget

paladinlogic:
The root cause of this issue is that the number of bytes needed to perform the copy of the rich data to the clipboard was exceeding the amount of memory addressable by a 32 bit integer, thus causing xTuple to crash.  The most straightforward solution would seem to be to build xTuple as a 64 bit binary.

Barring this, I looked for some way to determine how much addressable space was available, but did not find any approach that would work cross-platform.

What is presented here is a work-around that I feel leaves something to be desired but does address the issue.  I introduced a new preference and placed it on the userPreferences screen beside the choice of copying rich or plain text.  The new preference puts an approximate limit how many bytes of data the XTreeWidget is allowed to export as HTML.  By default the value is 0 GB, which means no limit (which would probably be appropriate for a 64 bit xTuple and probably for the majority of users out there who don't copy a ton of data to the clipboard).  If a crash is observed, the user can modify the preference to limit the amount of data the XTreeWidget will copy.

The process is essentially this:
* Copy the headers and include them in the data count
* Start adding rows until the data count exceeds the limit set
* If the limit is reached, a message box is displayed showing the user how many rows (X of Y) were processed.  No more row data is added, and the copy to clipboard is attempted.

The data count is the current sum of the character length of the data that has been added.  It does NOT include HTML formatting bytes, so the limit is strictly on the text length of the data being copied.

It is still possible to crash the 32 bit client if the limit is not set or is set too high and too much data are added.

### DIFF
--- a/widgets/xtreewidget.cpp
+++ b/widgets/xtreewidget.cpp
@@ -2631,7 +2631,7 @@ QString XTreeWidget::toHtml() const
     if (maxDataCount > 0 && dataCount > maxDataCount)
     {
       QString overflowMsg = tr("Maximum data limit was encountered.  Only %1 of %2 rows could be processed.");
-      QMessageBox::warning(this, tr("Data Limit Reached"), overflowMsg.arg(row).arg(rowcnt));
+      QMessageBox::warning(NULL, tr("Data Limit Reached"), overflowMsg.arg(row).arg(rowcnt));
     }
   }
   return doc->toHtml();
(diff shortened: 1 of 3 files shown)


andersdd:
This is an excellent analysis and explanation of a problem I have seen regularly at sites, and not been able to reproduce.  Most likely due to the Windows client being 32bit and my Linux client being 64bit.

Thanks!!

widgets/xtreewidget.cpp
```diff
@@ -2611,10 +2624,18 @@ QString XTreeWidget::toHtml() const
             cell.setFormat(format);
             cursor->insertText(item->text(counter));
             cursor->movePosition(QTextCursor::NextCell);
+
+            dataCount += (qlonglong)(header->text(counter).size());
```
gilmoskowitz: This should be the size of `item->text()`.

PR #1236, opened 2016-05-26
issue #27975 - reduce crash risk by changing startup and init sequence

gilmoskowitz:
The first commit comment refers to the wrong bug number.


davidbeauchamp:
+1 once conflicts are resolved

PR #1240, opened 2016-05-27
Tidy up (clarify) user privilege screen

andersdd:
Could be a little confusing and mistaken for a maintenance screen.


gilmoskowitz:
@piercetyler - you might want to update the docs for this tweak

PR #1243, opened 2016-05-30
Issue #27490: Profit Centre setup and delete issues

andersdd:


guiclient/profitCenters.cpp
```diff
@@ -100,21 +100,15 @@ void profitCenters::sDelete()
...
-    int result = profitDelete.value("result").toInt();
-    {
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Deleting Profit Center"),
-                             storedProcErrorLookup("deleteProfitCenter", result),
-                             __FILE__, __LINE__);
-      return;
-    }
+    QMessageBox::critical( this, tr("Cannot Delete Profit Center"),
+          tr("Cannot delete this Profit Center as it exists in General Ledger Accounts") );
+    return;
```
gilmoskowitz: The `ErrorReporter::error()` call here is correct as it allows translating based on the value of `result`. The block it's in should be protected by `if (result < 0)`.

gilmoskowitz:
see also https://github.com/xtuple/xtuple/pull/2711

guiclient/profitCenters.cpp
andersdd: @gilmoskowitz  Not sure I agree (or understand) your comment.  The message _is_ wrapped in a `result < 0`  (see line 103) and there is only one error reason in the function.  If something bombs completely the following lines pick up any generic sql error.

gilmoskowitz:
```
if (profitDelete.first())
{
  int result = ...;
  if (result < 0)
  {
    ErrorReporter::error(...);
    return;
  }
}
```
This allows `ErrorReporter::error()` to do the correct thing regardless of how we change `deleteProfitCenter()`. There's only one possible error code now but that may change. In addition, we want to use `ErrorReporter::error()` as much as possible. It collects and reports as much data as it can and will eventually be able to send the results to us.

PR #1255, opened 2016-06-06
issue #24842 Add Project Accounting to Cash Receipts

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#2715

guiclient/cashReceipt.cpp
```diff
@@ -957,6 +962,10 @@ bool cashReceipt::save(bool partial)
     else
       cashave.bindValue(":cashrcpt_alt_curr_rate", _exchRate->toDouble());
   }
+  if(_project->isValid())
+    cashave.bindValue(":cashrcpt_prj_id", _project->id());
+  else
+    cashave.bindValue(":cashrcpt_prj_id", -1);
```
gilmoskowitz: This will raise a database error when `_project` is not valid, as `-1` is not a valid foreign key value. Removing the `else` case should fix this.

PR #1260, opened 2016-06-09
Issue #28049: Add Hold Status to Customers Display and parameters

andersdd:
requires xtuple/xtuple#2721


gilmoskowitz:
+1
merge after xtuple/xtuple#2721 is tweaked

PR #1268, opened 2016-06-15
issue #28029 Check for valid work week when changing DateCluster Site

ChristopherCotnoir:


gilmoskowitz:
see also https://github.com/xtuple/qt-client/pull/1237

PR #1271, opened 2016-06-16
Issue #28041:add column and filter for create date

garyhgohoos:


gilmoskowitz:
works with https://github.com/xtuple/xtuple/pull/2732

PR #1272, opened 2016-06-16
Removed deprecated method call for Qt5

cptaffe:
Removed `QSslCertificate::isValid()` use in builds using Qt5 or above.


gilmoskowitz:
@cptaffe - Thanks for your contribution. Can you explain why you would like to patch 4_9_x? I'm happy to merge this if there's a compelling reason.

We're not anticipating another 4.9 release and do not expect 4_9_x to compile with Qt5. This issue has been [fixed in 4_10_x](https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/creditcardprocessor.cpp#L345-L368) but not as concisely as your solution.

cptaffe:
@gilmoskowitz You're welcome.

We use 4.9.1 currently where I work. The nightly packages were using libraries that were too old for my system, so I just wanted to get it working so I can use the UI from my laptop. Just guessed there might be other Linux users who would rather compile it than go searching for old library versions.

PR #1274, opened 2016-06-21
Use QHostAddress only.

bendiy:
Found this when testing dashboards on linux. Worked find on windows, but this is a bug.
https://doc.qt.io/qt-4.8/qtcpserver.html#listen
https://doc.qt.io/qt-5/qtcpserver.html#listen


davidbeauchamp:
Merging #1275 instead

PR #1278, opened 2016-06-23
Issue #28134: Add expand/collapse buttons to display class controllable from C++ and script

davidbeauchamp:
Buttons are hidden by default. To enable In C++ display class window:
`setExpandVisible(true);`
`setCollapseVisible(true);`

In QtScript:
`mywindow.setExpandVisible(true);`
`mywindow.setCollapseVisible(true);`

### DIFF
--- a/guiclient/display.cpp
+++ b/guiclient/display.cpp
@@ -82,6 +82,18 @@ public:
     _sep2 = _toolBar->addSeparator();
     _sep2->setVisible(false);
 
+    _expandBtn = new QToolButton(_toolBar);
+    _expandBtn->setObjectName("_expandBtn");
+    _expandBtn->setFocusPolicy(Qt::NoFocus);
+    _expandAct = _toolBar->addWidget(_expandBtn);
+    _expandAct->setVisible(false);
+
+    _collapseBtn = new QToolButton(_toolBar);
+    _collapseBtn->setObjectName("_expandBtn");
+    _collapseBtn->setFocusPolicy(Qt::NoFocus);
+    _collapseAct = _toolBar->addWidget(_collapseBtn);
+    _collapseAct->setVisible(false);
+
     // Print buttons
     _printBtn = new QToolButton(_toolBar);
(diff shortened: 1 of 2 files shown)


guiclient/display.cpp
```diff
@@ -82,6 +82,18 @@ class displayPrivate : public Ui::display
...
+    _expandBtn->setObjectName("_expandBtn");
+    _expandBtn->setFocusPolicy(Qt::NoFocus);
+    _expandAct = _toolBar->addWidget(_expandBtn);
+    _expandAct->setVisible(false);
+
+    _collapseBtn = new QToolButton(_toolBar);
+    _collapseBtn->setObjectName("_collapseBtn");
+    _collapseBtn->setFocusPolicy(Qt::NoFocus);
+    _collapseAct = _toolBar->addWidget(_collapseBtn);
+    _collapseAct->setVisible(false);
```
gilmoskowitz: Why default to invisible?

guiclient/display.cpp
```diff
@@ -589,6 +623,26 @@ bool display::closeVisible() const
...
+void display::setCollapseVisible(bool show)
+{
+  _data->_collapseAct->setVisible(show);
+}
+
+bool display::collapseVisible() const
+{
+  return _data->_collapseAct->isVisible();
+}
+
```
gilmoskowitz: Does it make sense to show/hide these buttons separately?

guiclient/display.cpp
davidbeauchamp: Most display class windows don't use an indented query, and while there is no harm in calling expandAll()/collapseAll() on a list with no children I felt they would be confusing to end users who would expect them to do something. This way the dev can turn it on (or use the two line script to add them to any core window)

guiclient/display.cpp
davidbeauchamp: I was conflicted about this, but I just know someone is going to want them individually controllable. One thing I considered was adding a setExpandAndCollapseVisible(bool) as a convenience function.

PR #1284, opened 2016-06-24
Issue #28058:auto fill list if order id argument exists

garyhgohoos:


guiclient/displays/dspSalesOrderStatus.cpp
```diff
@@ -60,7 +60,10 @@ enum SetResponse dspSalesOrderStatus::set(const ParameterList &pParams)
 
   param = pParams.value("sohead_id", &valid);
   if (valid)
+  {
     _so->setId(param.toInt());
+    sFillList();
```
gilmoskowitz: Calling [`setQueryOnStartEnabled(true);`](https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/display.cpp#L672-L706) would be more consistent with other displays.

PR #1285, opened 2016-06-27
Issue #28138: Add widget extractor to script editor window

davidbeauchamp:
Add button to script editor window to allow generating "var _foo = mywindow.findChild("_foo");" statements from a matching screen. If multiple matches are found with the same name as the script present a list and allow the user to choose.


guiclient/scriptEditor.ui
```diff
@@ -55,10 +55,10 @@ to be bound by its terms.</comment>
      <property name="enabled">
       <bool>false</bool>
      </property>
-     <property name="allowNull">
+     <property name="allowNull" stdset="0">
```
davidbeauchamp: Qt Creator 4.0 kept adding 'stdset="0"' to the uiform when saved, not sure what they represent but they only appeared on custom widgets.

guiclient/scriptEditor.cpp
```diff
@@ -481,3 +499,79 @@ void scriptEditor::sPositionChanged()
...
+  }
+  QDomDocument report;
+  report.setContent(source);
+
+  QDomNodeList widgets = report.elementsByTagName("widget");
+  for (int i = 0; i < widgets.size(); i += 1) {
+    QDomNode domNode = widgets.item(i);
+    QDomElement element = domNode.toElement();
+    // skip the root widget
+    if (element.attribute("class") == "QWidget" || element.attribute("class") == "QDialog")
```
davidbeauchamp: I should probably exclude XWidget and XDialog too..

PR #1293, opened 2016-06-30
Issue #28058:auto fill when order id is passed

garyhgohoos:


garyhgohoos:
for some reason, zen testing does not like setQueryOnStart.  switched to sFillList

gilmoskowitz:
The application handles Query on Start properly for me.

PR #1296, opened 2016-07-01
Expose QSqlQuery, QSqlError and QSql to Qt Script

bendiy:
I'm trying to share code between Node.js and Qt Script. My shim for `XT.dataSource.query()` works fine for basic queries, but not anything with parameters. `XSqlQuery` does not expose `prepare`, `bindValue` or `addBindValue`, so the only way to do parameterized queries is with MetaSQL, but `XT.dataSource.query()` does not support MetaSQL.

By adding `QSqlQuery`, I can now do this in my `XT.dataSource.query()` shim:
```
var foo = new QSqlQuery();
foo.prepare("SELECT cust_id FROM custinfo WHERE cust_number = ?");
foo.addBindValue("TTOYS");
foo.exec();
foo.first();
var bar = foo.value("cust_id"); //95
```

By string replacing the node-postgres `$1` style parameters with `?`, the `XT.dataSource.query()` shim can work in Qt Script with queries from Node.js.

`QSql` is needed for the `emun`s and `QSqlError` is needed to handle errors from `QSqlQuery`.


scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+  } else if (context->argumentCount() == 1) {
+    if (context->argument(0).isString()) {
+      obj = new QSqlQuery(context->argument(0).toString());
+    } else if (context->argument(0).isObject()){
+      // TODO: Need to expose QSqlResult for this to work.
+      //try {
+      //  QSqlResult *result = qscriptvalue_cast<QSqlResult*>(context->argument(0));
+      //  obj = new QSqlQuery(result);
+      //} catch (...) {
+        try {
```
gilmoskowitz: We don't use try/catch blocks in our C++

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+  if (item)
+    item->addBindValue(val, paramType);
+}
+
+int QSqlQueryProto::at() const
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->at();
+  return 0;
```
gilmoskowitz: See http://doc.qt.io/qt-5/qsqlquery.html#at - `0` is probably the wrong value to return here.

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+bool QSqlQueryProto::exec(const QString &query)
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->exec(query);
+  return false;
+}
+
+bool QSqlQueryProto::exec()
```
gilmoskowitz: Test if the JS/C++ interface handles the overloads correctly or complains about ambiguity.

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+    return item->isActive();
+  return false;
+}
+
+bool QSqlQueryProto::isForwardOnly() const
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->isForwardOnly();
+  return false;
```
gilmoskowitz: Safer to return `true` here.

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+}
+
+bool QSqlQueryProto::isNull(int field) const
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->isNull(field);
+  return false;
```
gilmoskowitz: Safer to return `true` here???

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+    return item->last();
+  return false;
+}
+
+QSqlError QSqlQueryProto::lastError() const
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->lastError();
+  return QSqlError();
```
gilmoskowitz: Better to return [`QSqlError::UnknownError`](http://doc.qt.io/qt-5/qsqlerror.html#ErrorType-enum) here

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,408 @@
...
+    return item->nextResult();
+  return false;
+}
+
+int QSqlQueryProto::numRowsAffected() const
+{
+  QSqlQuery *item = qscriptvalue_cast<QSqlQuery*>(thisObject());
+  if (item)
+    return item->numRowsAffected();
+  return 0;
```
gilmoskowitz: Better to return [`-1`](http://doc.qt.io/qt-5/qsqlquery.html#numRowsAffected)

gilmoskowitz:
Build is failing with
```
In file included from setupscriptapi.cpp:106:0:
qsqlproto.h:14:16: fatal error: QSql: No such file or directory
```

scriptapi/qsqlqueryproto.cpp
bendiy: I added some `qDebug()` messages here and the correct function is called when `foo.exec();` or `foo.exec("SELECT CURRENT_USER");` is called. This should work fine.

scriptapi/qsqlproto.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef __QSQLPROTO_H__
+#define __QSQLPROTO_H__
+
+#include <QScriptEngine>
+
+#if QT_VERSION >= 0x050000
+#include <QSql>
+#else
+#include <QtSql>
+#endif
```
bendiy: Fix build error. In Qt 4.8 it is `QtSql`.

scriptapi/qsqlqueryproto.cpp
```diff
@@ -0,0 +1,433 @@
...
+      if (connectionNameFunction.isFunction()) {
+        QScriptValue connectionName = connectionNameFunction.call(context->argument(0));
+        QSqlDatabase dbItem = QSqlDatabase::database(connectionName.toString());
+        if (dbItem.isValid()) {
+          obj = new QSqlQuery(dbItem);
+        } else {
+          context->throwError(QScriptContext::UnknownError,
+            "Could not construct a QSqlQuery with the supplied QSqlDatabase.");
+        }
+      } else {
```
bendiy: @gilmoskowitz My guess here is that `qscriptvalue_cast<QSqlDatabase*>(context->argument(0))` doesn't work with our QSqlDatabase to/from functions here:
https://github.com/xtuple/qt-client/blob/4_10_x/scriptapi/qsqldatabaseproto.cpp#L34-L44

They use `item = obj.toVariant().value<QSqlDatabase*>();`, which I don't think really gives us a QSqlDatabase. That is pretty old code. I've found that we don't need to/from functions most of the time as Qt just figures it out. I didn't want to break anything existing that might be using QSqlDatabase, so the above work around is used.

My testing works after this change. If CI passes, this is all I need for beta2.

PR #1300, opened 2016-07-08
issue #28156 Refactor todoList privilege checking

ChristopherCotnoir:


guiclient/incident.cpp
```diff
@@ -24,6 +24,47 @@
 
 #include <openreports.h>
 
+bool incident::userHasPriv(const int pMode, const int pId)
+{
+  bool allPriv = _privileges->check("MaintainAllIncidents");
+  bool personalPriv = _privileges->check("MaintainPersonalIncidents");
```
gilmoskowitz: This method can be simplified if we simply
```
if (_privileges->check("MaintainAllIncidents"))
  return true;
```
here at the top.

guiclient/incident.cpp
```diff
@@ -24,6 +24,47 @@
...
+  {
+    allPriv = allPriv || _privileges->check("ViewAllIncidents");
+    personalPriv = personalPriv || _privileges->check("ViewPersonalIncidents");
+  }
+
+  if(pMode==cNew)
+    if(allPriv||personalPriv)
+      return true;
+    else
+      return false;
```
gilmoskowitz: In cases like this, where we're going to return one or another boolean value based on a simple conditional check, just return the "truthiness" of the check:
```c++
if (pMode == cNew) {
  return personalPriv;
} else {
  ...
}
```

guiclient/customer.h
```diff
@@ -37,6 +37,8 @@ class customer : public XWidget, public Ui::customer
     customer(QWidget* parent = 0, const char* name = 0, Qt::WindowFlags fl = Qt::Window);
     ~customer();
 
+    static bool userHasPriv(const int = cView);
```
gilmoskowitz: For API consistency, add a `const int id = 0` parameter here, too.

guiclient/incident.cpp
```diff
@@ -24,6 +24,47 @@
...
+  else
+  {
+    bool isOwner = false;
+    bool isAssigned = false;
+
+    XSqlQuery usernameCheck;
+    usernameCheck.prepare( "SELECT incdt_owner_username AS owner, "
+                           " incdt_assigned_username AS assigned "
+                           "FROM incdt "
+                            "WHERE (incdt_id=:incdt_id);" );
```
gilmoskowitz: Let's ask the question we want answered: Is the user allowed to change this incident?
```sql
select :username in (incdt_owner_username, incdt_assigned_username) as canModify
  from incdt
 where (incdt_id = :incdt_id);
```
or even
```sql
select getEffectiveXtUser() in (incdt_owner_username, incdt_assigned_username) as canModify
  from incdt
 where (incdt_id = :incdt_id);
```

PR #1301, opened 2016-07-08
Issue #28200:add option for dspSingleLevelBill

garyhgohoos:


guiclient/itemAvailabilityWorkbench.cpp
```diff
@@ -247,7 +256,12 @@ itemAvailabilityWorkbench::itemAvailabilityWorkbench(QWidget* parent, const char
     if (!_privileges->check("ViewCosts"))
       _costedIndentedBOMButton->hide();
     if (!_privileges->check("ViewBOMs"))
+    {
+      _singleLevelBOMButton->hide();
       _whereUsedButton->hide();
+      _costedIndentedBOMButton->setChecked(true);
```
gilmoskowitz: Shouldn't this be `_singleLevelBOMButton`?

guiclient/itemAvailabilityWorkbench.cpp
```diff
@@ -283,10 +297,12 @@ void itemAvailabilityWorkbench::sHandleButtons()
...
     _availabilityStack->setCurrentIndex(2);
   
-  if (_costedIndentedBOMButton->isChecked())
+  if (_singleLevelBOMButton->isChecked())
     _bomStack->setCurrentIndex(0);
-  else if (_whereUsedButton->isChecked())
+  else if (_costedIndentedBOMButton->isChecked())
     _bomStack->setCurrentIndex(1);
+  else if (_whereUsedButton->isChecked())
+    _bomStack->setCurrentIndex(2);
```
gilmoskowitz: Picking the right `_bomStack` page should be based on the widget that page contains, not assumptions about the integer index of the page in the stack.

guiclient/itemAvailabilityWorkbench.cpp
garyhgohoos: looks ok.  _costedIndentedBOMButton is checked if the other buttons are hidden

PR #1309, opened 2016-07-12
updated translation files for release 4.10.0Beta2

ChristopherCotnoir:


gilmoskowitz:
The change to `share/dict/xTuple.base.ts` is too big - it adds lines from `ui_*.cpp` files. Please `make clean` before running `lupdate` and add that step to the documentation.

PR #1310, opened 2016-07-12
issue #25942 - load _all_ of the yp certs + small refactor to help

gilmoskowitz:
This resolves the private key error reported by ZenQ. It also moves most of the YourPay-specific code out of the generic CreditCardProcessor class.


guiclient/creditcardprocessor.h
```diff
@@ -126,6 +127,7 @@ class CreditCardProcessor : public QObject
     QString		_ppassword;
     QString		_pport;
     QString		_pserver;
+    QString             _pemfile;
```
bendiy: This `_pemfile` will only be set in `yourpayprocessor.cpp`?

gilmoskowitz: At least for now, that's correct. This allows separating the PEM file handling from specific reference to YourPay/FirstData and opens us up for the possibility of another service doing something similar.

guiclient/yourpayprocessor.cpp
```diff
@@ -76,6 +76,52 @@ YourPayProcessor::YourPayProcessor() : CreditCardProcessor()
...
+                              "This may cause communication problems.")
+                           .arg(_pemfile));
+    else
+    {
+      QSslConfiguration config = QSslConfiguration::defaultConfiguration();
+      QList<QSslCertificate> certs = config.caCertificates();
+      foreach(QSslCertificate c, certlist)
+      {
+        if (certificateIsValid(&c))
+          certs.append(c);
```
bendiy: What is `c` is already in `certs` list? This will create duplicates.

bendiy: ... each time getCreditcardProcessor() is called I believe.

bendiy:
+1 LGTM other than the `foreach` `certs.append(c);` possibly creating duplicates.

guiclient/creditcardprocessor.h
```diff
@@ -21,6 +21,8 @@
 #endif
 #include <parameter.h>
 
+class QSslCertificate;
```
bendiy: Still don't understand when you need `class Foo;` or not, but I think here, the CI error is:
```
error: forward declaration of ‘struct QSslCertificate’
```

Because you have a function that takes `QSslCertificate` now, this like just needs to be:
```
#include <QSslCertificate>
```
Might be wrong though.

guiclient/yourpayprocessor.cpp
gilmoskowitz: good catch

PR #1317, opened 2016-07-19
issue #28257 Change Cancel functionality to Clear for last open SSO for users with…

ChristopherCotnoir:
… only access to SSO

### DIFF
--- a/guiclient/salesOrderSimple.cpp
+++ b/guiclient/salesOrderSimple.cpp
@@ -1200,27 +1200,7 @@ void salesOrderSimple::closeEvent(QCloseEvent *pEvent)
   if(!_closeThis)
   {
     pEvent->ignore();
-
-    _saved = false;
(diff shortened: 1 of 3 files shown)


guiclient/salesOrderSimple.cpp
```diff
@@ -1179,6 +1197,33 @@ void salesOrderSimple::closeEvent(QCloseEvent *pEvent)
...
+    _bankaccnt->setType(XComboBox::ARBankAccounts);
+
+    connect(_orderNumber, SIGNAL(editingFinished()), this, SLOT(sHandleOrderNumber()));
+    ParameterList params;
+    params.append("mode", "new");
+    set(params);
+
+    return;
+  }
+
```
gilmoskowitz: Call `prepare()` here instead of duplicating the logic?

PR #1318, opened 2016-07-19
issue #28215 Add option to delete characteristics when changing customer type

ChristopherCotnoir:


guiclient/customer.cpp
```diff
@@ -660,6 +674,19 @@ bool customer::sSave()
...
 
   //Save characteristics
+  if (charProfile && QMessageBox::question(this, tr("Delete?"),
+                            tr("Do you want to DELETE all characteristics related to the old Customer Type?"),
+                            QMessageBox::Yes,
+                            QMessageBox::No | QMessageBox::Default) == QMessageBox::Yes)
+  {
+    XSqlQuery deleteChar;
+
+    deleteChar.prepare("DELETE FROM charass USING (SELECT DISTINCT b.charass_id AS id FROM custtype JOIN charass a ON (a.charass_target_type='CT') AND (a.charass_target_id=custtype_id) JOIN charass b ON (b.charass_target_type='C') AND (b.charass_target_id=:cust_id) AND (a.charass_char_id=b.charass_char_id) WHERE (custtype_id=:custtype_id)) qry WHERE charass_id=qry.id;");
```
gilmoskowitz: Please break this string onto multiple lines. I can't tell yet if this is deleting the proper `charass` records.

guiclient/customer.cpp
```diff
@@ -660,6 +674,19 @@ bool customer::sSave()
   }
 
   //Save characteristics
+  if (charProfile && QMessageBox::question(this, tr("Delete?"),
+                            tr("Do you want to DELETE all characteristics related to the old Customer Type?"),
+                            QMessageBox::Yes,
+                            QMessageBox::No | QMessageBox::Default) == QMessageBox::Yes)
```
gilmoskowitz: This is an old version of `question()`. The [new version](http://doc.qt.io/qt-5/qmessagebox.html#question) uses bit-or of the allowed answers followed by the default:
```
... ::question(..., QMessageBox::Yes | QMessageBox::No, QMessageBox::No) ...
```

PR #1323, opened 2016-07-22
issue #28283 Allow colons in text filters

ChristopherCotnoir:


widgets/parameterwidget.cpp
```diff
@@ -722,7 +724,15 @@ void ParameterWidget::applySaved(int pId, int filter_id)
...
-              lineEdit->setText(tempFilterList[1]);
+              QString text = "";
+              int i;
+              for (i=1; i<tempFilterList.size()-1; i++)
+              {
+                text+=tempFilterList[i];
+                if (i!=tempFilterList.size()-2)
+                  text+=":";
+              }
+              lineEdit->setText(text);
```
gilmoskowitz: `lineEdit->setText(tempFilterList.mid(...).join(":"))`? See http://doc.qt.io/qt-5/qlist.html#mid

gilmoskowitz:
I think CI is failing because `mid()` returns a QList and the compiler doesn't recognize it as a QStringList:-(

PR #1326, opened 2016-07-25
issue #28327 - allow multiple scripts to be edited at once

davidbeauchamp:
Nothing would stop you from opening the same script more than once, should I modify the window to use AppLock?


andersdd:
+1 on this feature - much needed.  I would also lock the script for editing is already open otherwise you could end up editing two versions differently.

davidbeauchamp:
Added locking to script, if it cant acquire lock it sets the window into view mode. Also addresses oversight with extractWidgets button not being disabled in view mode.

PR #1338, opened 2016-08-05
Issue28345: Created a scripting API for QButtonGroup

paladinlogic:
Created a scripting API for QButtonGroup.  The specific case where we needed scripting exposure of this class is on the customer workbench, namely the settingsButtonGroup.  For context, this is an issue we created in order to have a branch against which to contribute the change back.

Since this issue is not yet assigned to paladin, I was not able to make any updates to the incident in dogfood.

### DIFF
--- a/scriptapi/qbuttongroupproto.cpp
+++ b/scriptapi/qbuttongroupproto.cpp
@@ -12,6 +12,7 @@
 
 #include <QScriptEngine>
 #include <QButtonGroup>
+#include <QAbstractButton>
 
 QScriptValue QButtonGrouptoScriptValue(QScriptEngine *engine, QButtonGroup* const &item)
 {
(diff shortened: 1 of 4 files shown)


scriptapi/qbuttongroupproto.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  QScriptValue constructor = engine->newFunction(constructQButtonGroup, proto);
+
+  engine->globalObject().setProperty("QButtonGroup", constructor);
+}
+
+QScriptValue constructQButtonGroup(QScriptContext *context, QScriptEngine *engine)
+{
+  QButtonGroup *obj = 0;
+  if (context->argumentCount() == 1)
```
davidbeauchamp: I would add a "&& context->argument(0).isObject()" check here just in case

paladinlogic: I will add that check.  I based this script api on qdialogbuttonbox, which does not check that the arguments are objects in constructQDialogButtonBox.

PR #1339, opened 2016-08-08
Issue 28349: Script-extensible implementation of a VirtualCluster

davidbeauchamp:
Script-extensible implementation of a VirtualCluster and VirtualClusterLineEdit. Create a screen in the screen designer, find "ScriptableCluster" in the Widget Box and add one to the screen. If you leave the default name, you can use a script like the one below to create a cluster from the ITEMGRP entry in the source table :
```javascript
var scriptableCluster = mywindow.findChild("scriptableCluster"); // ScriptableCluster
scriptableCluster.loadFromSourceType("ITEMGRP");
scriptableCluster.setTitles("Item Group", "Item Groups");
scriptableCluster.setNameVisible(true);
scriptableCluster.setDescriptionVisible(true);
```
Alternatively, if you want to load data from a table you specify:
```javascript
var scriptableCluster = mywindow.findChild("scriptableCluster"); // ScriptableCluster
scriptableCluster.setTableAndColumnNames("itemgrp", "itemgrp_id", "itemgrp_name", "itemgrp_name", "itemgrp_descrip", "");
scriptableCluster.setUiName("itemGroup");
scriptableCluster.setTitles("Item Group", "Item Groups");
scriptableCluster.setNameVisible(true);
scriptableCluster.setDescriptionVisible(true);
scriptableCluster.setEditPriv("MaintainItemGroups");
scriptableCluster.setNewPriv("MaintainItemGroups");
```
Lastly, a constructor was exposed to the scripting engine allowing you to instantiate a new cluster in the normal way:
```javascript
var scriptableCluster = toolbox.createWidget("ScriptableCluster", mywindow, "scriptableCluster");
```


davidbeauchamp:
Any thoughts on renaming this class to XCluster to be more in line with XComboBox and such? "ScriptableCluster" seems like such a mouthful.

andersdd:
"XCluster" makes sense to me.

gilmoskowitz:
We're going to try making the VirtualCluster itself scriptable and we don't want this in 4.10.x anyway. Closing this PR with the expectation that we'll replace it later.

PR #1340, opened 2016-08-08
issue #28331 - check whether custtype changed before asking about charass

gilmoskowitz:
also fix 1 compiler warning and 1 line length violation


ChristopherCotnoir:
+1

PR #1343, opened 2016-08-12
issue #28306 - use unified getEdition function

gilmoskowitz:
and minor code simplifications
requires xtuple/xtuple#2792

easiest to review [ignoring whitespace changes](https://github.com/xtuple/qt-client/pull/1343/files?w=1)


gilmoskowitz:
@rkannadevara - You might need to update the process by which the Squish tests check the application edition. These changes might also force changing the object map to find the main application window.

PR #1352, opened 2016-08-30
issue #28421 Make next check # restriction consistent and limit to 9 digits

ChristopherCotnoir:


andersdd:
@ChristopherCotnoir  Why did you restrict # to 9 digits?  Bear in mind this is a "payments" number, not just checks and some of the electronic payment reference numbers are very long (possibly > 9 decimals)

ChristopherCotnoir:
Most 10 digit numbers and all numbers with more than 10 digits cause integer overflow. If we really need higher numbers, we can use a long, I suppose. As it currently stands, anything that won't fit in a 32-bit signed integer breaks it anyway, and the next check number in setup was already limited to 7 digits. It seems unlikely to me that anyone was trying higher than that, or they would have come across this.

gilmoskowitz:
Feature request 28467 covers the desire/need for longer payment numbers.

PR #1357, opened 2016-09-01
issue #19108 Load initMenu scripts in the order their schema appears in the search…

ChristopherCotnoir:
… path when the script order is the same

### DIFF
--- a/guiclient/guiclient.cpp
+++ b/guiclient/guiclient.cpp
@@ -894,11 +894,17 @@ void GUIClient::showEvent(QShowEvent *event)
     // We only want the scripting to work on the NEO menu
     // START script code
       XSqlQuery sq;
-      sq.prepare("SELECT script_source, script_order"
-              "  FROM script"
-              " WHERE((script_name=:script_name)"
-              "   AND (script_enabled))"
-              " ORDER BY script_order;");
+      sq.prepare("SELECT script_source "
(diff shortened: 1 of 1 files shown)


guiclient/guiclient.cpp
```diff
@@ -894,11 +894,17 @@ void GUIClient::showEvent(QShowEvent *event)
...
+                 "FROM (SELECT sequence_value, "
+                 "(SELECT regexp_split_to_array(buildsearchpath(), ','))[sequence_value] AS pkg "
+                 "FROM sequence) ord "
+                 "JOIN script "
+                 "ON ((pkg || '.pkgscript')=tableoid::regclass::text "
+                 "OR (pkg='public' AND tableoid::regclass::text='script')) "
+                 "WHERE((script_name=:script_name)"
+                 "AND (script_enabled))"
+                 "ORDER BY script_order, sequence_value;");
+
```
gilmoskowitz: The following evaluates about 5 times faster:
```sql
select script_source
  from script
  join (select unnest as pkgname, row_number() over () as seq
          from unnest(string_to_array(buildsearchpath(), ','))) as path
        on pkgname || '.pkgscript' = tableoid::regclass::text
           or (pkgname = 'public' and tableoid::regclass::text = 'script')
 where script_enabled
   and script_name = 'initMenu'
order by script_order, seq;
```

PR #1358, opened 2016-09-02
Fix regression from fix for issue #20393

ChristopherCotnoir:


guiclient/scripts.cpp
```diff
@@ -59,7 +59,7 @@ void scripts::sNew()
   scriptEditor *newdlg = new scriptEditor();
   newdlg->set(params);
 
-  omfgThis->handleNewWindow(newdlg);
+  omfgThis->handleNewWindow(newdlg, Qt::ApplicationModal);
```
gilmoskowitz: Does this undo the recent change to allow multiple simultaneous script editors?

ChristopherCotnoir: It does. If we wish to allow that, then we really shouldn't make the scriptEditor window modal at all. As it was before I made this change, it actually wasn't even truly modal. It kept focus, but you could still input to any other windows if you moved it out of the way, which defeats the whole purpose of modality. Bear in mind, WindowModal doesn't mean anything unless the window has a parent, which it does not. And if it did, then you actually wouldn't be able to open multiple script editors.

PR #1361, opened 2016-09-06
issue #28280 - Removed transaction block from voucherItem and fixed split receipt freight

ChristopherCotnoir:
removed transaction and fixed split freight functionality

### DIFF
--- a/guiclient/voucherItem.cpp
+++ b/guiclient/voucherItem.cpp
@@ -361,26 +361,17 @@ void voucherItem::sSave()
                "    voitem_qty=:voitem_qty, "
                "    voitem_freight=:voitem_freight, "
 	  		 "    voitem_taxtype_id=:voitem_taxtype_id "
-               "WHERE (voitem_id=:voitem_id);" );
+               "WHERE (voitem_id=:voitem_id) "
+               "RETURNING voitem_id;" );
+    voucherSave.bindValue(":voitem_id", _voitemid);
   }
   else
   {
-  // Get next voitem id
-    voucherSave.prepare("SELECT NEXTVAL('voitem_voitem_id_seq') AS voitemid");
-    voucherSave.exec();
-    if (voucherSave.first())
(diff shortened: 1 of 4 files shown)


ChristopherCotnoir:
Requires xtuple/xtuple#2812.

guiclient/voucherItem.cpp
```diff
@@ -325,17 +338,53 @@ void voucherItem::sSave()
...
+    voucherSave.prepare("SELECT NEXTVAL('voitem_voitem_id_seq') AS voitemid");
+    voucherSave.exec();
+    if (voucherSave.first())
+      _voitemid = (voucherSave.value("voitemid").toInt());
+    else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Voucher Item Information "),
+                                  voucherSave, __FILE__, __LINE__))
+    {
+      reject();
+      return;
+    }
```
gilmoskowitz: The `select nextval()` is unnecessary. Instead, put the `bindValue(":voitem_id", _voitemid)` in the `if` block, remove `voitem_id` from the insert column list, and make both update and insert `returning voitem_id`. Then `exec()` and `first()` the query to set  `_voitemid`. The result is simpler, more robust code.

guiclient/voucherItem.cpp
```diff
@@ -475,55 +542,13 @@ void voucherItem::sToggleReceiving(QTreeWidgetItem *pItem)
...
-  voucherToggleReceiving.exec();
-  if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Voucher Item Information"),
-                                voucherToggleReceiving, __FILE__, __LINE__))
-  {
-    reject();
-    return;
-  }
-  
   // Update the receipt record
   if (item->text(4) == "Yes")
```
gilmoskowitz: Please change this to use `f_tagged` instead of `4`.

PR #1364, opened 2016-09-07
issue #20404 Disconnect unnecessary signal/slot connections at the end of sSave be…

ChristopherCotnoir:
…fore closing window


ChristopherCotnoir:
Related to xtuple/xtuple#2813.

PR #1365, opened 2016-09-07
Issue #28494: Add rental line item comment source

davidbeauchamp:
Also fixes incorrect TATC entry in comment.cpp.

Requires PR from /xtuple/xtuple to populate the source table.


davidbeauchamp:
Requires https://github.com/xtuple/xtuple/pull/2814

widgets/comment.cpp
```diff
@@ -231,7 +231,7 @@ void comment::set(const ParameterList &pParams)
         _cmnttype->setType(XComboBox::SalesOrderItemCommentTypes);
       else if (_sourcetype == "TA")
         _cmnttype->setType(XComboBox::TaskCommentTypes);
-      else if (_sourcetype == "TIME")
+      else if (_sourcetype == "TATC")
```
gilmoskowitz: Do we need to update any existing comment data?

davidbeauchamp: I don't believe so, the records were inserted using the proper TATC type, this fix makes it so the comments screen will filter by the TATC comment types properly, instead of falling back to the else at the end of the conditional and showing all comment types.

PR #1375, opened 2016-09-12
issue #28517 - no need to convert back & forth

gilmoskowitz:
The real fix is a one-liner - change the return type of `ReceiverItem::slot()` to QString. The rest of the changes helped debug the problem.


bendiy:
+1

PR #1374, opened 2016-09-12
issue #16967 Added control for new recurrence function with temporary names. DO NOT MERGE

ChristopherCotnoir:

PR #1387, opened 2016-09-23
issue #16967 - Added control for new recurrence function with temporary names.

gilmoskowitz:
replaces xtuple/qt-client#1374
requires xtuple/xtuple#2831


widgets/recurrencewidget.cpp
```diff
@@ -419,6 +419,16 @@ bool RecurrenceWidget::maxVisible() const
...
+QString RecurrenceWidget::style() const
+{
+  if (_none->isChecked())
+    return "KeepNone";
+  if (_one->isChecked())
+    return "KeepOne";
+  if (_all->isChecked())
+    return "KeepAll";
+}
+
```
gilmoskowitz: Consider using an XComboBox with three calls to `append()` instead of radio buttons. What are the relative merits and demerits?

gilmoskowitz:
Please expand the API documentation to describe the meanings of the new styles.

widgets/recurrencewidget.cpp
ChristopherCotnoir: Offline discussion:

Decided radio buttons were better because these options are best to be able to see at a glance, and the aesthetic for the widget works better with this option as radio buttons.

PR #1389, opened 2016-09-26
issue #28589 Prevent issuing stock on Sales Order if credit limit exceeded

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/private-extensions#837.

guiclient/salesOrder.cpp
```diff
@@ -5569,6 +5578,39 @@ bool salesOrder::creditLimitCheck()
...
+                                       "it falls within the credit limit or obtain a payment first."));
+        return false;
+      }
+    }
+  }
+
+  save(true);
+
+  return true;
+}
```
gilmoskowitz: This is a duplication of the logic in `save()` above except for the use of `critical()` vs. `question()`. Make this a `const` method that doesn't change `_holdType` and doesn't call `save()`. It could accept a parameter whether to ask or tell the user what's going on. Then the caller could decide what to do with the result.

gilmoskowitz: Off-line discussion - the logic isn't quite close enough to share.

PR #1391, opened 2016-09-29
issue #28585 Added default customer credit status

ChristopherCotnoir:


guiclient/configureSO.cpp
```diff
@@ -118,6 +118,16 @@ configureSO::configureSO(QWidget* parent, const char* name, bool /*modal*/, Qt::
...
+  metric = _metrics->value("SoCreditStatus");
+    if (metric == "G")
+      _creditStatus->setCurrentIndex(0);
+    else if (metric == "W")
+      _creditStatus->setCurrentIndex(1);
+    else if (metric == "H")
+      _creditStatus->setCurrentIndex(2);
+    else
+      _creditStatus->setCurrentIndex(3);
+
```
gilmoskowitz: This will work and mirrors existing style but it's difficult to maintain. Our newer approach is to:
- populate the combobox using [XComboBox::append()](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/xcombobox.h#L138) instead of creating the `creditStatuses array` in `sSave()` and text to the .ui file
- call [XComboBox::setCode()](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/xcombobox.h#L104) with the metric value
- call [XComboBox::code()](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/xcombobox.h#L122) in `sSave()` to get the code instead of reading the array

The resulting code is smaller, more extensible, and easier to maintain because there's less translation back and forth between integers, code values, and human-readable strings. In addition there's no need to update both .ui and .cpp when adding or subtracting code values.

guiclient/configureSO.ui
```diff
@@ -1205,6 +1205,43 @@ to be bound by its terms.</comment>
...
+                     </property>
+                    </item>
+                    <item>
+                     <property name="text">
+                      <string>On Credit Hold</string>
+                    </property>
+                    </item>
+                    <item>
+                     <property name="text">
+                      <string>[ no default ]</string>
```
gilmoskowitz: The "no default" case is better handled via [XComboBox::allowNull](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/xcombobox.h#L37) and [nullStr](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/xcombobox.h#L38). This means "allow a null case in the combo box," not "allow a null case in the database." The latter should be managed in the save method of the UI and a constraint in the database.

guiclient/configureSO.cpp
```diff
@@ -34,6 +34,10 @@ configureSO::configureSO(QWidget* parent, const char* name, bool /*modal*/, Qt::
   _nextInNumber->setValidator(omfgThis->orderVal());
   _creditLimit->setValidator(omfgThis->moneyVal());
 
+  _creditStatus->append(0, "In Good Standing", "G");
+  _creditStatus->append(1, "On Credit Warning", "W");
+  _creditStatus->append(2, "On Credit Hold", "H");
```
gilmoskowitz: `tr("the text")` but not the code value

PR #1401, opened 2016-10-06
Issue #28659: Allow saving sales order with no sales lines if the rentals package is enabled

davidbeauchamp:


guiclient/salesOrder.cpp
```diff
@@ -806,6 +804,10 @@ bool salesOrder::save(bool partial)
          << GuiErrorCheck(!partial && !_project->isValid() && _metrics->boolean("RequireProjectAssignment"), _project,
                           tr("<p>You must enter a Project for this order before you may save it."))
   ;
+  if (!(_metrics->boolean("EnableRentals") && ISORDER(_mode))) {
```
davidbeauchamp: this looks funky at first but since rentals only runs when the screen is in Order mode i only want to exclude the check if both rentals is enabled and it is a sales order.

gilmoskowitz: Hulk sad


but we don't have a better solution for 4.10.0.

PR #1409, opened 2016-10-11
issue #9200 Add option to add custom prefix for ACH Company Number

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#2853.

guiclient/bankAccount.cpp
```diff
@@ -47,19 +49,24 @@ bankAccount::bankAccount(QWidget* parent, const char* name, bool modal, Qt::Wind
                    "WHERE form_key='Chck' "
                    "ORDER BY form_name;" );
 
+  if (_metrics->value("ACHCompanyIdPrefix")!="")
```
gilmoskowitz: This uses a metric, which by definition is a global, to handle the needs of a particular bank account. This doesn't seem correct.

ChristopherCotnoir: You're right, this was a mistake. However, looking closely at this, it seems this really shouldn't be on the bank account page in the first place like we originally discussed, but instead in Accounts Payable setup. I can't remember how we originally came to the conclusion to put it with the bank accounts, but the ACH Company Number is universal, the places that use it don't pass a bank account, and the company id is defined in setup.

PR #1416, opened 2016-10-14
issue #28706 Don't attempt background save if memo number not set

ChristopherCotnoir:


guiclient/creditMemo.cpp
```diff
@@ -321,6 +321,9 @@ void creditMemo::sSave()
 
 bool creditMemo::save()
 {
+  if(_memoNumber->text().length() == 0)
```
gilmoskowitz: Add `trimmed()` to the chain. Otherwise we'll wind up with credit memos with spaces in the number.

PR #1417, opened 2016-10-14
issue #28718 Made it possible to delete credit cards

ChristopherCotnoir:


guiclient/customer.cpp
```diff
@@ -410,6 +413,7 @@ void customer::setViewMode()
   _save->hide();
   _newCC->setEnabled(false);
   _editCC->setEnabled(false);
+  _deleteCC->setEnabled(false);
```
gilmoskowitz: We should also disable the delete button if the credit card has already been used, or change it to "Inactivate". Otherwise the deletion will fail and the user may not know why.

guiclient/customer.cpp
```diff
@@ -1533,6 +1537,22 @@ void customer::sViewCreditCard()
...
+                            tr("Are you sure that you want to delete this Credit Card?"),
+                              QMessageBox::Yes,
+                              QMessageBox::No | QMessageBox::Default) == QMessageBox::No)
+    return;
+
+  XSqlQuery m;
+  m.prepare("DELETE FROM ccard WHERE ccard_id=:ccard_id;");
+  m.bindValue(":ccard_id", _cc->id());
+  m.exec();
+
```
gilmoskowitz: Every database call should at least have an error check (or a comment describing why not). Eventually we should have error *handling*.

guiclient/customer.cpp
gilmoskowitz: We can't allow deletion of cards that have been used, at least not at this time. Doing so would orphan various records and probably cause certain queries to return incorrect results. If deleting a used `ccard` record does not fail then we need to add some foreign key constraints to enforce referential integrity.

ChristopherCotnoir: I have confirmed that deleting a used card does not cause a database failure. This is likely because, at present, there is no way to ever remove an entry from the ccard table through any xTuple application. I can keep this addition and add a trigger to prevent deletion of used cards, however this issue was brought up by a customer looking to delete a card that has been used, so such a change would not be a 'fix' for this incident. Should I mark this incident as unresolvable? And should we still make this change?

As for foreign key constraints, I believe ccpay is the only table that references the ccard table. I can add a foreign key constraint there, at least.

gilmoskowitz: - Please add the foreign key constraint so a ccard cannot be deleted if a corresponding ccpay record exists.
- If someone tries to delete a ccard that has been used, please mark it inactive and change the expiration month and year to either last month or, if it's been used recently, the current month. That's the best we can do right now.

PR #1426, opened 2016-10-20
Issue #28763: Expose guiErrorCheck to script

davidbeauchamp:
Sample Script:

    var list = {};
    list[0] = GuiErrorCheck(1 == 2, mywindow, "This is an error");
    list[1] = GuiErrorCheck(1 == 1, mywindow, "This is not an error");
    var hasErrors = GuiErrorCheck.reportErrors(mywindow, "Error Check", list);

guiscreenshot

### DIFF
--- a/guiclient/guiErrorCheck.cpp
+++ b/guiclient/guiErrorCheck.cpp
@@ -100,15 +100,12 @@ static QScriptValue guierrorcheck_reporterrors(QScriptContext *context, QScriptE
   return QScriptValue(result);
 }
 
-void setupGuiErrorCheckProto(QScriptEngine *engine)
+void setupGuiErrorCheck(QScriptEngine *engine)
 {
   qScriptRegisterMetaType(engine, QListGuiErrorChecktoScriptValue, QListGuiErrorCheckfromScriptValue);
 
-  QScriptValue proto = engine->newQObject(new GuiErrorCheckProto(engine));
-  engine->setDefaultPrototype(qMetaTypeId<GuiErrorCheck>(), proto);
-
-  QScriptValue constructor = engine->newFunction(constructGuiErrorCheck,  proto);
-  engine->globalObject().setProperty("GuiErrorCheck",  constructor, CTORPROPFLAGS);
+  QScriptValue constructor = engine->newFunction(constructGuiErrorCheck);
+  engine->globalObject().setProperty("GuiErrorCheck", constructor, CTORPROPFLAGS);
 
   constructor.setProperty("reportErrors", engine->newFunction(guierrorcheck_reporterrors), STATICPROPFLAGS);
 }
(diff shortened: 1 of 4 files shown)


scriptapi/setupscriptapi.cpp
```diff
@@ -23,6 +23,7 @@
 #include "exporthelper.h"
 #include "filemoveselector.h"
 #include "glclustersetup.h"
+#include "../guiclient/guiErrorCheck.h"
```
gilmoskowitz: Forward references in filesystem structure are evil. Whatever setup is required should be done [in the guiclient](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/guiclient.cpp#L2289).

guiclient/guiErrorCheck.h
```diff
@@ -31,4 +34,29 @@ class GuiErrorCheck
...
 };
 
+// Script exposure
+
+Q_DECLARE_METATYPE(GuiErrorCheck)
+
+void setupGuiErrorCheckProto(QScriptEngine *engine);
+QScriptValue constructGuiErrorCheck(QScriptContext *context, QScriptEngine *engine);
+
+class GuiErrorCheckProto : public QObject, public QScriptable
```
gilmoskowitz: This class should not be necessary. We try to limit `Proto` wrapper classes to code we don't control, such as Qt classes.

guiclient/guiErrorCheck.cpp
```diff
@@ -47,3 +55,113 @@ bool GuiErrorCheck::reportErrors(QWidget *parent, QString title,
...
+    return item.widget;
+  return 0;
+}
+
+QString GuiErrorCheckProto::msg() const
+{
+  GuiErrorCheck item = qscriptvalue_cast<GuiErrorCheck>(thisObject());
+  return item.msg;
+}
+
```
davidbeauchamp: They aren't methods on GuiErrorCheck, only properties. Seems I should move these methods to GuiErroCheck itself, then?

scriptapi/setupscriptapi.cpp
davidbeauchamp: Understood.

guiclient/guiErrorCheck.h
davidbeauchamp: I made the proto class here to give access to the same properties the C++ object has. It doesn't really matter, and if I move the methods to the GuiErrorCheck class then I don't think we need this either.

guiclient/guiErrorCheck.cpp
gilmoskowitz: Move these methods and `Q_PROPERTY` lines to `GuiErrorCheck` itself. Do **not** make the methods `Q_INVOKABLE`.

davidbeauchamp:
Using a javascript array didn't make the to/from functions any easier, and instead of polluting the class with properties and methods that didn't have an immediate use I decided to just remove them. Sample code still valid, only difference is without the proto class the debugger shows QVariant(GuiErrorCheck) instead of GuiErrorCheck(error/noError).

gilmoskowitz:
Is there a simple `#ifdef` that allows building with Qt4? If so then we'll be able to merge while working on the CI handoff.

davidbeauchamp:
Looks like that was the better choice anyways. Those error messages sure weren't the most direct...

PR #1430, opened 2016-10-25
issue #27345 Clear item fields when item is cleared

ChristopherCotnoir:


gilmoskowitz:
This doesn't appear to fix the main problem, which is that the item widget is enabled prematurely.

PR #1433, opened 2016-10-28
issue #28746 Properly handle batches where some items cannot be issued due to inve…

ChristopherCotnoir:
…ntory shortage

### DIFF
--- a/guiclient/issueWoMaterialBatch.cpp
+++ b/guiclient/issueWoMaterialBatch.cpp
@@ -131,7 +131,7 @@ void issueWoMaterialBatch::sIssue()
     }
   }
 
-  sqlissue = ("SELECT itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(CASE WHEN (womatl_qtyreq >= 0) THEN womatl_qtyreq - womatl_qtyiss ELSE womatl_qtyiss * -1 END)))) AS isqtyavail "
+  sqlissue = ("SELECT bool_and(itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(CASE WHEN (womatl_qtyreq >= 0) THEN womatl_qtyreq - womatl_qtyiss ELSE womatl_qtyiss * -1 END))))) AS isqtyavail "
               "FROM womatl "
               "JOIN itemsite ON (womatl_itemsite_id = itemsite_id) "
               "JOIN item ON (itemsite_item_id = item_id) "
@@ -143,15 +143,10 @@ void issueWoMaterialBatch::sIssue()
               " AND (womatl_wo_id=<? value('wo_id') ?>);");
   mqlissue.setQuery(sqlissue);
   issue = mqlissue.toQuery(params);
-  while(issue.next())
-  {
-    if(!(issue.value("isqtyavail").toBool()))
-      if(QMessageBox::question(this, tr("Continue?"), tr("One or more items cannot be issued due to insufficient inventory. Issue all other items?"),
+  if (issue.first() && ! issue.value("isqtyavail").toBool() &&
+      QMessageBox::question(this, tr("Continue?"), tr("One or more items cannot be issued due to insufficient inventory. Issue all other items?"),
                                QMessageBox::No | QMessageBox::Default, QMessageBox::Yes) == QMessageBox::No)
         return;
-      else
-        break;
-  }
 
   XSqlQuery rollback;
   rollback.prepare("ROLLBACK;");
--- a/guiclient/issueWoMaterialBatch.cpp
+++ b/guiclient/issueWoMaterialBatch.cpp
@@ -131,6 +131,36 @@ void issueWoMaterialBatch::sIssue()
     }
   }
 
+  sqlissue = ("SELECT CASE WHEN (womatl_qtyreq >= 0) THEN "
+              "  itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(womatl_qtyreq - womatl_qtyiss)))) "
+              "ELSE "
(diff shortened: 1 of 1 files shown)


guiclient/issueWoMaterialBatch.cpp
```diff
@@ -131,6 +131,36 @@ void issueWoMaterialBatch::sIssue()
...
+              " AND (womatl_issuemethod IN ('S', 'M')) "
+              " <? if exists(\"pickItemsOnly\") ?> "
+              " AND (womatl_picklist) "
+              " <? endif ?> "
+              " AND (womatl_wo_id=<? value(\"wo_id\") ?>);");
+  mqlissue.setQuery(sqlissue);
+  params = ParameterList();
+  params.append("wo_id", _wo->id());
+  if (!_nonPickItems->isChecked())
+    params.append("pickItemsOnly", true);
```
gilmoskowitz: This is identical to the parameter list created a few lines earlier so just reuse that one.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -131,6 +131,36 @@ void issueWoMaterialBatch::sIssue()
...
+              "  itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(womatl_qtyreq - womatl_qtyiss)))) "
+              "ELSE "
+              "  itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(womatl_qtyiss * -1)))) "
+              "END AS isqtyavail "
+              "FROM womatl "
+              "JOIN itemsite ON (womatl_itemsite_id = itemsite_id) "
+              "JOIN item ON (itemsite_item_id = item_id) "
+              "WHERE (fetchMetricBool('DisallowNegativeInventory') OR itemsite_costmethod='A') "
+              " AND (womatl_issuemethod IN ('S', 'M')) "
+              " <? if exists(\"pickItemsOnly\") ?> "
```
gilmoskowitz: Use `'` instead of `\"`

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -131,6 +131,36 @@ void issueWoMaterialBatch::sIssue()
     }
   }
 
+  sqlissue = ("SELECT CASE WHEN (womatl_qtyreq >= 0) THEN "
+              "  itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(womatl_qtyreq - womatl_qtyiss)))) "
+              "ELSE "
+              "  itemsite_qtyonhand < roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(womatl_qtyiss * -1)))) "
```
gilmoskowitz: The two cases are very similar. Consider moving the `CASE` construct to the real point of difference.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -131,6 +131,36 @@ void issueWoMaterialBatch::sIssue()
...
+  issue = mqlissue.toQuery(params);
+  while(issue.next())
+  {
+    if(!(issue.value("isqtyavail").toBool()))
+      if(QMessageBox::question(this, tr("Continue?"), tr("One or more items cannot be issued due to insufficient inventory. Issue all other items?"),
+                               QMessageBox::No | QMessageBox::Default, QMessageBox::Yes) == QMessageBox::No)
+        return;
+      else
+        break;
+  }
```
gilmoskowitz: [`bool_and`](https://www.postgresql.org/docs/9.3/static/functions-aggregate.html) would replace the loop with a single `if`:
```c++
if (issue.first() &&
    ! issue.value("isqtyavail").toBool() &&
    QMessageBox::question(...) == QMessageBox::No)
  return;
```

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -198,22 +228,14 @@ void issueWoMaterialBatch::sIssue()
...
+
+      issue.exec("COMMIT;");      
     }
     else
-    {
       rollback.exec();
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Occurred"),
-                           tr("Window:%1/n Error Issuing Material: Work Order ID #%2")
-                           .arg(windowTitle())
-                           .arg(_wo->id()),__FILE__,__LINE__);
```
gilmoskowitz: We should still say what went wrong if we got here.

guiclient/issueWoMaterialBatch.cpp
ChristopherCotnoir: The query returns whether or not issuing will cause the item to go negative for all work order material items with average costing, or all work order material items if negative inventory is disallowed. There can be more than one of those, a loop is necessary.

guiclient/issueWoMaterialBatch.cpp
ChristopherCotnoir: The problem is that if we do this, and many items go negative or have some unrelated error, the user will receive an error for each one since this is inside a loop and the loop continues even if failures occur. Is it OK to do that?

gilmoskowitz: Two choices: Either report each error as it occurs or report them as a set when the loop terminates. See the use of [`notConverted`](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/quotes.cpp#L185) in [quotes::sConvert()](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/quotes.cpp#L156-L342) for a rough example of error accumulation.

ChristopherCotnoir: If some items cannot be issued due to negative inventory, the user is prompted whether they want to continue anyway. If they select yes, this is dealt with by simply rolling back issuing any such items as they fail, while letting any others that do not fail go. The problem this presents for error reporting is that it would likely not be a good idea to report the failures caused by negative inventory, because the user has already acknowledged them and chosen to ignore them. But if we choose to report any other errors that may occur, I'm not sure how we should go about skipping the negative inventory errors. The only way to tell them apart, so far as I can tell, is from the error text, but checking for that is problematic because the error text may be changed in the future, and it would be easy to forget about this case. How should this be handled?

gilmoskowitz: Accumulate and report them all for now. This gives the user a list s/he to review and decide next steps. The user accepted that some inventory would go negative but might want to know which inventory should be adjusted.
We can deal with refining the list in a future iteration.

PR #1434, opened 2016-10-31
issue #28800 Ensure the most recent id created is fetched

ChristopherCotnoir:


guiclient/createLotSerial.cpp
```diff
@@ -457,7 +457,7 @@ void createLotSerial::sAssign()
        * next ls_id.  Fetch the last ls_id */
       int ls_id = -1;
       XSqlQuery ls_id_query;
-      ls_id_query.prepare("SELECT ls_id FROM ls WHERE ls_number=:lotserial");
+      ls_id_query.prepare("SELECT ls_id FROM ls WHERE ls_number=:lotserial ORDER BY ls_id DESC");
```
gilmoskowitz: The bug here is that the old (and the new code) guesses at the `ls_id`. There's enough information available to find _the_ `ls_id`. Note that `createlotserial()` generates the `ls_id` if necessary and feeds that to the insert on `itemlocdist`. Note also that the return value of `createlotserial()` is the `itemlocdist_id`.

Save the result of `createlotserial()` when called at line 424 or 430, then replace this `select ls_id from ls...` with `select itemlocdist_ls_id from itemlocdist ...`

PR #1444, opened 2016-11-07
Issue #27308: Sales Order query optimisation

andersdd:
Reduces Sales Order Item List query calls from 4 to 1 (in my system) by removing unnecessary signals and also preventing the sSalesOrderUpdated signal from calling itself


gilmoskowitz:
That `_shipVia` connection is old and I don't know why we had it. Nice sleuthing, @andersdd.

PR #1446, opened 2016-11-09
issue #28877 Make sure that id is greater than 0 to prevent future problems

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#2889.

PR #1447, opened 2016-11-10
Issue #28889: Make search on display class windows respond to both Return AND Enter keys

davidbeauchamp:
While researching the best way to do this I discovered setShortcuts() is available, allowing two shortcuts set from a QList (instead of my alternative idea to instantiate the shortcuts and use their signals).


davidbeauchamp:
If possible, I would request backporting this to 4.10.0...

PR #1452, opened 2016-11-14
Issue #28298: Add alert if selected ACH batch number is less than

andersdd:
a number already used

### DIFF
--- a/guiclient/configureGL.cpp
+++ b/guiclient/configureGL.cpp
@@ -637,6 +637,27 @@ bool configureGL::sSave()
     }
   }
 
+  if (_achGroup->isChecked())
+  {
+    // Check ACH Batch Number is greater than existing value
+    XSqlQuery check;
+    check.exec("SELECT max(checkhead_ach_batch)::INTEGER AS maxbatch "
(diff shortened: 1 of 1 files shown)


guiclient/configureGL.cpp
```diff
@@ -637,6 +637,28 @@ bool configureGL::sSave()
...
+    {
+      int ach = _nextACHBatchNumber->text().toInt();
+      if (check.value("maxbatch").toInt() >= ach)
+      {
+        if (QMessageBox::question(this, tr("Confirm ACH Batch"),
+                                tr("The ACH batch number is less than a batch number already used. "
+                                   "This may cause duplicated batch numbers in the future. "
+                                   "Would you like to proceed?"),
+                                    QMessageBox::Yes | QMessageBox::Default,
+                                    QMessageBox::No ) == QMessageBox::No)
```
gilmoskowitz: This is a [deprecated](http://doc.qt.io/qt-5/qmessagebox-obsolete.html#question-1) usage of [QMessageBox::question()](http://doc.qt.io/qt-5/qmessagebox.html#question)

PR #1460, opened 2016-11-17
issue #28919 - Init the remembered state in XCheckBox constructor.

bendiy:


guiclient/packingListBatch.ui
```diff
@@ -114,7 +114,7 @@ to be bound by its terms.</comment>
                <property name="text">
                 <string>Credit Hold</string>
                </property>
-               <property name="checked">
+               <property name="defaultChecked">
```
bendiy: With this change, if the property `QCheckBox.checked` is used in a .ui file, when the window initializes, the `XCheckBox` non-forgetful remembered checked state will be overridden and whatever the .ui file is set to (`true` or `false`) will be shown.

The `XCheckBox` has a `defaultChecked` property that should _always_ be used instead if a default needs to be specified in the .ui file. Using `XCheckBox.defaultChecked` in the .ui file with this change works fine. Using `XCheckBox.checked` does not.

I have not looked into the scope of .ui files with a `XCheckBox.checked` setting, so I don't know how big this change's impact is. I'm not sure if there is a way to fix this limitation.

widgets/xcheckbox.cpp
```diff
@@ -44,13 +46,28 @@ void XCheckBox::setData()
...
+  while (child->parentWidget()) {
+    if (child->parentWidget()->objectName() != "") {
+      child = child->parentWidget();
+    } else {
+      break;
+    }
+  }
+
+  return child;
+}
```
gilmoskowitz: Why is `window()` insufficient?
If the problem is that an XComboBox on an XMainWindow/XWidget embedded in another window must be distinguished from when that XMainWindow/XWidget is opened on its own, build a longer path using the `parentWidget()` chain.

Otherwise just let `window()` to do its job.

bendiy: Needing a longer path to distinguish the widgets isn't the problem here.

Using `window()` is insufficient because when calling `init()` through the constructor the `window()` object doesn't have its name yet. Calling `window()->objectName()` returns an empty string at this point. Even `this->objectName()` returns an empty string. That's why I had to connect the `objectNameChanged` signal. I tried doing the same with `window` `objectNameChanged`, but it fired too late (after the `showEvent` we originally used) so the double query issue was still present.

By walking the `parentWidget()` chain here, I was able to get the same `window()->objectName()` string that the original code used on the screens I debugged. I can only assume the root widget has it's name, but the `window()` object isn't completely setup yet.

I thought about building up a full path string like `packingListBatch/_holdGroup/_creditHold`, but I didn't want to break existing remembered checkbox state. The old path was `packingListBatch/_creditHold` and this change is creating the same path `packingListBatch/_creditHold`. I think a full path is a better way to do it, but users will have to reset any remembered checkboxes in the application. The other issue with the long path is if we ever move a `XCheckBox` around in the UI, it's path will most likely change breaking any remembered state. If you think we should use the full path, I can make that change.

guiclient/packingListBatch.ui
ChristopherCotnoir: I looked in the following repositories:

connect
custom
fixed-assets
fixed-assets-commercial
fx-integration-extension
migration-testing
private-extensions
qt-client
qt-sample-scripts
updater
xtcommission
xtdesktop
xtprjaccnt
xtte

I found these files with XCheckBoxes using the checked property:

custom/xProvare/xpmerge/dbscripts/uiforms/contactMerge.ui
custom/xProvare/xpmerge/updatescripts/from_000/uiforms/contactMerge.ui
custom/xTuple/fixrecur/resources/client/uiforms/fixinvoices.ui
custom/xTuple/xtadvincdtwb/dbscripts/uiforms/advIncidentWorkbench.ui
custom/xTuple/xtadvincdtwb/updatescripts/from_000/uiforms/advIncidentWorkbench.ui
private-extensions/source/manufacturing/foundation-database/xtmfg/tables/pkguiform/dspMRPException.ui
qt-client/guiclient/contactMerge.ui
qt-client/guiclient/dspAROpenItems.ui
qt-client/guiclient/dspTaxHistory.ui
qt-client/guiclient/employee.ui
qt-client/guiclient/metasqls.ui
qt-client/guiclient/packingListBatch.ui
qt-client/guiclient/purchaseOrderItem.ui
qt-client/guiclient/salesOrderItem.ui
qt-client/guiclient/scrapWoMaterialFromWIP.ui
qt-client/guiclient/scripts.ui
qt-client/guiclient/todoList.ui
qt-client/guiclient/uiforms.ui
qt-client/guiclient/voucherItemDistrib.ui
qt-client/guiclient/voucherMiscDistrib.ui
qt-client/widgets/alarmMaint.ui

There may be more, but that's what I could find.

bendiy: @ChristopherCotnoir Thanks!
It shouldn't be too difficult to convert all of those. I'm still not 100% sure this is the best approach to fix the underlying issue here, but it is one approach that does work. I'd like to be able to either hide `QCheckBox.checked` from Qt Designer or make it work like `XCheckBox.defaultChecked` does.

ChristopherCotnoir: I don't particularly like having the checked property be broken myself, if you can think of a better way to fix the problem, that would be great. If not, we'll have to settle. I don't know enough about how the qt ui files are processed, would overriding the setChecked(bool) function from QCheckBox to call init() there and call QCheckBox::setChecked(bool) if the XCheckBox is forgetful or no saved preference exists work?

gilmoskowitz: `checked` isn't broken. It's doing exactly what it's supposed to do - it's a bi-directional property of the object that controls and reports whether the checkbox is checked "now".

The problem is that we've confused ourselves by adding a second property with a similar but not identical purpose - whether the checkbox should be checked or not "the first time". Then we used the wrong property to set the "first time" state. Either that or we didn't review all XCheckBox uses when we first invented it.

gilmoskowitz: In retrospect the `XCheckBox` was a bad idea. We probably should have created a `XCheckBoxWithMemory` instead. Then the semantics would be clearer because we wouldn't have to support both forgetful and non-forgetful.

Short term - review all XCheckBox references to `checked`.

Long term - revisit the widget's relevance

PR #1465, opened 2016-11-20
issue #28932 Insert bankrec record on load instead of on save

ChristopherCotnoir:


ChristopherCotnoir:
This window operates differently than most. If you save, and then reopen the window, then cancel, it will delete everything (after prompting), rather than the changes you just made. Which means the code to clean up is actually already in place.

PR #1472, opened 2016-11-22
issue #28934 Only enable select button when valid invoice is selected and fix clos…

ChristopherCotnoir:
…e button functionality


guiclient/invoiceList.cpp
```diff
@@ -23,6 +23,7 @@ invoiceList::invoiceList(QWidget* parent, const char* name, bool modal, Qt::Wind
   connect(_invoice, SIGNAL(itemSelected(int)), this, SLOT(sSelect()));
   connect(_cust, SIGNAL(newId(int)), this, SLOT(sFillList()));
   connect(_dates, SIGNAL(updated()), this, SLOT(sFillList()));
+  connect(_invoice, SIGNAL(valid(bool)), _select, SLOT(setEnabled(bool)));
```
andersdd: Can we confirm with @gilmoskowitz whether these signals should be done in the cpp or as <connections> in the ui file?

I'm sure Gil indicated the preferred approach was a ui connection.  If so we also need to update the Coding Standards.

gilmoskowitz: When possible connections should be established in the .ui. I'll update the coding standards doc to be explicit.

PR #1476, opened 2016-11-29
Issue #27705: Resolve Item characteristic propagation to order documents

andersdd:
Only propagate to order item chars if selected in characteristic setup.

This change imports Item characteristics set to default into the order item, plus also displays existing assignments to the order item even if the setting was subsequently turned off.

consider in conjunction with xtuple/xtuple#2910


guiclient/purchaseOrderItem.cpp
```diff
@@ -527,16 +527,26 @@ void purchaseOrderItem::populate()
...
-               "    LEFT OUTER JOIN charass b"
-               "      ON (b.charass_target_type='PI'"
-               "      AND b.charass_target_id=:poitem_id"
-               "      AND b.charass_char_id=char_id) "
-               " WHERE ( (a.charass_char_id=char_id)"
-               "   AND   (a.charass_target_type='I')"
-               "   AND   (a.charass_target_id=:item_id) ) "
-               " ORDER BY char_name;" );
+                   "       COALESCE(b.charass_value, (SELECT c.charass_value FROM charass c WHERE ((c.charass_target_type='I') AND (c.charass_target_id=:item_id) AND (c.charass_default) AND (c.charass_char_id=char_id)) LIMIT 1)) AS charass_value"
+                   "   FROM (SELECT DISTINCT char_id, char_type, char_name, char_order "
```
ChristopherCotnoir: Why are char_type and char_order selected in this query and other queries below where they weren't originally present? They don't seem to be used.

guiclient/purchaseOrderItem.cpp
```diff
@@ -527,16 +527,26 @@ void purchaseOrderItem::populate()
...
+                   "         WHERE ((charass_char_id=char_id)"
+                   "         AND  (charass_target_type = 'PI' AND charass_target_id=:poitem_id))   ) AS data "
+                   "   LEFT OUTER JOIN charass b ON ((:poitem_id=b.charass_target_id)"
+                   "                                  AND ('PI'=b.charass_target_type)"
+                   "                                  AND (b.charass_char_id=char_id))"
+                   "   LEFT OUTER JOIN item     i1 ON (i1.item_id=:item_id)"
+                   "   LEFT OUTER JOIN charass  i2 ON ((i1.item_id=i2.charass_target_id)"
+                   "                                   AND ('I'=i2.charass_target_type)"
+                   "                                   AND (i2.charass_char_id=char_id)"
+                   "                                   AND (i2.charass_default))"
```
ChristopherCotnoir: Why these last two joins? They don't seem to be used. Same in cases below.

guiclient/purchaseOrderItem.cpp
andersdd: It's because you can have the same characteristic assigned to the same object multiple times with different or (I don't know why) the same values assigned.

guiclient/purchaseOrderItem.cpp
andersdd: This query pulls in Item characteristics also assigned to the PO Item even if they have not yet been assigned to the PO (default item characteristics).  So the second set of joins is required to bring in their potential values.  The change made here is to restrict those characteristics to only those actually pegged against the PO Item.  Also the UNION brings in assigned characteristics even if they are no longer pegged against PO Items.

guiclient/purchaseOrderItem.cpp
ChristopherCotnoir: I don't understand why that makes selecting those two necessary. They aren't used, so isn't the query the same either way?

andersdd: That's part of the original code - I've not touched it as I was trying to tread lightly - I'll have to go back and revisit the query.

ChristopherCotnoir: The original code from SalesOrderItem, that is. Since PurchaseOrderItem and WorkOrder don't use charass_price, among other things, their queries can be simpler, and originally were.

PR #1477, opened 2016-11-29
issue #22868 Inventory detail distribution locking

jcarlin:
*Dependent on:*
https://github.com/xtuple/xtuple/pull/2911

**Inventory > Transactions > Inventory Adjustment uses the new methods, sans locking.**

This pr removes locking for all transactions made from `Inventory > Transactions > Inventory Adjustment`. It does not resolve the incident, the changes required for the remaining `76` files that call `DistributeInventory::SeriesAdjust` will look similar to the changes below in `guiclient/adjustmentTrans.cpp`.

### More details:
**Old way:**
* Lock
* Perform inventory adjustment and inventory post related transactions
* Performing dist. detail related transactions
* Remove lock

**New way:** 
* Perform dist. detail related transactions:
  * Call the new `createItemlocdistSeries` function
  * Ask the user to fill in the lot/serial and location details in that itemlocdist series
  * If the user cancels filling in the details, return to the calling window without posting and orphan the itemlocdist records
* Perform inventory adjustment related transactions
* Post dist. detail and inventory records
* Delete dist. detail.


guiclient/adjustmentTrans.cpp
```diff
@@ -63,6 +65,8 @@ adjustmentTrans::adjustmentTrans(QWidget* parent, const char * name, Qt::WindowF
     _tab->removeTab(0);
 
   _item->setFocus();
+  _controlledItem = false;
+  _itemlocSeries = 0;
```
ChristopherCotnoir: Initialize _itemsiteId here as well.

guiclient/createLotSerial.cpp
```diff
@@ -68,16 +70,14 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
-               "       invhist_ordtype, invhist_transtype, invhist_ordnumber "
-               "FROM itemlocdist, itemsite, item, invhist "
-               "WHERE ( (itemlocdist_itemsite_id=itemsite_id)"
-               " AND (itemsite_item_id=item_id)"
-               " AND (itemlocdist_invhist_id=invhist_id) "
-               " AND (itemlocdist_id=:itemlocdist_id) );" );
+    createet.prepare( "SELECT item_fractional, itemsite_controlmethod, itemsite_item_id, "
+                      " itemsite_id, itemsite_perishable, itemsite_warrpurc, "
+                      " COALESCE(itemsite_lsseq_id,-1) AS itemsite_lsseq_id, "
+                      " itemlocdist_order_type, '' AS invhist_ordnumber " // TODO - these used to come from invhist. Review.
```
ChristopherCotnoir: This window is used in other places where the invhist entry will be available. Instead of removing invhist and using these, perform a left outer join on invhist and use coalesce to use these values if no invhist entry is found.

guiclient/adjustmentTrans.cpp
gilmoskowitz: _itemlocSeries is not needed at the class level. It's set and used within a single method and should be converted to a variable local to that method.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,25 +180,47 @@ void adjustmentTrans::sPost()
...
+    {
+      QMessageBox::information(this, tr("Inventory Adjustment"),
+                               tr("Transaction Canceled") );
+      return;
+    } 
+  }
+
+  //XSqlQuery rollback;
+  //rollback.prepare("ROLLBACK;");
+  //adjustmentPost.exec("BEGIN;");  // because of possible distribution cancelations
```
gilmoskowitz: Just remove the old code instead of commenting it out.

guiclient/adjustmentTrans.cpp
```diff
@@ -214,7 +241,7 @@ void adjustmentTrans::sPost()
     int result = adjustmentPost.value("result").toInt();
```
gilmoskowitz: `invadjustment()` can no longer return a negative value - :champagne: - so remove the check for that case entirely.

guiclient/adjustmentTrans.cpp
jcarlin: done

jcarlin: done

common/storedProcErrorLookup.cpp
```diff
@@ -874,6 +874,8 @@ const struct {
   { "insertIntoGLSeries", -4, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Cannot add to a G/L Series because the "
 				 "Accounting Period is closed."), 0, "" },
 
+  { "invAdjustment",  -1, "",  -1, "invAdjustment" },
+
```
ChristopherCotnoir: What is this supposed to do? It doesn't seem to do anything.

guiclient/adjustmentTrans.cpp
```diff
@@ -166,6 +172,8 @@ enum SetResponse adjustmentTrans::set(const ParameterList &pParams)
 void adjustmentTrans::sPost()
 {
   XSqlQuery adjustmentPost;
+  XSqlQuery cleanup;  
+  int _itemlocSeries = 0;
```
ChristopherCotnoir: Local variables should not start with underscore as per the Desktop Client Coding Standards.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,64 +184,109 @@ void adjustmentTrans::sPost()
...
+    if (newSeries.value("result").toInt() <= 0)
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error. "
+        "createitemlocdistseries(%, %, 'AD', '') Returned <= 0").arg(_itemsiteId).arg(qty),
+        newSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    _itemlocSeries = newSeries.value("result").toInt();
+    cleanup.prepare("SELECT deleteitemlocdistseries(:itemlocSeries) AS result;");
```
ChristopherCotnoir: "AS result" isn't needed, the return value doesn't need an alias as it is never used.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,64 +184,109 @@ void adjustmentTrans::sPost()
...
+  {
+    return;
+  }
+
+  // Create inventory transaction
+  adjustmentPost.prepare( "SELECT invAdjustment(itemsite_id, :qty, :docNumber, :comments, :date, "
+                          " :cost, NULL::INTEGER, :itemlocSeries) AS result "
+                          "FROM itemsite "
+                          "WHERE itemsite_item_id=:item_id "
+                          " AND itemsite_warehous_id=:warehous_id;");
```
ChristopherCotnoir: _itemsiteId can be used here just like above rather than fetching from itemsite.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,64 +184,109 @@ void adjustmentTrans::sPost()
...
   adjustmentPost.bindValue(":qty", qty);
   adjustmentPost.bindValue(":docNumber", _documentNum->text());
   adjustmentPost.bindValue(":comments",    _notes->toPlainText());
   adjustmentPost.bindValue(":item_id", _item->id());
   adjustmentPost.bindValue(":warehous_id", _warehouse->id());
   adjustmentPost.bindValue(":date",        _transDate->date());
+  if(_itemlocSeries != 0)
```
ChristopherCotnoir: This check isn't necessary.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,64 +184,109 @@ void adjustmentTrans::sPost()
...
     }
-
-    if (distributeInventory::SeriesAdjust(adjustmentPost.value("result").toInt(), this) == XDialog::Rejected)
+    else if (ErrorReporter::error(QtCriticalMsg, this, tr("Distribution Detail Posting Failed"),
+                              postDistDetail, __FILE__, __LINE__))
     {
-      rollback.exec();
-      QMessageBox::information(this, tr("Inventory Adjustment"),
-                               tr("Transaction Canceled") );
+      cleanup.exec();
```
ChristopherCotnoir: If an error is encountered at this point (and at several other points where cleanup is called), there are already many other things to clean up that aren't cleaned up by this call (like invhist entries, etc). They should be cleaned up too, or, if this isn't possible, then cleanup probably shouldn't be called because deleting itemlocdist records still referenced by other things not cleaned up is likely even worse than leaving them.

guiclient/adjustmentTrans.cpp
```diff
@@ -283,8 +340,12 @@ void adjustmentTrans::sPopulateQOH()
     _absolute->setStyleSheet("");
     if (populateAdjustment.first())
     {
+      _itemsiteId = populateAdjustment.value("itemsite_id").toInt();
       _cachedValue = populateAdjustment.value("itemsite_value").toDouble();
       _cachedQOH = populateAdjustment.value("itemsite_qtyonhand").toDouble();
+      _locControlled = populateAdjustment.value("itemsite_loccntrl").toDouble();
+      _lsControlled = populateAdjustment.value("lscntrl").toDouble();
```
ChristopherCotnoir: These should be toBool, not toDouble.

guiclient/assignLotSerial.cpp
```diff
@@ -213,10 +213,7 @@ void assignLotSerial::sAssign()
 
   assignAssign.prepare( "UPDATE itemlocdist "
              "SET itemlocdist_source_type='O' "
-             "WHERE (itemlocdist_series=:itemlocdist_series);"
-
-             "DELETE FROM itemlocdist "
-             "WHERE (itemlocdist_id=:itemlocdist_id);" );
+             "WHERE (itemlocdist_series=:itemlocdist_series);");
   assignAssign.bindValue(":itemlocdist_series", _itemlocSeries);
   assignAssign.bindValue(":itemlocdist_id", _itemlocdistid);
```
ChristopherCotnoir: This binding is no longer needed.

guiclient/createLotSerial.cpp
```diff
@@ -21,6 +21,8 @@
 #include "errorReporter.h"
 #include "guiErrorCheck.h"
 
+#define DEBUG true
+
```
ChristopherCotnoir: This macro isn't used, so there's no need to define it.

guiclient/distributeInventory.cpp
```diff
@@ -232,14 +248,14 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
               return XDialog::Rejected;
             }
           }
-
           assignLotSerial newdlg(pParent, "", true);
           newdlg.set(params);
           itemlocSeries = newdlg.exec();
           if (itemlocSeries == XDialog::Rejected)
             return XDialog::Rejected;
```
ChristopherCotnoir: If the function is exited at this point, there can be things that aren't properly cleaned up (like new ls table rows).

guiclient/distributeInventory.cpp
```diff
@@ -281,23 +297,47 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
...
+          if (DEBUG)
+            qDebug() << tr("ildsList.append(%1)").arg(itemlocSeries);
+          
           ildsList.append(itemlocSeries);
         }
+
+        // Set itemlocdist_child_series of parent itemlocdist record. Can this be removed since it's below?
+        query.prepare("UPDATE itemlocdist SET itemlocdist_child_series = :itemlocSeries "
+                      "WHERE itemlocdist_series = :pItemlocSeries "
+                      " AND :itemlocSeries IS NOT NULL "
```
ChristopherCotnoir: The bound value cannot be null. Should be checking if it is greater than -1.

guiclient/distributeInventory.cpp
```diff
@@ -125,7 +134,8 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
                      "            WHEN (invhist_transtype NOT IN ('RM','RP','RR','RX','IM')"
```
ChristopherCotnoir: invhist_transtype would be 'AD' if the invhist entry was already created (the way it worked before), but since it isn't it is NULL. This would be fine for most uses of invhist_transtype in this query, as the comparisons would yield the same result, but here NULL NOT IN (...) will evaluate to false, meaning before these changes if itemsite_location_dist is true auto_dist would become true, but now it would become false. To fix this, change invhist_transtype throughout this query to COALESCE(invhist_transtype, '').

guiclient/distributeInventory.cpp
```diff
@@ -281,23 +297,47 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
...
+        query.bindValue(":pItemlocSeries",  pItemlocSeries);
+        query.exec();
+        if (query.first())
+        {
+          if (query.count() != 1)
+          {
+            ErrorReporter::error(QtCriticalMsg, 0, tr("Updating Itemlocdist Parent Record Should Return One Row"),
+                           query, __FILE__, __LINE__);
+            return XDialog::Rejected;
+          }
```
ChristopherCotnoir: This is not necessarily the case. When performing an interwarehouse transfer for example, which also calls this function, there can be two itemlocdist entries with the passed series. This check should be removed or it may cause failures when this function is called from somewhere other than inventory adjustment.

guiclient/distributeInventory.cpp
```diff
@@ -281,23 +297,47 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
...
-      else
+      else // NOT a lot/serial controlled item (or a neg. qty, requiring "distribution from")
       {
         ParameterList params;
         params.append("itemlocdist_id", itemloc.value("itemlocdist_id").toInt());
         params.append("trans_type", itemloc.value("trans_type").toString());
 
         if (itemloc.value("itemlocdist_distlotserial").toBool())
-          params.append("includeLotSerialDetail");
+          params.append("includeLotSerialDetail", true); // true arg required here?
```
ChristopherCotnoir: No, it was fine as it was. The value doesn't matter, the check is merely to see if that parameter exists.

guiclient/distributeInventory.cpp
ChristopherCotnoir: Actually, this causes another problem. If there is more than one itemlocdist entry in the passed series, the itemlocdist_child_series for each of them can be set multiple times in this loop, and only the last value will remain. This means all the other series will not be processed by postdistdetail. The entire workaround here has to be redone another way to handle this.

guiclient/issueLineToShipping.cpp
```diff
@@ -304,6 +306,8 @@ void issueLineToShipping::sIssue()
   if (issue.first())
   {
     int result = issue.value("result").toInt();
+    if (DEBUG)
+      qDebug() << "IssueLineToShipping::sIssue result from issueToShipping: " << result;
```
ChristopherCotnoir: Why this change in this pull request? What does this have to do with any of the other changes?

guiclient/distributeInventory.cpp
jcarlin: From all of my testing, the "parent" record in `itemlocdist` has a unique `itemlocdist_series` value. This parent series is what is passed throughout all of the functions. It is very ambiguous but part of that is a flaw in the architecture of this whole table and related sql, which is not being changed in this pr. I will add some comments here and elsewhere.

guiclient/distributeInventory.cpp
jcarlin: All of these Rejected exits will be handled here: https://github.com/xtuple/qt-client/pull/1477/files#diff-692358a467ffa73470486904ba8beba3R227 (`deleteitemlocseries.sql`)

guiclient/adjustmentTrans.cpp
jcarlin: This is now handled in `deleteitemlocseries.sql` (see https://github.com/xtuple/qt-client/pull/1477/files#diff-692358a467ffa73470486904ba8beba3R220)

guiclient/distributeInventory.cpp
ChristopherCotnoir: If you have an item with at least two sites that are either location controlled or lot/serial controlled, in the application go to Inventory->Transactions->Site Transfer..., select the item and make the from and to sites both sites that are either location controlled or lot/serial controlled, enter a quantity, and click Post, and follow what happens with the current code and with the code from this and the related xtuple pull request. There will be two itemlocdist entries with the passed series when SeriesAdjust is called, which is why the results of the first query are in a loop. This will result in two child series being created, one for each of the two itemlocdist entries in the parent series. However, these updates will result in only one being stored in the itemlocdist_child_series of both parent entries by the time postdistdetail is called, leaving one unprocessed.

As a side note, when testing this I was surprised to find the "Updating Itemlocdist Parent Record Should Return One Row" error didn't get thrown. Turns out query.count() returns the number of columns, not rows. So the check as it is makes no sense anyway.

Pay close attention to the loops in this function and think about why they are loops in the first place, and you'll start seeing the problem. Things work fine in some cases, but Site Transfers at the very least will cause issues because the loops will start having more than one entry, which causes this code not to make sense.

jcarlin:
Thank you

guiclient/distributeInventory.cpp
ChristopherCotnoir: If the ls records are left alone, they will still show up in LotSerialClusters, even though they shouldn't exist. Gil says to find a way to delete them. It's likely impossible to tell new ls entries that need deleting from old ones as things are now by the time you get to deleteitemlocseries, you'll probably need more C++ code or another column to keep track of this.

Also, you need to consider what happens if a rejected exit occurs early, such as right here, before itemlocdist_child_series is set. The deleteitemlocseries function will be unable to find these child itemlocdist entries. In the WITH RECURSIVE query, you'll need to check for child.itemlocdist_source_id = _itemlocdist.itemlocdist_id as well to catch these.

guiclient/adjustmentTrans.cpp
ChristopherCotnoir: If invadjustment succeeds but postdistdetail fails, invalid gltrans/sltrans entries are left which cannot be deleted. The goal of this pull request is simply to ensure that no windows are open during a transaction block, because that can cause the block to last indefinitely. The second part of this function, after distributeInventory::SeriesAdjust (invadjustment and postdistdetail) can and should be wrapped in a transaction block, because they are immediate and require no user input. If something fails, simply rollback the transaction and then execute cleanup to clean up the changes from distributeInventory::SeriesAdjust.

guiclient/adjustmentTrans.cpp
```diff
@@ -176,64 +184,104 @@ void adjustmentTrans::sPost()
...
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error. "
+        "createitemlocdistseries(%, %, 'AD', '') Returned <= 0").arg(_itemsiteId).arg(qty),
+        newSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    itemlocSeries = newSeries.value("result").toInt();
+    cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, :invhistId, :failed);");
+    cleanup.bindValue(":itemlocSeries", itemlocSeries);
+    cleanup.bindValue(":failed", true);
```
ChristopherCotnoir: No need to bind a constant, just pass it directly.

jcarlin: Sure, I thought it was clearer to read this way.

common/storedProcErrorLookup.cpp
```diff
@@ -268,6 +268,9 @@ const struct {
 			       "an existing Tax Authority. Please choose a "
 			       "different number and save again."),	0, "" },
 
+  { "createItemlocdistSeries",  -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "The Resulting itemlocdist_series (%1) "
+        " Does Not Match the Series Passed (%2)"), 0, "" },
+
```
ChristopherCotnoir: This is no longer needed since the exception is no longer there.

guiclient/adjustmentTrans.cpp
```diff
@@ -253,12 +308,11 @@ void adjustmentTrans::sPost()
   else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Transaction"),
                                 adjustmentPost, __FILE__, __LINE__))
   {
-      rollback.exec();
```
ChristopherCotnoir: We still need to rollback if invadjustment fails.

guiclient/distributeInventory.cpp
```diff
@@ -324,50 +362,59 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
...
+                                      post, __FILE__, __LINE__))
+        {
+          return XDialog::Rejected;
+        }
+      }
+        
+      //Post inventory history for any remaining non-distributed transactions and trial balance
+      post.prepare("SELECT postItemlocseries(:itemlocseries) AS result;");
+      post.bindValue(":itemlocseries",  pItemlocSeries);
       post.exec();
```
ChristopherCotnoir: Why all this for the old method instead of just calling postdistdetail like before?

guiclient/distributeInventory.cpp
```diff
@@ -324,50 +362,59 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent, cons
...
+
+        post.prepare("SELECT distributeToLocations(:itemlocdist_id) AS result;");
+        post.bindValue(":itemlocdist_id", ildList.at(i));
+        post.exec();
+        if (ErrorReporter::error(QtCriticalMsg, 0, tr("Error Retrieving Lot/Serial Information"),
+                                      post, __FILE__, __LINE__))
+        {
+          return XDialog::Rejected;
+        }
+      }
```
gilmoskowitz: It should be possible to do this with a single database call instead of looping over a QList and issuing a db call in every iteration. The trick is to represent the QList as an object that PostgreSQL can treat as a table (see [unnest](https://www.postgresql.org/docs/9.3/static/functions-array.html), for example).

guiclient/distributeInventory.cpp
```diff
@@ -428,7 +475,7 @@ void distributeInventory::populate()
     else
       _bcQty->clear();
   }
-  else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Lot/Serial Information"),
+  else if (ErrorReporter::error(QtCriticalMsg, 0, tr("Error Retrieving Lot/Serial Information"),
```
gilmoskowitz: Why `0`? Either `this` or its parent would be good candidates.

guiclient/accountNumbers.cpp
```diff
@@ -244,7 +244,7 @@ void accountNumbers::sBuildList()
 
   if (_metrics->value("Application") != "PostBooks")
   {
-    _account->addColumn(tr("External"), _ynColumn, Qt::AlignCenter, false, "company_external");
+    _account->addColumn(tr("External"), 150, Qt::AlignCenter, false, "accnt_extref");
```
ChristopherCotnoir: This is from the merge, it's not actually part of the distribution locking stuff.

guiclient/distributeInventory.cpp
```diff
@@ -311,9 +311,9 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
 
         // Set itemlocdist_child_series of parent itemlocdist record
         query.prepare("UPDATE itemlocdist SET itemlocdist_child_series = :itemlocSeries "
-                      "WHERE itemlocdist_id = :itemlocdist_id;");
+                      "WHERE itemlocdist_series = :pItemlocSeries;"); //itemlocdist_id = :itemlocdist_id
```
ChristopherCotnoir: Why this change? It looked right before, this has the same problem it originally had.

guiclient/distributeInventory.cpp
```diff
@@ -360,61 +360,58 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
...
+    }
+    
+    // Post distribution detail pre-incident #28868 (the old way)
+    if (!pPreDistributed) 
+    {
+      XSqlQuery postDistDetail;
+      postDistDetail.prepare("SELECT postdistdetail(:itemlocSeries, NULL::INTEGER, :ls, :loc) AS result;");
+      postDistDetail.bindValue(":itemlocSeries", pItemlocSeries);
+      postDistDetail.bindValue(":ls", lsCntrld);
+      postDistDetail.bindValue(":loc", locCntrld);
```
ChristopherCotnoir: These last two values are set INSIDE the loop, meaning they can be set more than once. Only the last set of values will be preserved, and there could be multiple itemlocdist entries with the passed series but different ls/loc control. Rework the function to not require these parameters, the information is available just from the parent itemlocseries.

jcarlin:
@ChristopherCotnoir The 5 major pending todos are addressed here in the latest. I am still doing testing in the client. The latest update to `postDistDetail` simply uses the `itemlocdist_id`s that make up `ildList` array in `distributeInventory.cpp` by looking at the new `itemlocdist_reqdisttolocations` column. I really didn't want to add a column but believe it's best to maintain all of the logic in `DistributeInventory`; all of this because we want to be sure not to call `distributeToLocations` unnecessarily in `postDistDetail` _and_ I want to tally up the results from the posting functions. Otherwise, we can remove the column and go back to mirroring the distributeInventory logic in postDistDetail...

PR #1484, opened 2016-12-07
issue #29012 Don't match empty string

ChristopherCotnoir:


ChristopherCotnoir:
I am aware of this, however, since no entry is selected when the window is first opened and the search field is blank, it seems that no entry should be selected when the search field reverts to being blank. What is basically happening is that all the entries are considered to 'match' the empty string in the search field (only the first is picked, of course), which seems wrong as that's not what the search field is for. It's an incredibly minor issue, but I still see reason to make the change.

PR #1486, opened 2016-12-08
issue #29043 Have All Sites as an option in the dropdown

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#2919.

gilmoskowitz:
This addresses the bug as reported but does not handle the general case. Consider the ramifications of changing the parameter widget instead.

gilmoskowitz:
On second thought, this isn't a bug at all. If all sites are desired then the user should remove sites from the list of filters. That's what the - button is for.

ChristopherCotnoir:
The All Sites filter I added shows item sources with access to all sites, but NOT ones with access to only one site, so it does still function as a filter. Whether this filter is still needed, and whether filtering on one site should show sites with access to all sites (currently it doesn't) is up for discussion.

guiclient/displays/dspItemSources.cpp
```diff
@@ -35,9 +35,16 @@ dspItemSources::dspItemSources(QWidget* parent, const char*, Qt::WindowFlags fl)
 //  setNewVisible(true);
 //  setQueryOnStartEnabled(true);
 
+  QString qrySite = "SELECT 0, 'All Sites' AS name, '' AS code "
```
gilmoskowitz: Given prior comments on this issue and discussion with Support, let's change the text from `All Sites` to `[ Not Site Restricted ]`. We also need to make this string translatable.

The rationale is that `All Sites` in all(?) other filters in the application means "Show all rows regardless of the site/warehous_id;" here it means explicitly "Show all rows with a certain value, where that certain value means 'all sites/warehouses'."

PR #1496, opened 2016-12-20
Issue #26765 Landed Cost

andersdd:
Freight distribution to PO Voucher Items

Also resolves #26390 and probably other issues

requires xtuple/xtuple#2924

### DIFF
--- a/common/storedProcErrorLookup.cpp
+++ b/common/storedProcErrorLookup.cpp
@@ -54,6 +54,11 @@ const struct {
                             "it is already associated with an Opportunity.  You must "
                             "detach this Sales Order before you may attach it."),    0, "" },
 
+  { "calculateFreightDistribution", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No Items have been distributed to the Voucher."
+                                    "Please distribute items before attempting to distribute freight."),     0, "" },
+  { "calculateFreightDistribution", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "The Voucher you are attempting to distribute freight "
+                                    "to has incorrect item totals for the desired distribution method. Please check that Voucher"),     0, "" },
+
   { "changeCMHeadTaxAuth", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "This Return was not found."),	0, "" },
   { "changeCMHeadTaxAuth", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "This Tax Authority was not found."),	0, "" },
 
@@ -776,12 +781,6 @@ const struct {
                              "Upgrade to PostgreSQL 8.2 or later."),    0, "" },
   { "disablePackage", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Could not find a package with the internal id % "
                              "to enable or disable."),                  0, "" },
-
-  { "distributeVoucherFreight", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No Items have been distributed to the Voucher."
-                                    "Please distribute items before attempting to distribute freight."),     0, "" },
-  { "distributeVoucherFreight", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "An error occurred distributing freight to the Voucher."
-                                    "Please contact your administrator to resolve this issue."),     0, "" },
-
 
   { "distributeVoucherLine", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Distribution would result in zero "
                                     "quantity and amount."
--- a/common/storedProcErrorLookup.cpp
+++ b/common/storedProcErrorLookup.cpp
@@ -54,8 +54,9 @@ const struct {
                             "it is already associated with an Opportunity.  You must "
                             "detach this Sales Order before you may attach it."),    0, "" },
 
-  { "calculateFreightDistribution", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No Items have been distributed to the Voucher."
-                                    "Please distribute items before attempting to distribute freight."),     0, "" },
+  { "calculateFreightDistribution", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No valid Items have been distributed to the Voucher."
+                                    " Items must be Purchased goods and be costed items.  Please check and distribute items before"
+                                    " attempting to distribute freight."),     0, "" },
   { "calculateFreightDistribution", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "The Voucher you are attempting to distribute freight "
                                     "to has incorrect item totals for the desired distribution method. Please check that Voucher"),     0, "" },
 
--- a/common/storedProcErrorLookup.cpp
+++ b/common/storedProcErrorLookup.cpp
@@ -777,6 +777,12 @@ const struct {
   { "disablePackage", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Could not find a package with the internal id % "
                              "to enable or disable."),                  0, "" },
 
+  { "distributeVoucherFreight", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No Items have been distributed to the Voucher."
+                                    "Please distribute items before attempting to distribute freight."),     0, "" },
+  { "distributeVoucherFreight", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "An error occurred distributing freight to the Voucher."
+                                    "Please contact your administrator to resolve this issue."),     0, "" },
+
+
   { "distributeVoucherLine", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Distribution would result in zero "
                                     "quantity and amount."
                                     "Please distribute manually."),     0, "" },
--- a/guiclient/voucher.cpp
+++ b/guiclient/voucher.cpp
@@ -57,7 +57,7 @@ voucher::voucher(QWidget* parent, const char* name, Qt::WindowFlags fl)
   connect(_amountToDistribute,       SIGNAL(idChanged(int)),                            this,          SLOT(sFillList()));
   connect(_amountDistributed,        SIGNAL(valueChanged()),                            this,          SLOT(sPopulateBalanceDue()));
   connect(_freight,                  SIGNAL(valueChanged()),                            this,          SLOT(sPopulateDistributed()));
-  connect(_freightDistr,             SIGNAL(clicked()),                                 this,          SLOT(sFreightDistribution()));
+  connect(_freightDistr,             SIGNAL(toggled(bool)),                             this,          SLOT(sFreightDistribution()));
   connect(_distributeFreight,        SIGNAL(clicked()),                                 this,          SLOT(sDistributeFreight()));
 
   _terms->setType(XComboBox::APTerms);
@@ -187,11 +187,7 @@ enum SetResponse voucher::set(const ParameterList &pParams)
 //      _documents->setReadOnly(true);
       _charass->setReadOnly(true);
       _close->setText(tr("&Close"));
-      _freight->setEnabled(false);
-      _freightExpcat->setEnabled(false);
-      _freightDistr->setEnabled(false);
-      _freightStack->setEnabled(false);
-
+      _freightGroup->setEnabled(false);
 
       _save->hide();
 
@@ -216,7 +212,6 @@ enum SetResponse voucher::set(const ParameterList &pParams)
     enableWindowModifiedSetting();
   }
 
-  sFreightDistribution();
   return NoError;
 }
 
@@ -302,7 +297,7 @@ bool voucher::sSave()
              "    vohead_amount=:vohead_amount,"
              "    vohead_freight=:vohead_freight,"
              "    vohead_freight_expcat_id=:vohead_freight_expcat,"
-             "    vohead_freight_distributed=COALESCE(:frghtDistr, 0.00),"
+             "    vohead_freight_distributed=:frghtDistr,"
              "    vohead_1099=:vohead_1099, "
              "    vohead_curr_id=:vohead_curr_id, "
              "    vohead_notes=:vohead_notes "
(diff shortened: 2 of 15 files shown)


common/storedProcErrorLookup.cpp
```diff
@@ -777,6 +777,12 @@ const struct {
   { "disablePackage", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Could not find a package with the internal id % "
                              "to enable or disable."),                  0, "" },
 
+  { "distributeVoucherFreight", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "No Items have been distributed to the Voucher."
+                                    "Please distribute items before attempting to distribute freight."),     0, "" },
+  { "distributeVoucherFreight", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "An error occurred distributing freight to the Voucher."
+                                    "Please contact your administrator to resolve this issue."),     0, "" },
```
ChristopherCotnoir: These are actually from calculateFreightDistribution, not distributeVoucherFreight. The procName should be changed here and in the RAISE EXCEPTIONs in calculateFreightDistribution. The first message also makes the assumption that the error was raised when the user was attempting to distribute freight, which isn't necessarily the case, as calculateFreightDistribution is also called from postVoucher. The second message should probably be more specific, like the message in the RAISE EXCEPTION call.

guiclient/voucher.cpp
```diff
@@ -57,6 +57,8 @@ voucher::voucher(QWidget* parent, const char* name, Qt::WindowFlags fl)
   connect(_amountToDistribute,       SIGNAL(idChanged(int)),                            this,          SLOT(sFillList()));
   connect(_amountDistributed,        SIGNAL(valueChanged()),                            this,          SLOT(sPopulateBalanceDue()));
   connect(_freight,                  SIGNAL(valueChanged()),                            this,          SLOT(sPopulateDistributed()));
+  connect(_freightDistr,             SIGNAL(clicked()),                                 this,          SLOT(sFreightDistribution()));
```
ChristopherCotnoir: Prefer toggled(bool) to clicked() for check boxes.

guiclient/voucher.cpp
```diff
@@ -176,7 +189,10 @@ enum SetResponse voucher::set(const ParameterList &pParams)
       _close->setText(tr("&Close"));
       _freight->setEnabled(false);
       _freightExpcat->setEnabled(false);
-   
+      _freightDistr->setEnabled(false);
+      _freightStack->setEnabled(false);
```
ChristopherCotnoir: Now that all the freight components are in a group box, the whole group box can simply be disabled instead of disabling each component.

guiclient/voucher.cpp
```diff
@@ -200,6 +216,7 @@ enum SetResponse voucher::set(const ParameterList &pParams)
     enableWindowModifiedSetting();
   }
 
+  sFreightDistribution();
```
ChristopherCotnoir: This call isn't needed if the signal above is changed to toggled and if the currentIndex in the ui file is set to 0.

guiclient/voucher.cpp
```diff
@@ -283,6 +302,7 @@ bool voucher::sSave()
              "    vohead_amount=:vohead_amount,"
              "    vohead_freight=:vohead_freight,"
              "    vohead_freight_expcat_id=:vohead_freight_expcat,"
+             "    vohead_freight_distributed=COALESCE(:frghtDistr, 0.00),"
```
ChristopherCotnoir: The COALESCE is unnecessary, as the variable being bound is already initialized in the constructor.

guiclient/voucher.cpp
```diff
@@ -301,8 +321,9 @@ bool voucher::sSave()
   updq.bindValue(":vohead_reference", _reference->text().trimmed());
   updq.bindValue(":vohead_amount", _amountToDistribute->localValue());
   updq.bindValue(":vohead_freight", _freight->localValue());
+  updq.bindValue(":frghtDistr", _frghtdistr);
   if (_freightExpcat->isValid())
-    updq.bindValue(":vohead_freight_expcat", _freightExpcat->id());
+    updq.bindValue(":vohead_freightexpcat", _freightExpcat->id());
```
ChristopherCotnoir: This underscore should not be removed.

common/storedProcErrorLookup.cpp
andersdd: I moved the logic int a sub-function hence the discrepancy in naming.  Will resolve.  The first error can never be hit in postVoucher due to the workflow (voucher already posted) so is only relevant when manually distributing the freight.

guiclient/voucher.cpp
```diff
@@ -722,7 +777,9 @@ void voucher::sPopulateBalanceDue()
     _freightExpcat->setId(-1);
 
   _freightExpcat->setEnabled(_freight->localValue() > 0 && _mode != cView);
- 
+  _freightDistr->setEnabled(_freight->localValue() > 0 && _mode != cView);
+  _freightStack->setEnabled(_freight->localValue() > 0 && _mode != cView);
```
ChristopherCotnoir: _freightExpcat's setEnabled can be removed because it is inside _freightStack now.

guiclient/voucher.cpp
```diff
@@ -901,6 +965,7 @@ bool voucher::saveDetail()
                "    vohead_amount=:vohead_amount,"
                "    vohead_freight=:vohead_freight,"
                "    vohead_freight_expcat_id=:vohead_freightexpcat,"
+               "    vohead_freight_distributed=COALESCE(:frghtDistr, 0.00),"
```
ChristopherCotnoir: This COALESCE is also unneeded.

guiclient/voucher.cpp
```diff
@@ -1020,3 +1086,88 @@ void voucher::sDataChanged()
   setWindowModified(true);
 }
 
+void voucher::sFreightDistribution()
+{
+  if (_freightDistr->isChecked())
+    _freightStack->setCurrentIndex(1);
+  else
+    _freightStack->setCurrentIndex(0);
+}
```
ChristopherCotnoir: With toggled, this function is passed _freightDistr->isChecked() already. Consider using that for easier reading.

guiclient/voucher.cpp
```diff
@@ -1020,3 +1086,88 @@ void voucher::sDataChanged()
...
+  distr.prepare("SELECT SUM( (SELECT COALESCE(SUM(recv_qty), 0)"
+                " FROM recv "
+                " WHERE (recv_posted)"
+                "  AND (NOT recv_invoiced)"
+                "  AND (recv_vohead_id IS NULL)"
+                "  AND (recv_order_type='PO')"
+                "  AND (recv_orderitem_id=poitem_id)) ) AS qtyreceived"
+                " FROM poitem JOIN pohead ON (pohead_id=poitem_pohead_id)"
+                " WHERE (poitem_pohead_id=:poitem_id);");
+  distr.bindValue(":poitem_id", _poNumber->id());
```
ChristopherCotnoir: I think :poitem_id should be :pohead_id. Also, the subquery and two sums are unnecessary, as well as the join on pohead. Should be able to be simplified:

SELECT SUM(COALESCE(recv_qty, 0)) AS qtyreceived
FROM poitem JOIN recv ON poitem_id=recv_orderitem_id
WHERE poitem_pohead_id=:pohead_id
AND recv_posted
AND NOT recv_invoiced
AND recv_vohead_id IS NULL
AND recv_order_type='PO';

guiclient/voucher.cpp
```diff
@@ -1020,3 +1086,88 @@ void voucher::sDataChanged()
...
+  distr.exec();
+  if (distr.first())
+  {
+    if (distr.value("qtyreceived").toInt() > 0)
+    {
+      if (!QMessageBox::question(this, tr("Incomplete Distribution"),
+                              tr("The Purchase Order has not been "
+                                 "completely distributed. "
+                                 "Do you wish to continue?"),
+                                  QMessageBox::Yes | QMessageBox::No, QMessageBox::No))
```
ChristopherCotnoir: !QMessageBox::No does not evaluate to true. Compare to QMessageBox::No instead as above.

guiclient/voucher.cpp
```diff
@@ -1020,3 +1086,88 @@ void voucher::sDataChanged()
...
+    distr.bindValue(":dist_type", "V");
+  else if (_freightDistWgt->isChecked())
+    distr.bindValue(":dist_type", "W");
+  distr.bindValue(":freight", _freight->localValue());
+
+  distr.exec();
+  if (distr.first())  
+  {
+    int returnVal = distr.value("result").toInt();
+    if (returnVal < 0)
```
ChristopherCotnoir: This is not possible, as RAISE EXCEPTIONs are used instead of negative error codes in that function (or rather what it calls). Since we are looking to phase out negative error codes and their checks, checking for a negative error code here can be skipped.

guiclient/voucher.cpp
```diff
@@ -1020,3 +1086,88 @@ void voucher::sDataChanged()
...
+    {
+      ErrorReporter::error(QtCriticalMsg, this, storedProcErrorLookup("calculateFreightDistribution", returnVal),
+                         distr, __FILE__, __LINE__);
+      return;
+    }
+    _frghtdistr = _freight->localValue();
+    sIsDistributed();
+    sFillList();
```
ChristopherCotnoir: I think sPopulateDistributed also needs to be called here since _frghtdistr has been changed.

guiclient/voucherMiscDistrib.cpp
```diff
@@ -81,6 +93,9 @@ enum SetResponse voucherMiscDistrib::set(const ParameterList &pParams)
       param = pParams.value("amount", &valid);
       if (valid)
         _amount->setLocalValue(param.toDouble());
+
+      if (_metrics->value("DefaultFreightCostElement") > 0)
```
ChristopherCotnoir: toInt() needed.

guiclient/voucherMiscDistrib.cpp
```diff
@@ -126,29 +153,26 @@ void voucherMiscDistrib::populate()
...
+         << GuiErrorCheck(_expcatSelected->isChecked() && !_expcat->isValid(),
+                           _expcat,
+                           tr("<p>You must select an Expense Category to post this "
+                              "Miscellaneous Distribution to."))
+         << GuiErrorCheck(_taxSelected->isChecked() && !_taxCode->isValid(),
+                           _taxCode,
+                           tr("<p>You must select an Tax Code to post this "
+                              "Miscellaneous Distribution to."))
+         << GuiErrorCheck(_freightSelected->isChecked() && !_origVoucher->isValid(),
+                           _taxCode,
```
ChristopherCotnoir: Better to use _origVoucher here instead of _taxCode.

widgets/voucherCluster.cpp
```diff
@@ -0,0 +1,27 @@
+/*
+ * This file is part of the xTuple ERP: PostBooks Edition, a free and
+ * open source Enterprise Resource Planning software suite,
+ * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
```
ChristopherCotnoir: s/2014/2016/

widgets/voucherCluster.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "voucherCluster.h"
+
+VoucherCluster::VoucherCluster(QWidget* pParent, const char* pName) :
+    VirtualCluster(pParent, pName)
+{
+    addNumberWidget(new VoucherClusterLineEdit(this, pName));
+}
+
+VoucherClusterLineEdit::VoucherClusterLineEdit(QWidget* pParent, const char* pName) :
+    VirtualClusterLineEdit(pParent, "vohead", "vohead_id", "vohead_number", "(SELECT vend_number FROM vendinfo v WHERE v.vend_id=vohead_vend_id)", 0, 0, pName)
```
ChristopherCotnoir: The alias v for vendinfo isn't necessary.

widgets/widgets.pro
```diff
@@ -281,6 +283,7 @@ HEADERS += widgets.h \
     vendorcluster.h \
     vendorgroup.h \
     virtualCluster.h \
+    vouchercluster.h \
```
ChristopherCotnoir: The C is capitalized in the actual file name, so it probably should be here too. I'm not sure if it will make a difference, but I wouldn't be surprised if it caused problems on *NIX systems.

guiclient/voucherMiscDistrib.cpp
```diff
@@ -295,8 +345,10 @@ void voucherMiscDistrib::sPopulateVendorInfo(int pVendid)
 
 void voucherMiscDistrib::sHandleSelection()
 {
-  _taxTypeLit->setEnabled(_accountSelected->isChecked() || _expcatSelected->isChecked());
-  _taxType->setEnabled(_accountSelected->isChecked() || _expcatSelected->isChecked());
-  _discountable->setEnabled(_accountSelected->isChecked() || _expcatSelected->isChecked());
-  _discountable->setChecked(_accountSelected->isChecked() || _expcatSelected->isChecked());
+  bool _onoff = _accountSelected->isChecked() || _expcatSelected->isChecked()
```
ChristopherCotnoir: Local variables should not start with underscore as per the Desktop Client Coding Standards.

guiclient/voucher.cpp
```diff
@@ -427,18 +443,36 @@ void voucher::sDistributions()
...
     {
       int result = distq.value("result").toInt();
+      if (result == -5 &&
+          QMessageBox::question(this, tr("Error Distributing"),
+            tr("Item has multiple cost elements. Do you wish to distribute "
+               "to the Material cost element only?"),
+                                  QMessageBox::Yes | QMessageBox::No, QMessageBox::No) == QMessageBox::No)
+      {
+        return;
+      } else {
```
ChristopherCotnoir: Braces should have their own line as per the Desktop Client Coding Standards.

guiclient/voucher.cpp
andersdd: Agree on the pohead comments.  Unfortunately your simplified query does not work as it still returns NULL under some situations.  Current query may be more complex but it always returns an integer value

guiclient/voucher.cpp
andersdd: Disagree with you here.  This is how the Desktop standards state you should intercept EXCEPTIONS and provide a means to translate them in storedProcErrorLookup

guiclient/voucher.cpp
andersdd: I think this signal is called elsewhere as the distributed amount is updated regardless.

widgets/widgets.pro
andersdd: I have corrected this, but FYI it did not seem to cause problems on my *NIX system

guiclient/voucher.cpp
ChristopherCotnoir: Ah, right. Moving the COALESCE outside the SUM should fix that though, right?

PR #1497, opened 2016-12-22
issue #28624 Don't create empty cntcteml entries

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/connect#81 and xtuple/xtuple#2926.

PR #1498, opened 2016-12-27
Issue #28606: Characteristic Enhancements

andersdd:
Adds characteristic groups and numeric characteristics

requires xtuple/xtuple#2927


andersdd:
Looks like it is failing at

gilmoskowitz:
http://doc.qt.io/qt-4.8/qcombobox.html doesn't show `setCurrentText()` and Travis-CI makes it hard for us to build with Qt 5.

andersdd:
Do we develop in Qt 4.8 or 5.5?  If we say 4.8 then the entire Characteristics Group feature will not work correctly.

I need to check the Parameter Widget.

Validation is done in the characteristicAssignment widget line 88

gilmoskowitz:
We build for Qt 5; we just haven't had the time to fully set up an alternate CI service that can handle that.

PR #1500, opened 2016-12-28
Issue #29110: Prevent setting control method to None when Item Site

andersdd:
holds inventory quantity

### DIFF
--- a/guiclient/itemSite.cpp
+++ b/guiclient/itemSite.cpp
@@ -1149,6 +1149,14 @@ void itemSite::sHandleWOSupplied(bool pSupplied)
 
 void itemSite::sHandleControlMethod()
 {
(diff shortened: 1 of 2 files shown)


guiclient/itemSite.cpp
```diff
@@ -1149,6 +1149,14 @@ void itemSite::sHandleWOSupplied(bool pSupplied)
 
 void itemSite::sHandleControlMethod()
 {
+  if (_wasControlMethod != "N" && _controlMethod->code() == "N" && _qohCache > 0)
```
gilmoskowitz: We should initialize `_wasControlMethod` in the constructor so `sHandleControlMethod()` behaves sanely if inadvertently called before `populate()`.

PR #1504, opened 2016-12-30
Issue #28568: disable ability to add inactive items to price schedules

andersdd:


guiclient/itemPricingScheduleItem.cpp
```diff
@@ -81,7 +81,7 @@ itemPricingScheduleItem::itemPricingScheduleItem(QWidget* parent, const char* na
   _pricingRatio->setPrecision(omfgThis->percentVal());
   _stdMargin->setPrecision(omfgThis->percentVal());
   _actMargin->setPrecision(omfgThis->percentVal());
-  _item->setType(ItemLineEdit::cSold);
+  _item->setType(ItemLineEdit::cSold | ItemLineEdit::cActive);
```
gilmoskowitz: This is probably going to behave incorrectly when a pricing schedule item is opened for an inactive Item.

andersdd: Just tested and (somewhat surprisingly) it works fine without needing any changes.  An inactive item is still displayed in the widget.

PR #1519, opened 2017-01-24
Issue #29238: SQL in characteristic query when editing PO Items

andersdd:
This may conflict with PR #1498


gilmoskowitz:
`select ... from (select ... from (select ... )...)...` - yuck but it make sense

PR #1520, opened 2017-01-25
issue #9241 - cache combobox data

gilmoskowitz:
this is the first step in rewriting the xcombobox to solve several problems. future steps include fixing issue #25683 to avoid data corruption when documents refer to inactive records, using the model/view system the way qt intended, and perhaps driving this widget from the source table instead of using hard-coded queries.
requires xtuple/xtuple#2953 which adds `notify` on the database side


bendiy:
If the Qt client looses it's database connection and it reconnects, how does the xcombobox `listen` part in the Qt client get handled? Does it need to re-listen after re-connecting?

widgets/xcombobox.cpp
```diff
@@ -30,96 +31,812 @@
...
+                           "       salesrep_number || '-' || salesrep_name,"
+                           "       salesrep_number"
+                           "  FROM salesrep"
+                           " <? if exists('activeOnly') ?>WHERE salesrep_active<? endif ?>"
+                           " ORDER by salesrep_number;");
+static QString termsMQL("SELECT terms_id,"
+                        "       terms_code || '-' || terms_descrip, terms_code"
+                        "  FROM terms"
+                        " <? if exists('isAR') ?>WHERE terms_ar<? endif ?>"
+                        " <? if exists('isAP') ?>WHERE terms_ap<? endif ?>"
```
bendiy: Avoid possible invalid query if both params ever exist at the same time:
```sql
 WHERE true
  <? if exists('isAR') ?>AND terms_ar<? endif ?>
  <? if exists('isAP') ?>AND terms_ap<? endif ?>
```

widgets/xcombobox.cpp
```diff
@@ -30,96 +31,812 @@
...
+                       "", "",
+                       "SELECT regtype_id, regtype_code, regtype_code"
+                       "  FROM regtype"
+                       " ORDER BY regtype_code;", "regtype", false));
+    typeDescrip.insert(XComboBox::Reports,
+                       new XComboBoxDescrip(XComboBox::Reports,
+                       "reports", "MaintainReports",
+                       "SELECT a.report_id, a.report_name, a.report_name"
+                       "  FROM report a"
+                       "  JOIN (SELECT MIN(report_grade) AS report_grade, report_name"
```
bendiy: Shouldn't we be selecting the `MAX(report_grade)` to get the active report?

guiclient/guiclient.cpp
```diff
@@ -1044,42 +955,33 @@ void GUIClient::sTick()
...
-    // before we go doing something rash.
-    if (!QSqlDatabase::database().isOpen())
+    emit dbConnectionLost();
+    if (QMessageBox::question(this, tr("Database disconnected"),
+                              tr("It appears that you have been disconnected from the "
+                                 "database. Select Yes to try to reconnect or "
+                                 "No to terminate the application."),
+                                 QMessageBox::Yes | QMessageBox::No, QMessageBox::No) == QMessageBox::No)
+      qApp->quit();
+    else if (QSqlDatabase::database().open())
```
gilmoskowitz:
```
if the user clicks no [or hits the enter key]
  quit;
else if the database can be opened {
  call login();
  deal with login() errors;
}
```

widgets/xcombobox.cpp
gilmoskowitz: It doesn't matter which one we select here because we rarely use the id directly for reports. See line 1202-1234, for example (which is more complicated than necessary but that's a problem for another day).

PR #1521, opened 2017-01-26
Issue #29315: Resolve totals display on Standard Journals

andersdd:


guiclient/standardJournal.cpp
```diff
@@ -260,21 +258,21 @@ void standardJournal::sFillList()
...
-  standardFillList.prepare( "SELECT SUM( CASE WHEN (stdjrnlitem_amount < 0) THEN (stdjrnlitem_amount * -1)"
+  standardFillList.prepare( "SELECT formatMoney(SUM( CASE WHEN (stdjrnlitem_amount < 0) THEN (stdjrnlitem_amount * -1)"
              "                 ELSE 0"
-             "            END ) AS debit,"
-             "       SUM( CASE WHEN (stdjrnlitem_amount > 0) THEN stdjrnlitem_amount"
+             "            END )) AS debit,"
+             "       formatMoney(SUM( CASE WHEN (stdjrnlitem_amount > 0) THEN stdjrnlitem_amount"
              "                 ELSE 0"
-             "            END ) AS credit,"
+             "            END )) AS credit,"
```
gilmoskowitz: We're trying hard to get rid of `formatMoney`, `formatDate`, etc. function calls. They make the next phase of localization much harder and are recognized performance hogs. What's the benefit of calling them here instead of letting the fields format their data?

andersdd: @gilmoskowitz am I missing something? - because the existing logic was not formatting correctly in the fields, hence my change to force the format here.

gilmoskowitz: @andersdd - these fields on screen are never modified by the user from what i can tell. therefore it's better to use an `XLabel` than a disabled `XLineEdit`. `XLabel::setDouble()` is locale-aware and should format the numeric value properly.

PR #1522, opened 2017-02-15
issue #22868: Inventory Distribution Locking part 2

jcarlin:
https://github.com/xtuple/xtuple/pull/2961
https://github.com/xtuple/private-extensions/pull/889

The following inventory transactions have been updated to distribute inventory detail before posting inventory transactions:
- Material Receipt
- Expense Transaction
- Scrap Transaction
- Post Invoice
- Post PO Returns
- Enter PO Receipt
- Issue WO Material
- Issue WO Material Batch
- Return WO Material


guiclient/distributeInventory.cpp
```diff
@@ -324,24 +306,17 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
...
-            ildList.append(itemloc.value("itemlocdist_id").toInt());
-          else
+          if (!newdlg.sDefaultAndPost())
           {
-            result = newdlg.exec();
-            if (result == XDialog::Rejected)
+            if (newdlg.exec() == XDialog::Rejected)
               return XDialog::Rejected;
-            else
-              ildList.append(result);
```
jcarlin: Should we log these `itemlocdist_id`s using qDebug()?

guiclient/createLotSerial.cpp
```diff
@@ -102,15 +102,14 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
-                        "                              WHERE itemsite_item_id = (SELECT itemsite_item_id "
-                        "                               FROM itemsite WHERE itemsite_id = :itemsite))) "
-                        "AND (lsdetail_qtytoassign > 0) ) "
-                        "GROUP BY 2,3");
+      preassign.prepare("SELECT MAX(lsdetail_id) AS lsdetail_id, ls_number "
+                        "FROM lsdetail "
+                        " JOIN ls ON ls_id = lsdetail_ls_id "
+                        "WHERE lsdetail_source_number = :docnumber "
+                        " AND lsdetail_source_type = :transtype "
+                        " AND lsdetail_itemsite_id = :itemsite "
```
ChristopherCotnoir: Previously, this part checked that the itemsite referenced by lsdetail_itemsite_id was an itemsite for the same item as the passed itemsite. Now the itemsite has to be exactly the same (that is, it doesn't count same item, different warehouses), which is incorrect. Instead add 'JOIN itemsite ON ls_item_id=itemsite_item_id' and use 'AND itemsite_id = :itemsite'.

guiclient/createLotSerial.cpp
```diff
@@ -364,32 +363,30 @@ void createLotSerial::sAssign()
...
-                         "(SELECT itemloc_id AS count "
-                         "FROM itemsite JOIN itemloc ON (itemloc_itemsite_id=itemsite_id)"
-                         "              JOIN ls ON (ls_id=itemloc_ls_id) "
-                         "WHERE (itemsite_item_id=:item_id)"
-                         "  AND (UPPER(ls_number)=UPPER(:lotserial))"
+    createAssign.prepare("SELECT true "
+                         "FROM itemsite "
+                         "  JOIN itemloc ON itemloc_itemsite_id = itemsite_id "
+                         "  JOIN ls ON ls_id = itemloc_ls_id "
+                         "WHERE itemsite_id = :itemsite_id "
```
ChristopherCotnoir: This should still be using itemsite_item_id=:item_id. ls numbers are not allowed to be duplicated with the same item, even if the warehouse is different. Also add back the :item_id binding below and remove :itemsite_id.

guiclient/enterPoReceipt.cpp
```diff
@@ -271,7 +275,11 @@ void enterPoReceipt::sPost()
...
   while(qi.next())
-  {
+  { 
+    // Stage cleanup function to be called on error
+    XSqlQuery cleanup;
+    cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+
     if(_singleLot->isChecked() && !gotlot && qi.value("itemsite_controlmethod").toString() == "L")
     {
       getLotInfo newdlg(this, "", true);
```
ChristopherCotnoir: charass entries get inserted from this window that aren't cleaned up. Any ls entry that is deleted by deleteitemlocseries should also have any charass entries that reference them deleted.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,76 @@ void enterPoReceipt::sPost()
...
+      parentItemlocdist.bindValue(":itemsite_id", qi.value("itemsite_id").toInt());
+      parentItemlocdist.bindValue(":qty", qi.value("qty").toDouble());
+      parentItemlocdist.bindValue(":orderType", qi.value("recv_order_type").toString());
+      parentItemlocdist.bindValue(":orderNumber", qi.value("recv_order_number").toString());
+      parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+      parentItemlocdist.exec();
+      if (parentItemlocdist.first())
+      {
+        if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+          QDate(), true) == XDialog::Rejected)
```
ChristopherCotnoir: Don't pass empty QString/QDate here, pass lotnum, expdate, warrdate like before for if the getLotInfo window is used.

guiclient/createLotSerial.cpp
jcarlin: OK. Can't this skip the itemsite table altogether and in the where clause: `AND ls_item_id = :item_id`, `preassign.bindValue(":item_id", _item->id());`   ?

ChristopherCotnoir: Yes, that works.

guiclient/enterPoReceipt.cpp
jcarlin: done https://github.com/xtuple/xtuple/pull/2961/files#diff-b9de413efe9b58ca0cd086d5e1b686c9R74

guiclient/createLotSerial.cpp
```diff
@@ -102,18 +102,17 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
+      preassign.prepare("SELECT MAX(lsdetail_id) AS lsdetail_id, ls_number "
+                        "FROM lsdetail "
+                        " JOIN ls ON ls_id = lsdetail_ls_id "
+                        "WHERE lsdetail_source_number = :docnumber "
+                        " AND lsdetail_source_type = :transtype "
+                        " AND ls_item_id = :item_id "
+                        " AND lsdetail_qtytoassign > 0 "
+                        "GROUP BY ls_number;");
       preassign.bindValue(":transtype", createet.value("transtype").toString());
       preassign.bindValue(":docnumber", createet.value("invhist_ordnumber").toString());
```
ChristopherCotnoir: The transtype and docnumber bindings are now a problem in ways they weren't for invadjustment. invadjustment inserts with invhist_transtype as 'AD', which will never be a lsdetail_source_type. As such, having invhist_ordnumber be NULL was not a problem, as this query would not be returning any rows anyway. This is no longer the case with this batch. For example, with enterPoReceipt, when posting receipts for a RA, the invhist_transtype would be 'RR', which can be an lsdetail_source_type. This means having invhist_ordnumber be NULL since it is not yet created poses a problem, because this query will no longer find the lsdetail entries it is supposed to. Additionally, it is no longer valid to assume that both what was previously invhist_transtype and invhist_ordtype are the same as what is now itemlocdist_order_type. This is true for invhist_ordtype, and was true for invhist_transtype for invadjustment, because in that case pTransType and pOrderType passed to postInvTrans were both 'AD'. Now they can differ, a RA in enterPoReceipt will have an invhist_transtype of 'RR', but an invhist_ordtype and itemlocdist_order_type of 'RA', so the previous query will fetch 'RA' for transtype when it should fetch 'RR'. This query is important because it sets _preassigned = true and forces the user to pick one of the rows returned by this query. This matters because it means when they call createitemlocdist, the _preassigned=true query will be used, resulting in the lsdetail_qtyassigned being adjusted (as well as other operations for lsdetail_source_type='RR'). This is counteracted by deleteitemlocdist if the user then deletes the itemlocdist entry from the assignLotSerial window. If one of these happens but not the other, lsdetail_qtyassigned will be modified incorrectly. Additionally, there is a problem with deleteitemlocdist itself in that it also relies on invhist already existing to find the lsdetail to revert. Again, this wasn't a problem before because lsdetail_source_type was never 'AD', but becomes a problem now. Something has to be done to keep the lsdetail changes in createlotserial and deleteitemlocdist working properly.

jcarlin: One option may be to move all the `INSERT INTO lsdetail`s to occur before the call to `distributeInventory::SeriesAdjust` for each relevant C++ file: 
- issueLineToShipping.cpp
- issueToShipping.cpp
- issueWoMaterialBatch.cpp
- issueWoMaterialItem.cpp

and extract it out of `issuewortnreceipt.sql` (private-extensions). I think it would be possible to write out the conditions and have these inserts occur in `createItemlocdistParent.sql`.

ChristopherCotnoir: The timing of the inserts isn't the issue, I believe that is correct as is and are in that order for a reason. The issue is that this window can't find the records with these changes. This window is the only place in the code that actually uses lsdetail_qtytoassign (that is, the only place to SELECT it, every other use of it is merely inserting it or updating it, which doesn't present anything to the user), so the entire reason that column was inserted must have been for use here. In fact, most of lsdetail is only used (SELECTed) here, so if this window can't find the lsdetail entries, they are practically useless.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+    }
+    else
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+        parentSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+    if (qi.value("controlled").toBool())
```
ChristopherCotnoir: If the 'to' itemsite is not controlled but the 'from' itemsite is, this will skip the 'from' itemlocdist entry. Also, the interwarehousetransfer code creates itemlocdist entries even if itemsite_controlmethod='N', but the iscontrolleditemsite function returns false in that case. I talked with Gil about the second issue, and he said to change postinvtrans so that it works the same way as interwarehousetransfer with regards to that case.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+      if (qi.value("recv_order_type").toString() == "TO" && _metrics->boolean("MultiWhs"))
+      {
+        XSqlQuery tohead;
+        tohead.prepare("SELECT f.itemsite_id AS from_itemsite_id "
+                       "FROM toitem "
+                       "  JOIN tohead ON toitem_tohead_id = tohead_id "
+                       "  JOIN itemsite AS f ON toitem_item_id = f.itemsite_item_id " 
+                       "    AND f.itemsite_warehous_id = tohead_trns_warehous_id "
+                       "WHERE toitem_id = :recv_orderitem_id;");
+        tohead.bindValue(":recv_orderitem_id", qi.value("recv_orderitem_id").toInt());
```
ChristopherCotnoir: recv_orderitem_id needs to be selected above.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+        tohead.exec();
+        if (tohead.first())
+        {
+          parentItemlocdist.bindValue(":itemsite_id", tohead.value("from_itemsite_id").toInt());
+          parentItemlocdist.bindValue(":qty", qi.value("qty").toDouble() * -1);
```
ChristopherCotnoir: Should just be recv_qty in this case (multiplied by -1), not the expression that is 'qty'.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+        parentSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+    if (qi.value("controlled").toBool())
+    {
+      int itemlocdistId;
```
ChristopherCotnoir: Should be initialized to -1 to ensure it isn't bound if not a transfer order.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+          ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving itemsite For Transfer Order"),
+                             tohead, __FILE__, __LINE__);
+          return; 
+        }
+      }
+
+      if (itemlocdistId > 0)
+        parentItemlocdist.bindValue(":itemlocdistId", itemlocdistId);
+      parentItemlocdist.bindValue(":itemsite_id", qi.value("itemsite_id").toInt());
+      parentItemlocdist.bindValue(":qty", qi.value("qty").toDouble());
```
ChristopherCotnoir: Should be recv_qty again.

guiclient/createLotSerial.cpp
jcarlin: You're right. For some reason I was thinking it needed to be here for the first transaction, whereas it's only for previous inventory transaction.

guiclient/enterPoReceipt.cpp
jcarlin: Change postInvTrans, so remove https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/functions/postinvtrans.sql#L60?

In that case, shouldn't `isControlledItemsite` just drop the `itemsite_controlmethod != 'N'` condition all together?

ChristopherCotnoir: Yes, remove that check but still skip the invhist insertion if the control method is 'N'. And yes, remove it from isControlledItemsite.

ChristopherCotnoir: Actually, hold off on that. I talked with Gil, we'll come back to this later. For now, assume control method of 'N' works like before, nothing happens. But this check still needs to take into account the from and to site not both having the same control.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +295,117 @@ void enterPoReceipt::sPost()
...
+      parentItemlocdist.exec();
```
ChristopherCotnoir: Should not be done if recv_order_type is 'TO' but MultiWhs metric is false, because interWarehouseTransfer will not be performed in that case.

guiclient/enterPoReturn.cpp
```diff
@@ -171,19 +168,80 @@ void enterPoReturn::sPost()
...
+                          " LEFT OUTER JOIN recv ON (recv_id=poreject_recv_id) "
+                          "WHERE pohead_id = :poheadId "
+                          " AND isControlledItemsite(itemsite_id) "
+                          "GROUP BY poreject_id, pohead_number, itemsite_id;");
+  controlledItems.bindValue(":poheadId", _po->id());
+  controlledItems.exec();
+  while (controlledItems.next())
+  {
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'RP', "
```
ChristopherCotnoir: 'RP' should be 'PO'.

guiclient/enterPoReturn.cpp
```diff
@@ -171,19 +168,80 @@ void enterPoReturn::sPost()
...
+  else
+  {
+    QMessageBox::information(this, tr("Enter Receipt"),
+                               tr("Failed to Retrieve itemloc_series_seq") );
+    return;
+  }
+  
+  // Sql from postPoReturns here because itemlocdist 'parent' records need to be created here for each, post #22868
+  XSqlQuery controlledItems;
+  controlledItems.prepare("SELECT itemsite_id, SUM(poreject_qty) AS qty, pohead_number "
```
ChristopherCotnoir: SUM(poreject_qty) should be SUM(poreject_qty) * poitem_invvenduomratio * -1.

guiclient/enterPoReturn.cpp
```diff
@@ -171,19 +168,80 @@ void enterPoReturn::sPost()
...
+    parentItemlocdist.bindValue(":poheadNumber", controlledItems.value("pohead_number").toString());
+    parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+        QDate(), true) == XDialog::Rejected)
+      {
+        cleanup.exec();
+        QMessageBox::information( this, tr("Enter Receipts"), tr("Posting Distribution Detail Failed") );
```
ChristopherCotnoir: "Enter PO Return" instead of "Enter Receipts".

guiclient/enterPoReturn.cpp
```diff
@@ -171,19 +168,80 @@ void enterPoReturn::sPost()
...
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'RP', "
+                              " :poheadNumber, :itemlocSeries) AS result;");
+    parentItemlocdist.bindValue(":itemsite_id", controlledItems.value("itemsite_id").toInt());
+    parentItemlocdist.bindValue(":qty", controlledItems.value("qty").toDouble());
+    parentItemlocdist.bindValue(":poheadNumber", controlledItems.value("pohead_number").toString());
+    parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
```
ChristopherCotnoir: Should be called once after the loop, not inside it.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,136 @@ void enterPoReceipt::sPost()
...
+        parentItemlocdist.bindValue(":orderNumber", qi.value("recv_order_number").toString());
+        parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+        parentItemlocdist.exec();
+        if (parentItemlocdist.first())
+        {
+          // If the "to" itemsite is not controlled, distribute here
+          if (!qi.value("controlled").toBool())
+          {
+            if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+              QDate(), true) == XDialog::Rejected)
```
ChristopherCotnoir: s/QString(), QDate(), QDate()/lotnum, expdate, warrdate/

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,136 @@ void enterPoReceipt::sPost()
...
+            "'From' itemsite"), parentItemlocdist, __FILE__, __LINE__);
+          return;
+        }
+      }
+      else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving 'From' itemsite"),
+        tohead, __FILE__, __LINE__))
+        return;
+    }
+    
+    if (qi.value("controlled").toBool())
```
ChristopherCotnoir: && (qi.value("recv_order_type").toString() != "TO" || _metrics->boolean("MultiWhs"))

To ensure that this section isn't performed if the order is a Transfer Order but MultiWhs is false, because interwarehousetransfer will be skipped in that case.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,136 @@ void enterPoReceipt::sPost()
...
+                     "  AND warehous_transit=FALSE "
+                     "  AND isControlledItemsite(itemsite_id);");
+      tohead.bindValue(":toitem_id", qi.value("recv_orderitem_id").toInt());
+      tohead.exec();
+      if (tohead.first())
+      {
+        parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, '', '', "
+          ":itemlocSeries, NULL, :itemlocdistId) AS result;");
+        parentItemlocdist.bindValue(":itemsite_id", tohead.value("itemsite_id").toInt());
+        parentItemlocdist.bindValue(":qty", qi.value("recv_qty").toDouble());
```
ChristopherCotnoir: Should be multiplied by -1 here.

guiclient/enterPoitemReceipt.cpp
```diff
@@ -323,34 +319,92 @@ void enterPoitemReceipt::sReceive()
...
+
+  // Stage cleanup function to be called on error
+  XSqlQuery cleanup;
+  cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+
   if (_mode == cNew)
   {
     enterReceive.prepare("SELECT enterReceipt(:ordertype, :poitem_id, :qty, :freight, :notes, "
-              ":curr_id, :effective, :purchcost) AS result;");
+              ":curr_id, :effective, :purchcost) AS itemlocSeries;");
```
ChristopherCotnoir: Changing the alias of the result prevents it from being accessed below, change it back.

guiclient/enterPoitemReceipt.cpp
```diff
@@ -323,34 +319,92 @@ void enterPoitemReceipt::sReceive()
...
     enterReceive.bindValue(":ordertype",	_ordertype);
     storedProc = "enterReceipt";
   }
   else if (_mode == cEdit)
   {
-    enterReceive.exec("BEGIN;");	// because of possible lot, serial, or location distribution cancelations
+    // Get parent series id
+    XSqlQuery parentSeries;
+    parentSeries.prepare("SELECT NEXTVAL('itemloc_series_seq') AS result, "
+                         "  isControlledItemsite(:itemsite_id) AS controlled;");
```
ChristopherCotnoir: This will raise an exception if _itemsiteId is -1. Seeing as correctreceipt has code for that case specifically, we don't want that.

guiclient/enterPoitemReceipt.cpp
```diff
@@ -323,34 +319,92 @@ void enterPoitemReceipt::sReceive()
...
+    }
+    else
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+                              parentSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+    if (parentSeries.value("controlled").toBool())
```
ChristopherCotnoir: correctreceipt runs without error, but does not call postInvTrans (meaning we don't want to create an itemlocdist entry) if the receipt is not posted (recv_posted is false), _toReceive->toDouble()==_received->toDouble(), or _itemsiteId is -1. Check for these cases and skip itemlocdist creation and distribution in those cases.

guiclient/enterPoitemReceipt.cpp
```diff
@@ -323,34 +319,92 @@ void enterPoitemReceipt::sReceive()
...
+    {
+      XSqlQuery parentItemlocdist;
+      parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, :orderType, "
+                                " :orderNumber, :itemlocSeries);");
+      parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
+      parentItemlocdist.bindValue(":qty", _toReceive->toDouble());
```
ChristopherCotnoir: Should be (_toReceive->toDouble()-_received->toDouble())*_invVendorUOMRatio->toDouble().

guiclient/enterPoitemReceipt.cpp
```diff
@@ -323,34 +319,92 @@ void enterPoitemReceipt::sReceive()
...
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Records"),\
+          parentItemlocdist, __FILE__, __LINE__);
+        return;
+      }
+    }
+
     enterReceive.prepare("UPDATE recv SET recv_notes = :notes WHERE (recv_id=:recv_id);" );
     enterReceive.bindValue(":notes",	_notes->toPlainText());
     enterReceive.bindValue(":recv_id",	_recvid);
     enterReceive.exec();
```
ChristopherCotnoir: This statement isn't undone if correctreceipt fails.

guiclient/expenseTrans.cpp
```diff
@@ -142,40 +143,80 @@ void expenseTrans::sPost()
   if (GuiErrorCheck::reportErrors(this, tr("Cannot Post Transaction"), errors))
     return;
 
-  XSqlQuery rollback;
-  rollback.prepare("ROLLBACK;");
+  // Stage cleanup function to be called on error
+  XSqlQuery cleanup;
+  cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+  cleanup.bindValue(":itemlocSeries", itemlocSeries);
```
ChristopherCotnoir: Don't bindValue here.

guiclient/expenseTrans.cpp
```diff
@@ -142,40 +143,80 @@ void expenseTrans::sPost()
...
+    return;
+  }
+
+  // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+  if (_controlledItem)
+  {
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'EX'::TEXT, :docNumber, :itemlocSeries) AS result;");
+    parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
+    parentItemlocdist.bindValue(":qty", _qty->toDouble());
```
ChristopherCotnoir: Multiply by -1, because the sense is negative.

guiclient/invoice.cpp
```diff
@@ -805,76 +807,134 @@ void invoice::postInvoice()
...
+  items.bindValue(":invchead_id", _invcheadid);
+  items.exec();
+  while (items.next())
+  {
+    // Create the parent itemlocdist record for each line item requiring distribution, call distributeInventory::seriesAdjust
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IN', "
+                              " :orderNumber, :itemlocSeries);");
+    parentItemlocdist.bindValue(":itemsite_id", items.value("itemsite_id").toInt());
+    parentItemlocdist.bindValue(":qty", items.value("qty").toDouble());
```
ChristopherCotnoir: Should be * -1.

guiclient/invoice.cpp
```diff
@@ -805,76 +807,134 @@ void invoice::postInvoice()
...
+    parentItemlocdist.bindValue(":orderNumber", items.value("invchead_invcnumber").toString());
+    parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+        QDate(), true) == XDialog::Rejected)
```
ChristopherCotnoir: Should be called after the loop, not inside it.

guiclient/invoice.cpp
```diff
@@ -805,76 +807,134 @@ void invoice::postInvoice()
...
+                          __FILE__, __LINE__);
+      return;
+    }
+  }
+  // contains() string is hard-coded in stored procedure
+  else if (post.lastError().databaseText().contains("post to closed period"))
+  {
+    cleanup.exec();
+    return;
+  }
```
ChristopherCotnoir: Instead cleanup if there is any error, not just that one.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -174,29 +174,80 @@ void issueWoMaterialBatch::sIssue()
...
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+        parentSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    if (items.value("controlled").toBool())
+    {
+      // Create the parent itemlocdist record for each line item requiring distribution, call distributeInventory::seriesAdjust
+      XSqlQuery parentItemlocdist;
+      parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IM', "
```
ChristopherCotnoir: 'IM' should be 'WO'.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -174,29 +174,80 @@ void issueWoMaterialBatch::sIssue()
...
+                                " :orderNumber, :itemlocSeries);");
+      parentItemlocdist.bindValue(":itemsite_id", items.value("itemsite_id").toInt());
+      parentItemlocdist.bindValue(":qty", items.value("qty").toDouble());
```
ChristopherCotnoir: Should be selected above with a different alias as roundQty(item_fractional, itemuomtouom(item_id, womatl_uom_id, NULL, roundQty(itemuomfractionalbyuom(item_id, womatl_uom_id), noNeg(CASE WHEN (womatl_qtyreq >= 0) THEN womatl_qtyreq - womatl_qtyiss ELSE womatl_qtyiss * -1 END))))*-1

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -174,29 +174,80 @@ void issueWoMaterialBatch::sIssue()
...
+      if (result < 0 || result != itemlocSeries)
       {
         rollback.exec();
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Information; Work Order ID #%1")
+        cleanup.exec();
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Invoice"),
+                          issue, __FILE__, __LINE__);
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Issuing Work Order Material; Work Order ID #%1")
                              .arg(_wo->id()),
                              issue, __FILE__, __LINE__);
```
ChristopherCotnoir: Don't raise two errors.

guiclient/issueWoMaterialItem.cpp
```diff
@@ -170,35 +193,60 @@ void issueWoMaterialItem::sIssue()
           .arg(issueIssue.value("warehous_code").toString()));
       return;
     }
+
+    // Create the parent itemlocdist record for each line item requiring distribution, call distributeInventory::seriesAdjust
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IM', "
```
ChristopherCotnoir: s/IM/WO/

guiclient/issueWoMaterialItem.cpp
```diff
@@ -170,35 +193,60 @@ void issueWoMaterialItem::sIssue()
...
+                              " :orderNumber, :itemlocSeries);");
+    parentItemlocdist.bindValue(":itemsite_id", issueIssue.value("itemsite_id").toInt());
+    parentItemlocdist.bindValue(":qty", _qtyToIssue->toDouble());
```
ChristopherCotnoir: Should be roundQty(item_fractional, itemuomtouom(itemsite_item_id, womatl_uom_id, NULL, :qty))

guiclient/postInvoices.cpp
```diff
@@ -115,83 +116,174 @@ void postInvoices::sPost()
...
     }
-    else if (distributeInventory::SeriesAdjust(result, this) == XDialog::Rejected)
+    else
     {
-      rollback.exec();
-      QMessageBox::information( this, tr("Post Invoices"), tr("Transaction Canceled") );
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+                              parentSeries, __FILE__, __LINE__);
       return;
     }
```
ChristopherCotnoir: The above and the SeriesAdjust calls should both be outside the loop, there should only be one itemlocSeries shared by everything.

guiclient/postInvoices.cpp
```diff
@@ -115,83 +116,174 @@ void postInvoices::sPost()
...
-      params.append("startJrnlnum", journalNumber);
-      params.append("endJrnlnum", journalNumber);
-
-      if (_metrics->boolean("UseJournals"))
+      // Create the parent itemlocdist record for each line item requiring distribution, call distributeInventory::seriesAdjust
+      XSqlQuery parentItemlocdist;
+      parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IN', "
+                                " :orderNumber, :itemlocSeries);");
+      parentItemlocdist.bindValue(":itemsite_id", items.value("itemsite_id").toInt());
+      parentItemlocdist.bindValue(":qty", items.value("qty").toDouble());
```
ChristopherCotnoir: Should be * -1.

guiclient/returnWoMaterialItem.cpp
```diff
@@ -101,39 +102,78 @@ void returnWoMaterialItem::sReturn()
...
+    ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+      parentSeries, __FILE__, __LINE__);
+    return;
+  }
 
+  // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+  if (_controlledItem)
+  {
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IM', "
```
ChristopherCotnoir: s/IM/WO/

guiclient/returnWoMaterialItem.cpp
```diff
@@ -101,39 +102,78 @@ void returnWoMaterialItem::sReturn()
...
+                              " formatWoNumber(:woId), :itemlocSeries) AS result;");
+    parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
+    parentItemlocdist.bindValue(":qty", _qty->toDouble());
```
ChristopherCotnoir: Should be itemuomtouom(itemsite_item_id, womatl_uom_id, NULL, pQty) where pQty is _qty->toDouble().

guiclient/returnWoMaterialItem.cpp
```diff
@@ -101,39 +102,78 @@ void returnWoMaterialItem::sReturn()
...
+    parentItemlocdist.bindValue(":woId", _wo->id());
+    parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+        QDate(), true) == XDialog::Rejected)
+      {
+        cleanup.exec();
+        QMessageBox::information(this, tr("Enter Receipt"),
```
ChristopherCotnoir: Material Return instead of Enter Receipt.

guiclient/returnWoMaterialItem.cpp
```diff
@@ -172,16 +212,20 @@ void returnWoMaterialItem::sSetQOH(int pWomatlid)
...
+                 "WHERE womatl_itemsite_id = itemsite_id "
+                 "  AND womatl_uom_id = uom_id "
+                 "  AND womatl_id=:womatl_id;" );
     qoh.bindValue(":womatl_id", pWomatlid);
     qoh.exec();
     if (qoh.first())
     {
+      _itemsiteId = qoh.value("itemsite_id").toInt();
+      _controlledItem = qoh.value("controlled").toBool();
+      _uom->setText(qoh.value("uom_name").toString());
```
ChristopherCotnoir: No need to duplicate this line.

guiclient/scrapTrans.cpp
```diff
@@ -151,51 +151,77 @@ void scrapTrans::sPost()
...
+    return;
+  }
+
+  // If controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+  if (_controlledItem)
+  {
+    XSqlQuery parentItemlocdist;
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'SI'::TEXT, :docNumber, :itemlocSeries) AS result;");
+    parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
+    parentItemlocdist.bindValue(":qty", _qty->toDouble());
```
ChristopherCotnoir: Should be multiplied by -1.

guiclient/scrapTrans.cpp
```diff
@@ -151,51 +151,77 @@ void scrapTrans::sPost()
...
+    parentItemlocdist.bindValue(":docNumber", _documentNum->text());
+    parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+        QDate(), true) == XDialog::Rejected)
+      {
+        cleanup.exec();
+        QMessageBox::information(this, tr("Enter Receipt"),
```
ChristopherCotnoir: Enter Receipt should be Scrap Transaction.

guiclient/scrapTrans.cpp
```diff
@@ -151,51 +151,77 @@ void scrapTrans::sPost()
...
-                               __FILE__, __LINE__);
-      return;
-    }
-    else if (scrapPost.lastError().type() != QSqlError::NoError)
+    if (result < 0 || result != itemlocSeries)
     {
-      rollback.exec();
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Scrap Transaction"),
+      cleanup.exec();
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Work Order Information"),
```
ChristopherCotnoir: The previous error title was fine as it was.

guiclient/distributeInventory.cpp
```diff
@@ -379,11 +379,11 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
     if (!pPreDistributed)
     {
       XSqlQuery postDistDetail;
-      postDistDetail.prepare("SELECT postdistdetail(:itemlocSeries, NULL::INTEGER) AS result;");
+      postDistDetail.prepare("SELECT postdistdetail(:itemlocSeries) AS result;");
       postDistDetail.bindValue(":itemlocSeries", pItemlocSeries);
       postDistDetail.exec();
       if (!postDistDetail.first() || ErrorReporter::error(QtCriticalMsg, 0, tr("Distribution Detail Posting Failed"),
-                                postDistDetail, __FILE__, __LINE__))
+                                postDistDetail, __FILE__, __LINE__) || postDistDetail.value("result") <= 0)
```
ChristopherCotnoir: The result can be 0 if not controlled, in which case this is still called to call postitemlocseries.

guiclient/invoice.cpp
```diff
@@ -910,6 +900,15 @@ void invoice::postInvoice()
     }
   }
 
+  // If the above query had results, call distributeInventory
+  if (items.numRowsAffected() > 0 && (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
```
ChristopherCotnoir: numRowsAffected is undefined for SELECT statements, use size() instead.

guiclient/postInvoices.cpp
```diff
@@ -226,6 +215,14 @@ void postInvoices::sPost()
       }
     }
 
+    if (items.numRowsAffected() > 0 && distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(), true)
```
ChristopherCotnoir: Use size() instead of numRowsAffected().

guiclient/returnWoMaterialItem.cpp
```diff
@@ -126,8 +126,12 @@ void returnWoMaterialItem::sReturn()
...
   {
     XSqlQuery parentItemlocdist;
-    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, :qty, 'IM', "
-                              " formatWoNumber(:woId), :itemlocSeries) AS result;");
+    parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, "
+                              " itemuomtouom(itemsite_item_id, womatl_uom_id, NULL, :qty), "
+                              " 'WO', formatWoNumber(:woId), :itemlocSeries) AS result "
+                              "FROM womatl "
+                              " JOIN itemsite ON womatl_itemsite_id = itemsite_id "
+                              "WHERE womatl_id = :womatl_id;");
```
ChristopherCotnoir: :womatl_id needs to be bound.

guiclient/createLotSerial.cpp
jcarlin: Looking at this today...

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,140 @@ void enterPoReceipt::sPost()
...
+      }
+      else
+      {
+        ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving 'From' itemsite"), tohead, __FILE__, __LINE__);
+        return;
+      }
+    }
+    
+    if (qi.value("controlled").toBool() && 
+      (qi.value("recv_order_type").toString() != "TO" || _metrics->boolean("MultiWhs")))
```
jcarlin: This should be replaced with `if (qi.value("controlled").toBool() && 
      (qi.value("recv_order_type").toString() == "TO" ? _metrics->boolean("MultiWhs") : true))`

guiclient/returnWoMaterialItem.cpp
jcarlin: Done: https://github.com/xtuple/qt-client/pull/1522/files/1712625519ced4a10488c952f507f7fb8edc0b9f#diff-a5f65adb5c8ca25d1e333b0187c7136fR157

guiclient/enterPoReceipt.cpp
jcarlin: Done: https://github.com/xtuple/qt-client/pull/1522/files#diff-684c6b8020341b7c622cf22098dfd4deR387

guiclient/createLotSerial.cpp
jcarlin: @ChristopherCotnoir 
> Additionally, there is a problem with deleteitemlocdist itself in that it also relies on invhist already existing to find the lsdetail to revert.

`deleteItemlocSeries.sql` does not depend on `invhist` for deleting `lsdetail` records: https://github.com/xtuple/xtuple/pull/2961/files#diff-b9de413efe9b58ca0cd086d5e1b686c9R60

jcarlin: The remaining issue is the null `invhist` records and the affect that has on the `transtype` and `ordtype` values as you pointed out.

jcarlin: I believe the creation of the records this code block is looking for only occurs in these 4 files: 

issueLineToShipping.cpp
329:        lsdetail.prepare("INSERT INTO lsdetail "
issueToShipping.cpp
550:      lsdetail.prepare("INSERT INTO lsdetail "
issueWoMaterialBatch.cpp
261:        lsdetail.prepare("INSERT INTO lsdetail "
issueWoMaterialItem.cpp
257:      lsdetail.prepare("INSERT INTO lsdetail "

In that case, being that issueLineToShipping and issueToShipping has not been updated yet for issue #22868, could we just replace `itemlocdist_ordertype` with `CASE WHEN itemlocdist_ordertype = 'WO' THEN 'IM' ELSE itemlocdist_ordertype END AS ordtype`

guiclient/distributeInventory.cpp
```diff
@@ -362,7 +363,7 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
         query.exec();
         if (query.first())
         {
-          if (query.numRowsAffected() != 1)
+          if (query.size() != 1)
```
ChristopherCotnoir: size() returns -1 for non-SELECT queries. Leave it as it was.

guiclient/createLotSerial.cpp
jcarlin: I think the following would work: 
`COALESCE(invhist_transtype, CASE "
                      "   WHEN itemlocdist_order_type = 'WO' AND itemlocdist_qty > 0 THEN 'IM' "
                      "   WHEN itemlocdist_order_type = 'TO' AND itemlocdist_qty > 0 THEN 'TR' "
                      "   ELSE itemlocdist_order_type END) AS transtype`

but `invhist_ordnumber` would need to be replaced. If the `itemlocdist` record has itemlocdist_type and itemlocdist_order_id, then we could attempt to find the record in invhist through orderhead. However, an additional problem is invhist_docnumber is sometimes null, other times formatted with `||line number` - all of that would need to be corrected so that invhist_docnumber is null or the actual number from the table, not a formatted number. 

The entire point here is we're trying to find an invhist record through an itemlocdist_id that has no tie to invhist. The only way to do that through itemlocdist that I can think of is all the above. Or, it's passed as a param but that again is limited through the itemloc query in distributeInventory which also uses itemlocdist.

I'm going to move on and correct the only remaining issue in my testing (so far): receiving a transfer order is not posting inventory detail.

guiclient/createLotSerial.cpp
```diff
@@ -102,24 +110,28 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
+        preassign.bindValue(":invhist_ordnumber", createet.value("invhist_ordnumber").toString());
+      preassign.bindValue(":sourceId", createet.value("itemlocdist_order_id").toInt());
+      preassign.bindValue(":item_id", _item->id());
       preassign.exec();
       if (preassign.first())
       {
         _lotSerial->setAllowNull(true);
         _lotSerial->populate(preassign);
         _preassigned = true;
+        _lotSerial->setEditable(false);
```
jcarlin: IMO this is much better. Why allow the user to enter any lot number only to throw an error later after trying to save? If there are preassigned records, make the user select from the drop down.

guiclient/createLotSerial.cpp
```diff
@@ -409,7 +428,7 @@ void createLotSerial::sAssign()
     createAssign.exec();
     if (createAssign.first())
     {
-      if (!_serial)
+      if (!_serial && !_preassigned)
```
jcarlin: Don't ask the user if it's required to use a preassigned lot number from the drop down list.

guiclient/createLotSerial.cpp
jcarlin: This is working now that the 'order' item id is being inserted into itemlocdist_order_id.

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,140 @@ void enterPoReceipt::sPost()
...
+    }
+
+    // Wrap the remaining sql function calls in a transaction in case of errors
+    enterPost.exec("BEGIN;");
+
+    // Post the inventory transaction(s)
     XSqlQuery postLine;
-    postLine.prepare("SELECT postReceipt(recv_id, 0) AS result "
+    postLine.prepare("SELECT postReceipt(recv_id, :itemlocSeries, true) AS result "
                      "FROM recv "
```
jcarlin: Unnecessary from or good to check that it still exists?

guiclient/enterPoReceipt.cpp
```diff
@@ -287,30 +296,140 @@ void enterPoReceipt::sPost()
...
+    {
+      XSqlQuery tohead;
+      tohead.prepare("SELECT itemsite_id "
+                     "FROM toitem "
+                     "  JOIN tohead ON toitem_tohead_id = tohead_id "
+                     "  JOIN itemsite ON toitem_item_id = itemsite_item_id "
+                     "    AND itemsite_warehous_id = tohead_trns_warehous_id " // from wh
+                     "  JOIN whsinfo ON itemsite_warehous_id = warehous_id "
+                     "WHERE toitem_id = :toitem_id "
+                     "  AND warehous_transit=FALSE "
```
jcarlin: Something's not right here... above joining itemsite on tohead_trns_warehous_id and now filtering out?

guiclient/enterPoReceipt.cpp
gilmoskowitz: The `FROM` is unnecessary and might hide bugs because we expect the `recv` record to exist here.

guiclient/enterPoReceipt.cpp
jcarlin: Aha! This entire from block can be removed because `postReceipt`, which calls interwarehouseTransfer always sent the the transit warehouse (`pFromWarehouseId`) https://github.com/xtuple/private-extensions/blob/4_10_x/source/inventory/foundation-database/public/functions/interwarehousetransfer.sql#L140-L145

guiclient/enterPoitemReceipt.cpp
```diff
@@ -234,12 +234,13 @@ void enterPoitemReceipt::populate()
     else if (_ordertype == "RA")
       _orderType->setText(tr("R/A"));
 
-    int itemsiteid = enterpopulate.value("itemsiteid").toInt();
+    _itemsiteId = enterpopulate.value("itemsiteid").toInt();
+    _recvPosted = enterpopulate.value("recv_posted").toBool();
```
jcarlin: only available if `_mode == cEdit`

guiclient/createLotSerial.cpp
```diff
@@ -72,12 +72,15 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
+                      " itemlocdist_order_id, "
                       // When #22868 is complete, the attempt to use invhist will be removed and replaced with something (hopefully not a case statement) 
                       // that can match the types that were inserted in creation of the original lsdetail records.
                       // Example of the lsdetail record creation that we're now looking for in this query: 
                       // https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/issueWoMaterialItem.cpp#L211
-                      " COALESCE(invhist_transtype, CASE WHEN itemlocdist_order_type = 'WO' "
-                      "   THEN 'IM' ELSE itemlocdist_order_type END) AS transtype, "
+                      " COALESCE(invhist_transtype, CASE "
+                      "   WHEN itemlocdist_order_type = 'WO' THEN 'IM' "
+                      "   WHEN itemlocdist_order_type = 'TO' THEN 'TR' "
```
ChristopherCotnoir: Should also have WHEN itemlocdist_order_type = 'RA' THEN 'RR'.

guiclient/createLotSerial.cpp
ChristopherCotnoir: deleteitemlocseries does not depend on invhist, deleteitemlocdist does.

guiclient/createLotSerial.cpp
jcarlin: I didn't see any place that created preassigned `lsdetail` records for Return Authorizations.

guiclient/enterPoReceipt.cpp
```diff
@@ -311,78 +311,19 @@ void enterPoReceipt::sPost()
...
-        {
-          cleanup.exec();
-          ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Record for "
-            "'From' itemsite"), parentItemlocdist, __FILE__, __LINE__);
-          return;
-        }
-      }
-      else if (tohead.lastError().type() != QSqlError::NoError)
-        return;
-    }
```
ChristopherCotnoir: Why is this being removed? It is still needed.

guiclient/createLotSerial.cpp
ChristopherCotnoir: The raitemls trigger in private-extensions.

guiclient/createLotSerial.cpp
jcarlin: `deleteItemlocDist` is only called from `assignLotSerial::sDelete();` I don't see any issue with calling it if there is no invhist records yet, the function will continue to delete the appropriate records https://github.com/xtuple/private-extensions/blob/4_11_x/source/inventory/foundation-database/public/functions/deleteitemlocdist.sql#L38 & L32

guiclient/enterPoReceipt.cpp
jcarlin: Because, when receiving, if the From site is a transit site there is no distribution required. See how https://github.com/xtuple/private-extensions/blob/4_10_x/source/inventory/foundation-database/public/functions/interwarehousetransfer.sql#L141 filters on NOT `warehous_transit`...

ChristopherCotnoir: Ah. Bring it back anyway for now. We are considering removing the warehous_transit check from interwarehousetransfer, along with dealing with other inconsistencies between interwarehousetransfer and postinvtrans, but Gil and I decided to do that last.

guiclient/createLotSerial.cpp
ChristopherCotnoir: L28-30 is the issue. It is meant to reverse L81-83 in createlotserial.sql, but that won't happen if it can't find the lsdetail record, leaving the createlotserial changes intact.

guiclient/distributeInventory.cpp
```diff
@@ -115,7 +113,7 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
     XSqlQuery itemloc;
```
ChristopherCotnoir: The below query has usages of invhist, and the tactic used to deal with them for invAdjust is no longer sufficient as there are now transactions with transtypes of RP, RR, RX, etc being refactored.

guiclient/distributeInventory.cpp
```diff
@@ -369,11 +379,17 @@ int distributeInventory::SeriesAdjust(int pItemlocSeries, QWidget *pParent,
...
+      postDistDetail.prepare("SELECT postdistdetail(:itemlocSeries) AS result;");
       postDistDetail.bindValue(":itemlocSeries", pItemlocSeries);
       postDistDetail.exec();
-      if (!postDistDetail.first() || ErrorReporter::error(QtCriticalMsg, 0, tr("Distribution Detail Posting Failed"),
-                                postDistDetail, __FILE__, __LINE__))
+
+      // Return error for any of the following 3 cases
+      if (!postDistDetail.first() 
+        || ErrorReporter::error(QtCriticalMsg, 0, tr("Distribution Detail Posting Failed"),
+          postDistDetail, __FILE__, __LINE__)
```
ChristopherCotnoir: This will prevent ErrorReporter from reporting raised exceptions to the user because !postDistDetail.first() will evaluate to true causing the subsequent statements to be skipped due to c++ using short-circuit evaluation.

guiclient/enterPoReceipt.cpp
```diff
@@ -365,11 +366,8 @@ void enterPoReceipt::sPost()
           return;
         }
       }
-      else
-      {
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving 'From' itemsite"), tohead, __FILE__, __LINE__);
+      else if (tohead.lastError().type() != QSqlError::NoError)
         return;
-      }
```
ChristopherCotnoir: Canceling without telling the user seems to me to be a bad idea. This change should probably be undone.

guiclient/enterPoReturn.cpp
```diff
@@ -171,18 +168,88 @@ void enterPoReturn::sPost()
...
+  enterPost.bindValue(":itemlocSeries", itemlocSeries);
   enterPost.exec();
   int result = 0;
   if (enterPost.first())
   {
     result = enterPost.value("result").toInt();
-    if (result < 0)
+    if (result < 0 || result != itemlocSeries)
     {
+      cleanup.exec();
```
jcarlin: Should be after the rollback()

guiclient/invoice.cpp
```diff
@@ -805,76 +807,140 @@ void invoice::postInvoice()
...
-     unpostedPost.exec("COMMIT;");
-   }
-   // contains() string is hard-coded in stored procedure
-   else if (post.lastError().databaseText().contains("post to closed period"))
-     rollback.exec();
-
+    int result = post.value("result").toInt();
+    if (result < 0 || result != itemlocSeries)
+    {
+      cleanup.exec();
```
jcarlin: These should be after the rollbacks.

guiclient/postInvoices.cpp
```diff
@@ -115,83 +116,179 @@ void postInvoices::sPost()
...
+    post.bindValue(":invchead_id", invoiceIds.at(i));
+    post.bindValue(":journal", journalNumber);
+    post.bindValue(":itemlocSeries", itemlocSeries);
+    post.exec();
+    if (post.first())
+    {
+      int result = post.value("result").toInt();
+      if (result < 0 || result != itemlocSeries)
+      {
+        cleanup.exec();
```
jcarlin: These should be after the rollbacks.

guiclient/distributeInventory.cpp
jcarlin: A new column `itemlocdist_invhist_transtype`? I can't reproduce the same results with a itemlocdist case statement. This column would be available for `createLotSerial.cpp` as well. Any other suggestion?

ChristopherCotnoir: Yes, I think that would be best. I don't really see any other way around it.

guiclient/enterPoReceipt.cpp
```diff
@@ -387,12 +387,18 @@ void enterPoReceipt::sPost()
       if (qi.value("recv_order_type").toString() == "TO")
       {
         parentItemlocdist.bindValue(":qty", qi.value("recv_qty").toDouble());
-        parentItemlocdist.bindValue(":transType", 'TR');
+        parentItemlocdist.bindValue(":transType", "TR");
       }
       else if (qi.value("recv_order_type").toString() == "RA")
-        parentItemlocdist.bindValue(":transType", 'RR');
+      {
+        parentItemlocdist.bindValue(":qty", qi.value("qty").toDouble());
```
jcarlin: Commit message should say "... and, add missing qty bindings"!

guiclient/createLotSerial.cpp
```diff
@@ -102,24 +107,28 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
-                        "                              WHERE itemsite_item_id = (SELECT itemsite_item_id "
-                        "                               FROM itemsite WHERE itemsite_id = :itemsite))) "
-                        "AND (lsdetail_qtytoassign > 0) ) "
-                        "GROUP BY 2,3");
+      preassign.prepare("SELECT MAX(lsdetail_id) AS lsdetail_id, ls_number "
+                        "FROM lsdetail "
+                        " JOIN ls ON ls_id = lsdetail_ls_id "
+                        "WHERE "
+                        " CASE WHEN :invhist_ordnumber IS NOT NULL THEN lsdetail_source_number = :invhist_ordnumber "
+                        "   ELSE lsdetail_source_id = :sourceId END "
```
ChristopherCotnoir: This doesn't always work. For example, when lsdetail_source_type='RR', lsdetail_source_id points to a raitem_id, and lsdetail_source_number is (SELECT rahead_number::TEXT || '-' || raitem_linenumber FROM raitem JOIN rahead ON raitem_rahead_id=rahead_id WHERE raitem_id=lsdetail_source_id). This number can match more than one raitem, if they have the same raitem_rahead_id and raitem_linenumber but a different raitem_subnumber. So previously, this would match the raitem_id that corresponds to the lsdetail_source_id as well as any raitem_id with the same raitem_rahead_id and raitem_linenumber but different raitem_subnumber, but now it only matches the former.

I suspect this may also be the case with other lsdetail_source_types, but I haven't gotten that far yet.

ChristopherCotnoir: Also seems to be the case with 'IM', as the lsdetail_source_id seems to come from a Work Order Material but the lsdetail_source_number seems to come from the Work Order, which can have multiple Materials.

jcarlin:
I originally went down that route to use doc numbers but had not realized
specifying more with orderitem_id would exclude records. I see the issue
now. Invhist_docnumber and invhist_ordnumber (invhist_order_number?) are
confusing as well and should be cleaned up which threw me off. But for now
I will look into replicating the same order number formatting here in a
case statement.

guiclient/createLotSerial.cpp
jcarlin: I originally went down that route to use doc numbers but had not realized
specifying more with orderitem_id would exclude records. I see the issue
now. Invhist_docnumber and invhist_ordnumber (invhist_order_number?) are
confusing as well and should be cleaned up which threw me off. But for now
I will look into replicating the same order number formatting here in a
case statement.

jcarlin: Will this work?

    createet.prepare(SELECT item_fractional, itemsite_controlmethod, itemsite_item_id, 
                       itemsite_id, itemsite_perishable, itemsite_warrpurc, 
                       COALESCE(itemsite_lsseq_id,-1) AS itemsite_lsseq_id, 
                       COALESCE(invhist_ordnumber, CASE 
                         WHEN orderhead_id IS NULL AND itemlocdist_order_type = 'WO' THEN formatWoNumber(womatl_wo_id) 
                         WHEN orderhead_id IS NOT NULL AND orderhead_type = 'SO' THEN formatSoNumber(orderitem_id) 
                         WHEN orderhead_id IS NOT NULL AND orderhead_type = 'TO' THEN formatToNumber(orderitem_id) 
                         WHEN orderhead_id IS NOT NULL AND orderhead_type = 'RA' THEN orderhead_number || '-' || orderitem_linenumber 
                         ELSE '' END) AS ordnumber, 
                       itemlocdist_order_id, 
                       COALESCE(invhist_transtype, itemlocdist_transtype) AS transtype, 
                       COALESCE(invhist_ordtype, itemlocdist_order_type) AS ordtype 
                      FROM itemlocdist 
                       JOIN itemsite ON itemlocdist_itemsite_id = itemsite_id 
                       JOIN item ON itemsite_item_id = item_id 
                       LEFT JOIN orderitem ON itemlocdist_order_id = orderitem_id AND itemlocdist_order_type = orderitem_orderhead_type  
                       LEFT JOIN orderhead ON orderitem_orderhead_id = orderhead_id AND orderhead_type = itemlocdist_order_type 
                       LEFT JOIN womatl ON itemlocdist_order_id = womatl_id AND itemlocdist_order_type = 'WO' 
                       LEFT OUTER JOIN invhist ON itemlocdist_invhist_id = invhist_id 
                      WHERE itemlocdist_id = :itemlocdist_id;);

jcarlin: I don't understand the raitemls.sql trigger. In what scenario would the "first" inserted lsdetail record be referenced here? Receiving and posting an RA item and then later receiving and posting more qty?

ChristopherCotnoir: Create a Return Authorization and on the Return Authorization Item screen for a Lot/Serial controlled item. check 'Authorize only these Lot/Serial Numbers', click New and add number(s), and lsdetail records will be created for them. After you save the Return Authorization and authorize the lines, select the Return Authorization on the Enter Order Receipts screen and upon posting when you get to createLotSerial.cpp the query will grab the lsdetail entries.

guiclient/distributeInventory.cpp
```diff
@@ -616,6 +621,7 @@ void distributeInventory::sFillList()
   distributeFillList.prepare( "SELECT itemsite_id, "
              "       COALESCE(itemsite_location_id,-1) AS itemsite_location_id,"
              "       formatlotserialnumber(itemlocdist_ls_id) AS lotserial,"
+             // TODO - this order column which populates the _order text field fails for everything except Sales Orders
              "       (itemlocdist_order_type || ' ' || formatSoItemNumber(itemlocdist_order_id)) AS order,"
```
ChristopherCotnoir: Wrap this in a CASE statement to return this if itemlocdist_order_type='SO' and NULL otherwise to preserve previous behavior now that itemlocdist_order_id can be not NULL when itemlocdist_order_type is not 'SO'.

guiclient/createLotSerial.cpp
ChristopherCotnoir: Agreed.

guiclient/createLotSerial.cpp
```diff
@@ -102,24 +115,28 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
-                        "                              WHERE itemsite_item_id = (SELECT itemsite_item_id "
-                        "                               FROM itemsite WHERE itemsite_id = :itemsite))) "
-                        "AND (lsdetail_qtytoassign > 0) ) "
-                        "GROUP BY 2,3");
+      preassign.prepare("SELECT MAX(lsdetail_id) AS lsdetail_id, ls_number "
+                        "FROM lsdetail "
+                        " JOIN ls ON ls_id = lsdetail_ls_id "
+                        "WHERE "
+                        " CASE WHEN :ordnumber IS NOT NULL THEN lsdetail_source_number = :invhist_ordnumber "
+                        "   ELSE lsdetail_source_id = :sourceId END "
```
ChristopherCotnoir: Should just be lsdetail_source_number=:invhist_ordnumber.

guiclient/createLotSerial.cpp
```diff
@@ -102,24 +115,28 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
+                        " AND lsdetail_source_type = :transtype "
+                        " AND ls_item_id = :item_id "
+                        " AND lsdetail_qtytoassign > 0 "
+                        "GROUP BY ls_number;");
       preassign.bindValue(":transtype", createet.value("transtype").toString());
-      preassign.bindValue(":docnumber", createet.value("invhist_ordnumber").toString());
-      preassign.bindValue(":itemsite", createet.value("itemsite_id").toInt());
+      if (createet.value("invhist_ordnumber").toString().length())
+        preassign.bindValue(":invhist_ordnumber", createet.value("invhist_ordnumber").toString());
+      preassign.bindValue(":sourceId", createet.value("itemlocdist_order_id").toInt());
```
ChristopherCotnoir: Should bind ordnumber, not invhist_ordnumber as :invhist_ordnumber. No need to bind :sourceId anymore.

guiclient/enterPoReceipt.cpp
```diff
@@ -366,8 +366,11 @@ void enterPoReceipt::sPost()
          return;
         }
       }
-      else if (tohead.lastError().type() != QSqlError::NoError)
+      else
```
ChristopherCotnoir: Leave as else if. We only want to throw an error if the function threw an error, if it found no valid records, that just means the source itemlocdist doesn't need to be created (for example, it isn't controlled).

guiclient/createLotSerial.cpp
```diff
@@ -68,18 +68,31 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
                       " itemsite_id, itemsite_perishable, itemsite_warrpurc, "
                       " COALESCE(itemsite_lsseq_id,-1) AS itemsite_lsseq_id, "
-                      " invhist_ordnumber, "
-                      " COALESCE(invhist_transtype, itemlocdist_order_type) AS transtype, "
+                      " COALESCE(invhist_ordnumber, CASE "
+                      "   WHEN orderhead_id IS NULL AND itemlocdist_order_type = 'WO' THEN formatWoNumber(womatl_wo_id) "
+                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'SO' THEN formatSoNumber(orderitem_id) "
+                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'TO' THEN formatToNumber(orderitem_id) "
+                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'RA' THEN orderhead_number || '-' || orderitem_linenumber "
+                      "   ELSE '' END) AS ordnumber, "
```
jcarlin: I think this should include: 
`"   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'PO' THEN orderhead_number "`

ChristopherCotnoir: I don't think that's ever needed because I don't think lsdetail ever inserts anything that would be matched by it. It's certainly not needed in this batch, so save it for later if it does come up.

PR #1530, opened 2017-02-28
Issue 2903: Post Cash Receipt option from inside CR screen

andersdd:
This has been a constant request from customers.

### DIFF
--- a/guiclient/cashReceipt.cpp
+++ b/guiclient/cashReceipt.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
  * version 1.0, the full text of which (including xTuple-specific Exhibits)
  * is available at www.xtuple.com/CPAL.  By using this software, you agree
@@ -21,6 +21,7 @@
 #include "creditCard.h"
 #include "creditcardprocessor.h"
 #include "errorReporter.h"
+#include "getGLDistDate.h"
 #include "mqlutil.h"
 #include "storedProcErrorLookup.h"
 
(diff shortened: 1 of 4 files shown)


guiclient/cashReceipt.cpp
```diff
@@ -973,6 +982,80 @@ bool cashReceipt::save(bool partial)
...
+                  "                    cashrcpt_applydate=CASE WHEN (cashrcpt_applydate < :distdate) THEN :distdate"
+                  "                                            ELSE cashrcpt_applydate END "
+                  "WHERE cashrcpt_id=:cashrcpt_id;");
+   
+  if (changeDate)
+  {
+    setDate.bindValue(":distdate",    newDate);
+    setDate.bindValue(":cashrcpt_id", _cashrcptid);
+    setDate.exec();
+    ErrorReporter::error(QtCriticalMsg, this, tr("Changing Dist. Date"),
```
gilmoskowitz: This can open a dialog waiting for user input during a transaction, which is generally a bad idea. In addition, the transaction tries to continue after a failure instead of rolling back and returning control to the user.

guiclient/cashReceipt.cpp
```diff
@@ -973,6 +982,80 @@ bool cashReceipt::save(bool partial)
...
+  cashPost.exec();
+  if (cashPost.first())
+  {
+    int result = cashPost.value("result").toInt();
+    if (result < 0)
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Cash Receipt Entry"),
+                             storedProcErrorLookup("postCashReceipt", result),
+                             __FILE__, __LINE__);
+      tx.exec("ROLLBACK;");
```
gilmoskowitz: Swap the `rollback` with the `ErrorReporter::error()` call.

guiclient/cashReceipt.cpp
```diff
@@ -779,11 +781,18 @@ void cashReceipt::sSave()
 
   if (save(false))
   {
+    if (_postReceipt->isChecked())
+    {    
+      if (!postReceipt())
+        return;
```
gilmoskowitz: This may try to post the receipt even if it's been posted. The changes elsewhere hide the checkbox, but as an XCheckBox it still has memory of its checked state.

guiclient/cashReceipt.cpp
```diff
@@ -973,6 +982,80 @@ bool cashReceipt::save(bool partial)
...
+    setDate.bindValue(":distdate",    newDate);
+    setDate.bindValue(":cashrcpt_id", _cashrcptid);
+    setDate.exec();
+    ErrorReporter::error(QtCriticalMsg, this, tr("Changing Dist. Date"),
+                         setDate, __FILE__, __LINE__);
+  }
+  
+  cashPost.prepare("SELECT postCashReceipt(:cashrcpt_id, :journalNumber) AS result;");
+  cashPost.bindValue(":cashrcpt_id", _cashrcptid);
+  cashPost.bindValue(":journalNumber", journalNumber);
```
gilmoskowitz: The `journalNumber` variable isn't necessary. The value is set once above and used once here. Just call `fetchJournalNumber()` here.

andersdd:
@gilmoskowitz Updated logic as suggested.  Also went back to original code source and updated that area as well.

PR #1531, opened 2017-02-28
issue #28225 - infrastructure for scriptable widgets

gilmoskowitz:
There were a bunch of unnecessary files in the scriptapi
directory to expose widgets to scripting. That's better
done in the widget sources.

Backwards compatibility is an explicit goal of this commit.


gilmoskowitz:
This needs a lot more testing before it can be merged. The first pass reviews and testing can and should take place in parallel.

widgets/xsqltablemodel.h
```diff
@@ -76,40 +82,100 @@ class XTUPLEWIDGETS_EXPORT XSqlTableModel : public QSqlRelationalTableModel
...
-    virtual QString selectStatement() const;
+    Q_INVOKABLE virtual QSqlDatabase  database() const;
+    Q_INVOKABLE virtual EditStrategy  editStrategy() const;
+    Q_INVOKABLE virtual int           fieldIndex(const QString &fieldName) const;
+    Q_INVOKABLE virtual QString       filter() const;
+    Q_INVOKABLE virtual bool          insertRecord(int row, const QSqlRecord &record);
+    Q_INVOKABLE virtual bool          isDirty(const QModelIndex &index) const;
+    Q_INVOKABLE virtual bool          isDirty() const;
+    Q_INVOKABLE virtual QSqlIndex     primaryKey() const;
+    Q_INVOKABLE virtual QSqlRecord    record() const;
```
bendiy: We should consider exposing the parent object and inheriting from that vs. having to wrap all the parent methods on the child and expose them here. There are a few object like that in this PR.

Here is an example @davidbeauchamp did:
https://github.com/xtuple/qt-client/blob/95e6a0025eb1265aa988ec672af6e323ada82743/scriptapi/qserialportproto.h#L38
https://github.com/xtuple/qt-client/blob/95e6a0025eb1265aa988ec672af6e323ada82743/scriptapi/qiodeviceproto.h#L25

The benefit is the parent object gets exposed and this child object exposure is limited to only it's methods.

widgets/addresscluster.h
```diff
@@ -242,4 +243,8 @@ class XTUPLEWIDGETS_EXPORT AddressCluster : public VirtualCluster
 	QString  _fieldNameCountry;
 };
 
+Q_DECLARE_METATYPE(AddressCluster*)
```
gilmoskowitz: The [Qt docs](http://doc.qt.io/qt-5/qmetatype.html#Q_DECLARE_METATYPE) say
> Some types are registered automatically and do not need this macro:
>
>Pointers to classes derived from QObject

The next commit will strip out `Q_DECLARE_METATYPE(Foo*)` where it appears to be unnecessary in the widgets directory.

bendiy: +1

widgets/xdatawidgetmapper.h
```diff
@@ -12,8 +12,8 @@
 #define XDATAWIDGETMAPPER_H
 
 #include <QDataWidgetMapper>
+#include <QPersistentModelIndex>
```
bendiy: I think this is missing the new `QDataWidgetMapperProto` as the inherited parent. 
```
class XTUPLEWIDGETS_EXPORT XDataWidgetMapper : public QDataWidgetMapperProto
```
That would allow a Qt Script `XDataWidgetMapper` objects to have the parent methods that are now exposed on `QDataWidgetMapperProto`.
It might also be possible to do this in `setupXDataWidgetMapper()` by making the exposed `QDataWidgetMapperProto` `XDataWidgetMapper`'s `prototype`.

Would need to verify this in the Qt Script Debugger with:
```
var mapper = new XDataWidgetMapper();
if (typeof mapper.setRootIndex === 'function') {
  console.log("it worked");
}
```
Where `setRootIndex()` comes from the parent `QDataWidgetMapperProto`.
It might work as is, but in @davidbeauchamp example, we have `class QSerialPortProto : public QIODeviceProto`, I'm not sure this will work as it is now.

gilmoskowitz: I just tested. An XDataWidgetMapper has `setRootIndex()`.

bendiy:
@davidbeauchamp Let us know if you think you'll have time to review this.

PR #1532, opened 2017-03-02
issue #25224 Eliminate duplicated query calls

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#2969

guiclient/salesOrderItem.cpp
```diff
@@ -1141,6 +1145,19 @@ void salesOrderItem::sSave(bool pPartial)
...
+                      "WHERE coitem_id=:soitem_id;");
+    salesSave.bindValue(":soitem_id", _soitemid);
+    salesSave.exec();
+    if (salesSave.first())
+      _supplyOrderId = salesSave.value("coitem_order_id").toInt();
+    else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Item Information"),
+                                  salesSave, __FILE__, __LINE__))
+    {
+      return;
+    }
```
gilmoskowitz: This query can be avoided by `RETURNING coitem_order_id` from the `INSERT`.

guiclient/salesOrderItem.cpp
```diff
@@ -2616,15 +2644,15 @@ void salesOrderItem::sHandleSupplyOrder()
...
+
     double valqty = 0.0;
     ordq.prepare( "SELECT validateOrderQty(itemsite_id, :qty, true) AS qty "
                   "FROM itemsite "
                   "WHERE ((itemsite_item_id=:item_id)"
                   " AND (itemsite_warehous_id=:warehous_id));" );
     ordq.bindValue(":qty", _qtyOrdered->toDouble() * _qtyinvuomratio);
     ordq.bindValue(":item_id", _item->id());
-    ordq.bindValue(":warehous_id", (_item->itemType() == "M") ? _supplyWarehouse->id() : _warehouse->id());
+    ordq.bindValue(":warehous_id", _warehouse->id());
```
gilmoskowitz: This change doesn't look right.

guiclient/salesOrderItem.cpp
```diff
@@ -2767,45 +2770,33 @@ void salesOrderItem::sHandleSupplyOrder()
...
+        else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating Purchase Order"),
+                                      ordq, __FILE__, __LINE__))
+        {
+          return;
+        }
+
+        ordq.prepare("UPDATE coitem "
+                     "SET coitem_order_id=:soitem_order_id "
+                     "WHERE coitem_id=:soitem_id;");
+        ordq.bindValue(":soitem_order_id", _supplyOrderId);
```
gilmoskowitz: This `update coitem` and the characteristic assignment below should probably be handled in `createPurchaseToSale()`.

guiclient/salesOrderItem.ui
```diff
@@ -1287,7 +1287,7 @@ to be bound by its terms.</comment>
                              <bool>false</bool>
                             </property>
                             <property name="type">
-                             <enum>WComboBox::Supply</enum>
+                             <enum>WComboBox::Sold</enum>
```
gilmoskowitz: This change doesn't look right.

guiclient/salesOrderItem.cpp
ChristopherCotnoir: We discussed this with Mike Atherton before the dist locking issue took all my time. The Supply Warehouse dropdown selection is always the same as the Warehouse (except in one case, which we identified as a glitch, addressed below). We decided to ignore it completely for the time being, but Mike said to leave the dropdown intact for the future.

guiclient/salesOrderItem.ui
ChristopherCotnoir: It turns out that the Supply Warehouse is supposed to be the same as the Warehouse, as its behavior wasn't fully implemented (see above). This fixes a glitch wherein a user selects a Warehouse that doesn't exist in the Supply Warehouse dropdown, leaving the signal emitted doing nothing and the Supply Warehouse being whatever happened to be selected before by making its type the same as Warehouse. This was the solution we agreed upon some months ago.

ChristopherCotnoir:
Some things can't be moved to the trigger because they involve prompting the user with windows and dialogs where their answer may result in a Supply Order not being created. The way I dealt with this is not to create the Supply Order in the trigger in cases where this could occur, and if they decide to create it after all, create it here.

guiclient/salesOrderItem.cpp
ChristopherCotnoir: Ah, turns out it already is. Not sure why it was previously done here as well, but I'll remove it.

PR #1534, opened 2017-03-06
Issue #28773: Add Customer PO to Open Receivables

andersdd:
requires xtuple/xtuple#2973


gilmoskowitz:
👍 when CI passes on xtuple/xtuple#2973

PR #1551, opened 2017-03-13
Issue #29602: Return Authorization reporting (Lookup screen)

andersdd:
requires xtuple/xtuple#2998


gilmoskowitz:
👍 merge when xtuple/xtuple#2998 is ready

PR #1561, opened 2017-03-20
issue #22868: Inventory Distribution Locking part 3

jcarlin:
With https://github.com/xtuple/xtuple/pull/3006

Remove locking during inventory distribution for:
- salesOrder
- salesOrderSimple
- issueToShipping
- issueLineToShipping
- unpostedPoReceipts

Extra error handling to allow Cancel All or Continue for: 
- enterPoReceipt
- issueWoMaterialBatch
- unpostedPoReceipts
- postInvoices

Note: Return Stock is still locking (salesOrder, issueToShipping), but functioning.


guiclient/salesOrder.cpp
```diff
@@ -4422,52 +4461,80 @@ void salesOrder::sIssueLineBalance()
...
       }
 
-      issueSales.prepare("SELECT issueLineBalanceToShipping('SO', :soitem_id, now(), :itemlocseries, :invhist_id) AS result;");
-      ;
+      // Before issueToShipping, if controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+      if (controlled)
+      {
+        parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, 'SO', :orderitemId, :itemlocSeries, :invhistId, :itemlocdistId, 'SH') AS result;");
+        parentItemlocdist.bindValue(":itemsite_id", itemsiteId);
+        parentItemlocdist.bindValue(":qty", (balance * issueSales.value("coitem_qty_invuomratio").toDouble()) * -1);
```
jcarlin: Remove `issueSales.value("coitem_qty_invuomratio").toDouble()` now that `issueToShippingLineBalance` handles that math.

guiclient/salesOrder.cpp
```diff
@@ -4304,78 +4303,122 @@ void salesOrder::sIssueLineBalance()
...
+          else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Records"),
+                                    parentItemlocdist, __FILE__, __LINE__))
+          {
+            return;
+          }
+        }
+
         XSqlQuery prod;
-        prod.prepare("SELECT postSoItemProduction(:soitem_id, now()) AS result;");
+        prod.exec("BEGIN"); // Because of possible issueLineBalanceToShipping errors
```
jcarlin: Can BEGIN be called a second time after already in a transaction? Otherwise, L4504 should have a condition.

jcarlin: Missing ';', and in many other places.

guiclient/issueLineToShipping.cpp
```diff
@@ -419,6 +505,9 @@ void issueLineToShipping::populate()
     _qtyReturned->setDouble(itemq.value("qtyreturned").toDouble());
     _qtyReserved->setDouble(itemq.value("qtyreserved").toDouble());
     _balance->setDouble(itemq.value("balance").toDouble());
+
+    _itemsiteId = itemq.value("itemsite_id").toInt();
+    _controlled = itemq.value("controlled").toInt();
```
ChristopherCotnoir: Should be toBool().

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
+      parentSeries, __FILE__, __LINE__);
+    return;
+  }
 
   // If this is a lot/serial controlled job item, we need to post production first
-  if (issueIssue.value("itemsite_costmethod").toString() == "J")
+  if (issueIssue.value("itemsite_costmethod").toString() == "J" && _qtyToIssue->toDouble() > 0)
   {
+    // Before postSoItemProduction, if controlled: create the itemlocdist record and call distributeInventory::SeriesAdjust
+    if (_controlled)
```
ChristopherCotnoir: _controlled here comes from coitem_itemsite_id whereas in the postinvtrans called directly by postProduction wo_itemsite_id is used where wo_ordid=coitem_id. I couldn't find any way for them to be different, but an exhaustive search is proving difficult and it would save more time to just change this to be sure.

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
+    {
+      parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, 'WO', :orderitemId, :itemlocSeries, NULL, NULL, 'RM') AS result;");
+      parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
+      parentItemlocdist.bindValue(":qty", (_qtyToIssue->toDouble() * issueIssue.value("coitem_qty_invuomratio").toDouble()) * -1);
```
ChristopherCotnoir: Sense is positive. Also, should be roundQty(item_fractional, :qty) FROM wo JOIN itemsite on wo_itemsite_id=itemsite_id JOIN item ON itemsite_item_id=item_id.

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
+      parentItemlocdist.bindValue(":orderitemId", _itemid);
```
ChristopherCotnoir: This is the coitem_id, whereas the invhist_ordnumber comes from the wo_id. Additionally, createLotSerial.cpp will require changes to handle this case, as currently it looks for a womatl_id, whereas here a wo_id is used and the work order may have no materials. The best option is probably to have createLotSerial expect a wo_id instead of a womatl_id and simply reformat issueWoMaterialItem.cpp, issueWoMaterialBatch.cpp, and returnWoMaterialItem.cpp to pass the womatl_wo_id instead of womatl_id to createItemlocdistParent as the orderitem_id.

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
+        return;
+      }
+    }
+
+    // postSoItemProduction
     XSqlQuery prod;
-    prod.prepare("SELECT postSoItemProduction(:soitem_id, :qty, :ts) AS result;");
+    prod.exec("BEGIN;");
+    prod.prepare("SELECT postSoItemProduction(:soitem_id, :qty, :ts, :itemlocSeries, TRUE) AS result;");
     prod.bindValue(":soitem_id", _itemid);
```
ChristopherCotnoir: :itemlocSeries needs to be bound.

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
     prod.bindValue(":ts", _transTS);
     prod.exec();
     if (prod.first())
     {
-      itemlocSeries = prod.value("result").toInt();
+      int result = prod.value("result").toInt();
 
-      if (itemlocSeries < 0)
+      if (result < 0 || result != itemlocSeries)
       {
```
ChristopherCotnoir: Call cleanup here.

guiclient/issueLineToShipping.cpp
```diff
@@ -291,21 +338,60 @@ void issueLineToShipping::sIssue()
     }
```
ChristopherCotnoir: Check for error here needed.

guiclient/issueLineToShipping.cpp
```diff
@@ -291,21 +338,60 @@ void issueLineToShipping::sIssue()
...
   }
 
-  issue.prepare("SELECT issueToShipping(:ordertype, :lineitem_id, :qty, :itemlocseries, :ts, :invhist_id) AS result;");
+  // Before issueToShipping, if controlled item: create the parent itemlocdist record, call distributeInventory::seriesAdjust
+  if (_controlled)
+  {
+    parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, :orderType, :orderitemId, :itemlocSeries, :invhistId, :itemlocdistId, 'SH') AS result;");
```
ChristopherCotnoir: itemlocdistId and invhistId shouldn't be passed here. If the order type is SO, another itemlocdist needs to be created. That's where the pInvhistId passing comes into play. See postInvTrans, which will also need some modification to handle this case, as the lsdetail still needs to be created.

guiclient/issueLineToShipping.cpp
```diff
@@ -237,39 +239,84 @@ void issueLineToShipping::sIssue()
...
+                                parentItemlocdist, __FILE__, __LINE__))
+      {
+        return;
+      }
+    }
+
+    // postSoItemProduction
     XSqlQuery prod;
-    prod.prepare("SELECT postSoItemProduction(:soitem_id, :qty, :ts) AS result;");
+    prod.exec("BEGIN;");
```
ChristopherCotnoir: Because we want everything as one big transaction, but we don't want either seriesAdjust in a transaction block, the createItemlocdistParent(s) below needs to be moved up above here, but outside the if statement, and with a different series, and then the BEGIN to COMMIT/ROLLBACK needs to encompass issueToShipping and this, if applicable.

guiclient/issueLineToShipping.cpp
```diff
@@ -291,21 +338,60 @@ void issueLineToShipping::sIssue()
...
+      parentItemlocdist.bindValue(":invhistId", invhistid);
+    if (itemlocdistId > 0)
+      parentItemlocdist.bindValue(":itemlocdistId", itemlocdistId);
+    parentItemlocdist.exec();
+    if (parentItemlocdist.first())
+    {
+      if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(), true)
+        == XDialog::Rejected)
+      {
+        rollback.exec();
```
ChristopherCotnoir: No rollback here.

guiclient/issueLineToShipping.cpp
jcarlin: Yes, I've been working on postInvTrans, createItemlocdistParent changes to handle the invhistId properly. I had noticed that the lsdetail record was not being created yesterday. However, here I need the resulting invhist_id from postSoItemProduction _before_ calling createItemlocdistParent (and SeriesAdjust) for the issue to shipping distribution. Any ideas on how to achieve this?

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
-      setDate.bindValue(":recv_id", id);
-      setDate.exec();
-      if (setDate.lastError().type() != QSqlError::NoError)
-      {
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Receipt Information"),
-                             setDate, __FILE__, __LINE__);
-      }
-    }
-  }
-
```
ChristopherCotnoir: The above doesn't need to be moved, it's separate from the distlocking issue. We should probably avoid changing behavior unrelated to distribution.

guiclient/issueLineToShipping.cpp
jcarlin: Does this also need to be wrapped in a loop since postSoItemProduction is a select from wo where wo_ordid = orderitem_id? Can there be multiple work orders for a orderitem?

ChristopherCotnoir: It shouldn't be possible. I was unable to rule it out looking through the code, but if it does occur, then it's a bug anyway. So don't worry about it.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
+          if (parentItemlocdist.first())
+          {
+            // If the "to" itemsite is not controlled, distribute the From side here
+            if (!recvInfo.value("controlled").toBool())
+            {
+              if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(),
+                true) == XDialog::Rejected)
+              {
+                cleanup.exec();
+                QMessageBox::information( this, tr("Enter Receipts"),
```
ChristopherCotnoir: Unposted Receipts instead of Enter Receipts.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
+          }
+          else 
+          {
+           cleanup.exec();
+           ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Record for "
+             "'From' itemsite"), parentItemlocdist, __FILE__, __LINE__);
+           return;
+          }
+        }
+        else if (tohead.lastError().type() != QSqlError::NoError)
```
ChristopherCotnoir: Should throw an error here instead of just returning.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
+          return;
+      }
+      
+      // Controlled (if 'TO', this is the to itemsite)
+      if (recvInfo.value("controlled").toBool() && 
+        // IF Transfer Order, must be MultiWhs as well
+        (recvInfo.value("recv_order_type").toString() == "TO" ? _metrics->boolean("MultiWhs") : true))
+      {
+        qDebug() << "unpostedPoReceipts controlled";
```
ChristopherCotnoir: Don't output these kinds of messages unless a DEBUG macro is set to true.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
+        XSqlQuery parentItemlocdist;
+        parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, :orderType, :orderitemId, "
+          ":itemlocSeries) AS result;");
```
ChristopherCotnoir: Should pass :itemlocdistId and :transType.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
       if (postLine.first())
       {
         int result = postLine.value("result").toInt();
-        if (result < 0)
+        if (result < 0 || result != itemlocSeries)
         {
             ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Receipt Information"),
                                  storedProcErrorLookup("postReceipt", result),
                                  __FILE__, __LINE__);
           rollback.exec();
```
ChristopherCotnoir: Rollback before throwing error.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -309,24 +454,65 @@ void unpostedPoReceipts::sPost()
...
 
-            if (!_soheadid.contains(issue.value("coitem_cohead_id").toInt()))
-              _soheadid.append(issue.value("coitem_cohead_id").toInt());
+            // Ship any drop shipped orders not previously shipped
+            int coheadId = issue.value("coitem_cohead_id").toInt();
+            if (!_soheadid.contains(coheadId))
+            {
+              _soheadid.append(coheadId);
+
+              // simply moving execute here instead of later iterating, to keep within the transaction
```
ChristopherCotnoir: Moving this here is a problem. Seems the reason it was originally outside the loop is because it is called on the cohead, and needs to be called after all the coitems are called with issueToShipping. shipShipment will set shiphead_shipped to TRUE, meaning if you have two coitems for the cohead the first one will have issueToShipping called, then shipShipment will be called on the cohead even the one coitem hasn't been issues, then the second coitem will be issued, but here shipShipment will be skipped because shiphead_shipped is TRUE. There may be other problems caused by doing this too, I'm still looking into those. It's probably safest to move it back outside the loop even though that means removing it from the transaction block, I don't see any other choice. Perfectly preserving the existing behavior just doesn't seem possible.

ChristopherCotnoir: Disregard this. Turns out subsequent calls to issueToShipping will just create a new shiphead since the previous one had shiphead_shipped set to TRUE, rather than adding shipitems to it.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -196,74 +196,218 @@ void unpostedPoReceipts::sPost()
...
+                      " WHERE orderitem_id = recv_orderitem_id "
+                      "  AND orderitem_orderhead_type = recv_order_type "
+                      "  AND recv_id = :recv_id "
+                      " ORDER BY orderitem_linenumber;");
+      recvInfo.bindValue(":recv_id", id);
+      recvInfo.exec();
+      if (!recvInfo.first())
+      {
+        QMessageBox::information( this, tr("Unposted Receipts"), tr("Failed to retrieve recv and orderitem info.") );
+        return;
```
ChristopherCotnoir: We shouldn't return here, we should continue instead so that if one receipt fails to post it doesn't cause all the subsequent receipts to fail (which is especially bad if some receipts already successfully posted, meaning that some valid receipts will have posted and others won't have decided at random based on where the failing receipt happened to end up). I made the mistake of not catching this on enterPoReceipt, issueWoMaterialBatch, and postInvoices, in the previous PR, even though I fixed this exact bug on issueWoMaterialBatch already. We'll have to go back and change those too.

guiclient/unpostedPoReceipts.cpp
ChristopherCotnoir: Gil says this should still be moved back to where it was, because having multiple shipheads instead of one will change things at invoicing. Even though it is a change in behavior, he said it is better instead of trying to preserve the old behavior where one failure undid everything to instead at the end attempt to ship any cohead that was not involved in at least one failure in the above loop over the receipts.

guiclient/unpostedPoReceipts.cpp
jcarlin: `continue` only here? What about all the other places where it can fail?

xtuple-ci-user:
Build finished.

xtuple-ci-user:
Build finished.

guiclient/distributeInventory.cpp
```diff
@@ -621,9 +621,9 @@ void distributeInventory::sFillList()
   distributeFillList.prepare( "SELECT itemsite_id, "
              "       COALESCE(itemsite_location_id,-1) AS itemsite_location_id,"
              "       formatlotserialnumber(itemlocdist_ls_id) AS lotserial,"
-             // TODO - this order column which populates the _order text field is null for everything except Sales Orders
-             "       CASE WHEN itemlocdist_order_type = 'SO' AND itemlocdist_order_id IS NOT NULL "
-             "            THEN (itemlocdist_order_type || ' ' || formatSoItemNumber(itemlocdist_order_id)) "
+             // TODO - include WO, and any other order types that have an order number
+             "       CASE WHEN itemlocdist_order_type IS NOT NULL AND itemlocdist_order_id IS NOT NULL "
+             "            THEN (itemlocdist_order_type || ' ' || formatOrderLineItemNumber(itemlocdist_order_type, itemlocdist_order_id)) "
```
ChristopherCotnoir: Not sure if we want to change behavior here. It may have only done this for Sales Orders for a reason.

guiclient/issueLineToShipping.cpp
```diff
@@ -233,56 +238,228 @@ void issueLineToShipping::sIssue()
...
+      ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+        parentSeries, __FILE__, __LINE__);
+      return;
+    }
+
+    // Before postSoItemProduction, if controlled: create the itemlocdist record and call distributeInventory::SeriesAdjust
+    if (_controlled)
+    {
+      // Handle creation of itemlocdist records for each eligible backflush item (sql below from postProduction backflush handling)
+      XSqlQuery backflushItems;
```
jcarlin: These backflush items handling should take place outside of the _controlled case. Here, salesOrder, simpleSalesOrder, issueToShipping.

guiclient/salesOrderSimple.cpp
```diff
@@ -1539,111 +1539,209 @@ void salesOrderSimple::sAllocateCreditMemos()
...
+          postProdItemlocSeries = parentSeries.value("result").toInt();
+          postProdCleanup.bindValue(":itemlocSeries", postProdItemlocSeries);
+        }
+        else
+        {
+          ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+            parentSeries, __FILE__, __LINE__);
           return false;
         }
+
```
jcarlin: Missing backflush items handling.

guiclient/issueLineToShipping.cpp
jcarlin: Also, the sql should filter on isControlledItemsite(itemsite_id)

guiclient/enterPoReceipt.cpp
```diff
@@ -350,24 +354,49 @@ void enterPoReceipt::sPost()
...
         }
         else 
         {
-         cleanup.exec();
-         ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Record for "
-           "'From' itemsite"), parentItemlocdist, __FILE__, __LINE__);
-         return;
+          cleanup.exec();
+          ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Record for "
+            "'From' itemsite"), parentItemlocdist, __FILE__, __LINE__);
```
ChristopherCotnoir: Skip the ErrorReporter::error call, we don't want the user to get a message every time, that's why we are appending to the lists to handle at the end.

guiclient/enterPoReceipt.cpp
```diff
@@ -438,7 +489,9 @@ void enterPoReceipt::sPost()
         ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting P/O Receipt Information"),
                                storedProcErrorLookup("postReceipt", result),
                                __FILE__, __LINE__);
-        return;
+        failedItems.append(qi.value("item_number").toString());
+        errors.append(postLine.lastError().text());
```
ChristopherCotnoir: lastError() doesn't work here or in other places with negative return codes; use storedProcErrorLookup, or, if the return code is positive or 0 but not the same as the passed itemlocseries, it means that the function didn't fail outright but postinvtrans/interwarehousetransfer weren't called, meaning distribution didn't occur properly, in which case you'll have to come up with another error message.

guiclient/enterPoReceipt.cpp
```diff
@@ -540,7 +601,9 @@ void enterPoReceipt::sPost()
             rollback.exec();
             cleanup.exec();
             QMessageBox::warning(this, tr("Cannot Ship Order"), msg);
-            return;
+            failedItems.append(qi.value("item_number").toString());
+            errors.append(issue.lastError().text());
```
ChristopherCotnoir: Should be msg, not issue.lastError().text(). Do not call QMessageBox.

guiclient/issueLineToShipping.cpp
```diff
@@ -233,56 +238,228 @@ void issueLineToShipping::sIssue()
...
+      backflushItems.prepare(
+        "SELECT item_number, itemsite_id, womatl_id, "
+        // issueMaterial qty = noNeg(expected - consumed)
+        " noNeg(womatl_qtyfxd + ((roundQty(item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap) - " //expected
```
ChristopherCotnoir: The parentheses aren't matched right for noNeg. It should encompass the entire expression, but it actually encompasses the expression before * (1 + womatl_scrap). Add another parentheses after noNeg and another parentheses before AS qtyToIssue to fix this.

Also, roundQty is using the wrong item_fractional for _parentQty. In postproduction, it comes from wo_itemsite_id, here it comes from womatl_itemsite_id.

guiclient/issueLineToShipping.cpp
```diff
@@ -233,56 +238,228 @@ void issueLineToShipping::sIssue()
...
+        " AND wo_ordid = :coitem_id "
+        " AND wo_ordtype = 'S' "
+        " AND itemsite_item_id=item_id;");
+      backflushItems.bindValue(":qty", _qtyToIssue->toDouble() * issueIssue.value("coitem_qty_invuomratio").toDouble());
+      backflushItems.bindValue(":coitem_id", _itemid);
+      while (backflushItems.next())
+      {
+        if (backflushItems.value("qtyToIssue").toDouble() > 0)
+        {
+          parentItemlocdist.prepare("SELECT createItemlocdistParent(itemsite_id, roundQty(item_fractional, :qty), 'WO', :orderitem_id, "
```
ChristopherCotnoir: :qty should be itemuomtouom(itemsite_item_id, womatl_uom_id, NULL, :qty) where womatl_id=backflushItems.value("womatl_id").

guiclient/issueLineToShipping.cpp
```diff
@@ -233,56 +238,228 @@ void issueLineToShipping::sIssue()
...
+        "     (CASE WHEN (womatl_qtywipscrap >  ((womatl_qtyfxd + (roundQty(item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap)) "
+        "           THEN (womatl_qtyfxd + (roundQty(item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
+        "      ELSE womatl_qtywipscrap END)) AS qtyToIssue " // consumed
+        "FROM womatl, wo, itemsite, item "
+        "WHERE womatl_issuemethod IN ('L', 'M') "
+        " AND womatl_wo_id=wo_id "
+        " AND womatl_itemsite_id=itemsite_id "
+        " AND wo_ordid = :coitem_id "
+        " AND wo_ordtype = 'S' "
+        " AND itemsite_item_id=item_id;");
```
ChristopherCotnoir: ORDER BY womatl_id here and in postproduction to ensure the invhist setting happens correctly.

guiclient/issueLineToShipping.cpp
```diff
@@ -233,56 +238,228 @@ void issueLineToShipping::sIssue()
...
+        }
+      } 
+
+      // Create the post production itemlocdist parent
+      parentItemlocdist.prepare("SELECT createItemlocdistParent(itemsite_id, roundQty(item_fractional, :qty), 'WO', :orderitem_id, "
+                                " :itemlocSeries, NULL, NULL, 'RM') AS result "
+                                "FROM itemsite "
+                                " JOIN item ON itemsite_item_id = item_id "
+                                "WHERE itemsite_id = :itemsite_id;");
+      parentItemlocdist.bindValue(":itemsite_id", _itemsiteId);
```
ChristopherCotnoir: This is coitem_itemsite_id. Should be wo_itemsite_id.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -218,16 +218,38 @@ void issueWoMaterialBatch::sIssue()
...
         {
           cleanup.exec();
-          QMessageBox::information( this, tr("Material Issue"), tr("Error Distributing Inventory Detail (distributeInventory::SeriesAdjust)") );
-          return;
+          // If it's not the last item in the loop, ask the user to exit loop or continue
+          if (items.at() != (items.size() -1))
+          {
+            if (QMessageBox::question(this,  tr("Material Issue"),
+            tr("Posting distribution detail for item number %1 was cancelled but "
+              "there are more items to post. Continue posting the remaining receipts?")
```
ChristopherCotnoir: s/post. Continue posting the remaining receipts/issue. Continue issuing the remaining materials?/

guiclient/distributeInventory.cpp
ChristopherCotnoir: OK, Gil said this change is fine. However, in the CASE statement check that itemlocdist_order_type is one of the types handled by formatOrderLineItemNumber and not just that it isn't NULL, otherwise for say, inventory adjustment, this field will say 'AD'.

ChristopherCotnoir: Or alternatively, just hide this field and its label if it isn't actually an order.

guiclient/postInvoices.cpp
```diff
@@ -210,17 +215,18 @@ void postInvoices::sPost()
...
-
+    // Distribute the items from above
     if (items.size() > 0 && distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(), true)
       == XDialog::Rejected)
     {
       cleanup.exec();
-      QMessageBox::information( this, tr("Post Invoices"), tr("Error Posting Invoice Item Distribution") );
+      failedInvoiceIds.append(invoiceId);
+      errors.append("Detail Distribution was Cancelled");
       return;
```
ChristopherCotnoir: Prompt user whether to continue?

guiclient/postInvoices.cpp
```diff
@@ -243,25 +249,37 @@ void postInvoices::sPost()
       {
         rollback.exec();
         cleanup.exec();
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Invoice"),
-                            storedProcErrorLookup("postInvoice", result),
-                            __FILE__, __LINE__);
-        return;
+        failedInvoiceIds.append(invoiceId);
+        errors.append("Error Posting Invoice");
```
ChristopherCotnoir: Grab the actual error?

guiclient/postInvoices.cpp
```diff
@@ -243,25 +249,37 @@ void postInvoices::sPost()
...
     else if (postPost.lastError().type() != QSqlError::NoError)
     {
       rollback.exec();
       cleanup.exec();
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Invoice Information"),
-                           postPost, __FILE__, __LINE__);
-      return;
-      // TODO - pass into failure array for error reporting
+      failedInvoiceIds.append(invoiceId);
+      errors.append("Error Posting Invoice Information");
```
ChristopherCotnoir: postPost.lastError().text() instead.

guiclient/postInvoices.cpp
```diff
@@ -243,25 +249,37 @@ void postInvoices::sPost()
...
   }
 
+  if (errors.size() > 0)
+  {
+    QMessageBox dlg(QMessageBox::Critical, "Errors Posting Invoice", "", QMessageBox::Ok, this);
+    dlg.setText(tr("%1 Invoices succeeded.\n%2 Invoices failed.").arg(succeeded).arg(failedInvoiceIds.size()));
+
+    QString details;
+    for (int i=0; i<failedInvoiceIds.size(); i++)
+      details += tr("Invoice ID %1 failed with:\n%2\n").arg(failedInvoiceIds[i]).arg(errors[i]);
```
ChristopherCotnoir: ID is a database concept hidden from the user. Better to give invchead_invcnumber instead of invchead_id.

guiclient/unpostedPoReceipts.cpp
ChristopherCotnoir: Ironically, you forgot the continue here.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -211,7 +235,8 @@ void unpostedPoReceipts::sPost()
       int itemlocSeries = -1;
 
       XSqlQuery recvInfo;
-      recvInfo.prepare("SELECT recv_id, itemsite_id, itemsite_controlmethod, itemsite_perishable, itemsite_warrpurc, "
+      recvInfo.prepare("SELECT COALESCE(item_number, recv_vend_item_number) AS item_number, "
+                      "  recv_id, itemsite_id, itemsite_controlmethod, itemsite_perishable, itemsite_warrpurc, "
```
ChristopherCotnoir: After this query, return needs to be replaced with continue.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -288,42 +315,66 @@ void unpostedPoReceipts::sPost()
...
       // Controlled (if 'TO', this is the to itemsite)
       if (recvInfo.value("controlled").toBool() && 
         // IF Transfer Order, must be MultiWhs as well
         (recvInfo.value("recv_order_type").toString() == "TO" ? _metrics->boolean("MultiWhs") : true))
       {
-        qDebug() << "unpostedPoReceipts controlled";
         XSqlQuery parentItemlocdist;
         parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, :orderType, :orderitemId, "
-          ":itemlocSeries) AS result;");
+          ":itemlocSeries, NULL, :itemlocdistId, :transType, :itemlocSeries) AS result;");
```
ChristopherCotnoir: Don't pass :itemlocSeries twice.

guiclient/enterPoReceipt.cpp
```diff
@@ -350,24 +354,49 @@ void enterPoReceipt::sPost()
...
+                  "there are more items to post. Continue posting the remaining receipts?")
+                .arg(qi.value("item_number").toString()),
+                QMessageBox::Yes | QMessageBox::No, QMessageBox::No) == QMessageBox::Yes)
+                {
+                  failedItems.append(qi.value("item_number").toString());
+                  errors.append("Detail Distribution Cancelled");
+                  continue;
+                }
+                else
+                  return;
```
ChristopherCotnoir: If the function returns here, any previous errors added to the lists won't be reported. Instead set a flag and continue, and continue at the beginning of the loop if the flag is set.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -457,62 +513,27 @@ void unpostedPoReceipts::sPost()
               cleanup.exec();
               QMessageBox::warning(this, tr("Cannot Ship Order"), msg);
               sFillList();
-              return;
+              failedItems.append(recvInfo.value("item_number").toString());
+              errors.append(issue.lastError().text());
```
ChristopherCotnoir: Use msg here. Don't call warning.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -531,10 +552,50 @@ void unpostedPoReceipts::sPost()
...
+        failedItems.append(recvInfo.value("item_number").toString());
+        errors.append(postLine.lastError().text());
+        continue;
       }
+      succeeded++;
       unpostedPost.exec("COMMIT;"); 
     } // for each selected line
 
+    // Ship any drop shipped orders
+    while (_soheadid.count())
```
ChristopherCotnoir: We only want to do this for coheads where none of the coitems failed on issuetoShipping.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -324,9 +528,12 @@ void unpostedPoReceipts::sPost()
...
+            cleanup.exec();
+            failedItems.append(recvInfo.value("item_number").toString());
+            errors.append(issue.lastError().text());
+            continue;
           }
-        }    
+        }
       }
       // contains() string is hard-coded in stored procedure
       else if (postLine.lastError().databaseText().contains("posted to closed period"))
```
ChristopherCotnoir: This and the do ; while construct in general are a mess. Gil said to change that. Instead of doing this here, we can determine at the top of the loop whether this will happen (or would happen, if it weren't for the typo) with

SELECT period_closed
FROM recv JOIN period ON COALESCE(recv_gldistdate, recv_date) BETWEEN period_start AND period_end
WHERE recv_id=:id

If that is the case, only set/append to triedToClosed if _privileges->check("ChangePORecvPostDate"), then continue regardless, but don't append to errors. Present the user with the errors accumulated in the lists in the do ; while loop instead of outside it, before the failedPostList dialog. If the user accepts the failedPostList dialog, give the user a chance to actually change the recv_gldistdate for the failed receipts like at the top of this function, as it stands now the whole thing is pointless because the user isn't given that chance meaning it's just going to fail again anyway.

Also, make sure you don't call shipShipment multiple times for the same cohead and do call it for coheads that previously didn't get shipped because one or more coitems failed if all failed coitems eventually succeed on subsequent tries.

guiclient/distributeInventory.cpp
jcarlin: I thought checking that both order_type and order_id are not null was sufficient to determine if it's an "order". If it's not an order_type handled by the format function, it will just display as `<order_type> || ' '` AD and other transactional types will not have an order_id value.

ChristopherCotnoir: Do we want it showing even the Order Type if it's not actually an order? 'Order #: AD' looks awkward. If we're still going to show that, should probably change the label to 'Transaction Type' or something more appropriate in that case.

guiclient/unpostedPoReceipts.cpp
jcarlin: I left this as return because that query should always return a record.

guiclient/issueLineToShipping.cpp
jcarlin: I thought you determined that they are always the same.

guiclient/unpostedPoReceipts.cpp
jcarlin: > If the user accepts the failedPostList dialog, give the user a chance to actually change the recv_gldistdate for the failed receipts...

Can't we just move the date change code to the top of the `do` loop to handle both the first time and the case you mention above?

guiclient/unpostedPoReceipts.cpp
ChristopherCotnoir: Should, but things can go wrong. If we weren't accounting for that possibility, we wouldn't have the if block and error message in the first place. And if this one fails for some reason (maybe data corruption), we don't have reason to expect that all other recv records will fail.

guiclient/issueLineToShipping.cpp
ChristopherCotnoir: No, I did not. It seems likely, but proving it will take longer than just checking for wo_itemsite_id, and better safe than sorry.

guiclient/unpostedPoReceipts.cpp
ChristopherCotnoir: I think that should work.

guiclient/distributeInventory.cpp
jcarlin: Done.

guiclient/createLotSerial.cpp
```diff
@@ -76,21 +76,12 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
                       " itemsite_id, itemsite_perishable, itemsite_warrpurc, "
                       " COALESCE(itemsite_lsseq_id,-1) AS itemsite_lsseq_id, "
-                      " COALESCE(invhist_ordnumber, CASE "
-                      "   WHEN orderhead_id IS NULL AND itemlocdist_order_type = 'WO' THEN formatWoNumber(womatl_wo_id) "
-                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'SO' THEN formatSoNumber(orderitem_id) "
-                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'TO' THEN formatToNumber(orderitem_id) "
-                      "   WHEN orderhead_id IS NOT NULL AND orderhead_type = 'RA' THEN orderhead_number || '-' || orderitem_linenumber "
-                      "   ELSE '' END) AS ordnumber, "
-                      " itemlocdist_order_id, "
+                      " COALESCE(invhist_ordnumber, formatOrderLineItemNumber(itemlocdist_order_type, itemlocdist_order_id)) AS ordnumber, "
```
ChristopherCotnoir: This doesn't work for itemlocdist_order_type='RA'. It should be looking for rahead_number || '-' || raitem_linenumber, but formatOrderLineItemNumber returns rahead_number || '-' || formatralinenumber(raitem_id), which is different.

guiclient/issueLineToShipping.cpp
```diff
@@ -312,136 +318,76 @@ void issueLineToShipping::sIssue()
...
-          postProdCleanup.exec();
+          cleanup.exec();
           return;
         }
       }
     }
 
-    // If Job item is controlled, create the itemlocdist record for postProduction transaction
+    // If it's a controlled job item, set the relavant params
     if (_controlled)
```
ChristopherCotnoir: This is from coitem_itemsite_id, when we want wo_itemsite_id.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -180,90 +180,311 @@ void unpostedPoReceipts::sViewOrderItem()
...
+      XSqlQuery closedPeriod;
+      closedPeriod.prepare("SELECT period_closed "
+                           "FROM recv "
+                           "  JOIN period ON COALESCE(recv_gldistdate, recv_date) BETWEEN period_start AND period_end "
+                           "WHERE recv_id=:recv_id;");
+      closedPeriod.bindValue(":recv_id", id);
+      closedPeriod.exec();
+      if (!closedPeriod.first() || closedPeriod.value("period_closed").toBool())
+      {
+        if (!_privileges->check("ChangePORecvPostDate"))
```
ChristopherCotnoir: Shouldn't be NOT.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -309,9 +533,13 @@ void unpostedPoReceipts::sPost()
...
                   .arg(issue.value("pohead_number").toString());
               rollback.exec();
-              QMessageBox::warning(this, tr("Cannot Ship Order"), msg);
+              cleanup.exec();
               sFillList();
-              return;
+              failedItems.append(recvInfo.value("item_number").toString());
+              errors.append(msg);
+              // Remove from drop ship list so that it's not shipped
+              _soheadid.removeAll(issue.value("coitem_cohead_id").toInt());
```
ChristopherCotnoir: This is insufficient because if there is one that succeeds after one that fails, it will just be added back. Additionally, this is only done here but the order shouldn't be shipped if a line item on that order that should be shipped doesn't get shipped because of a failure anywhere else in the function.

guiclient/issueLineToShipping.cpp
```diff
@@ -197,21 +201,31 @@ void issueLineToShipping::sIssue()
...
+                "  ELSE NULL END AS postprodqty, "
                 "  (noNeg(coitem_qtyord - coitem_qtyshipped + coitem_qtyreturned) <"
-                "           (COALESCE(SUM(shipitem_qty), 0) + <? value(\"qty\") ?>)) AS overship"
+                "           (COALESCE(SUM(shipitem_qty), 0) + <? value(\"qty\") ?>)) AS overship, wo_id "
                 "  FROM coitem LEFT OUTER JOIN"
                 "        ( shipitem JOIN shiphead"
                 "          ON ( (shipitem_shiphead_id=shiphead_id) AND (NOT shiphead_shipped) )"
                 "        ) ON  (shipitem_orderitem_id=coitem_id)"
                 "  JOIN itemsite ON (coitem_itemsite_id=itemsite_id) "
+                "  JOIN item ON itemsite_item_id = item_id "
```
ChristopherCotnoir: Remove this, it isn't being used.

guiclient/issueLineToShipping.cpp
```diff
@@ -274,83 +287,81 @@ void issueLineToShipping::sIssue()
...
-        " AND itemsite_item_id=item_id;");
-      backflushItems.bindValue(":qty", _qtyToIssue->toDouble() * issueIssue.value("coitem_qty_invuomratio").toDouble());
-      backflushItems.bindValue(":coitem_id", _itemid);
-      while (backflushItems.next())
+      qDebug() << "backflushItems qtytoissue: " << backflushItems.value("qtyToIssue").toDouble();
+      if (backflushItems.value("qtyToIssue").toDouble() > 0)
       {
-        if (backflushItems.value("qtyToIssue").toDouble() > 0)
+        hasControlledBackflushItems = true;
+        parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, itemuomtouom(:item_id, womatl_uom_id, NULL, :qty) * -1, 'WO', :orderitem_id, "
```
ChristopherCotnoir: Put back the roundQty. Should be roundQty(item_fractional, itemuomtouom(:item_id, womatl_uom_id, NULL, :qty)) * -1

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,270 @@ void issueLineToShipping::sIssue()
...
+    {
+      parentItemlocdist.bindValue(":orderitemId", issueIssue.value("wo_id").toInt());
+      parentItemlocdist.bindValue(":orderType", "WO");
+      parentItemlocdist.bindValue(":transType", "RM");
+      parentItemlocdist.bindValue(":qty", issueIssue.value("issuelineqty").toDouble());
+    }
+  } // job item
+
+  // Create the itemlocdist record if controlled item and distribute detail if controlled or controlled backflush items
+  if (_controlled || hasControlledBackflushItems)
```
ChristopherCotnoir: || issueIssue.value("woItemControlled").toBool() here and below.

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,270 @@ void issueLineToShipping::sIssue()
...
+      }
+    }
+
+    // If it's a controlled job item, set the relavant params
+    if (issueIssue.value("woItemControlled").toBool())
+    {
+      parentItemlocdist.bindValue(":orderitemId", issueIssue.value("wo_id").toInt());
+      parentItemlocdist.bindValue(":orderType", "WO");
+      parentItemlocdist.bindValue(":transType", "RM");
+      parentItemlocdist.bindValue(":qty", issueIssue.value("issuelineqty").toDouble());
```
ChristopherCotnoir: Should be postprodqty instead of issuelineqty. Also, bind the wo_itemsite_id to :itemsite_id.

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,270 @@ void issueLineToShipping::sIssue()
...
-      itemlocSeries = prod.value("result").toInt();
+      int result = prod.value("result").toInt();
 
-      if (itemlocSeries < 0)
+      if (result < 0 || result != itemlocSeries)
       {
         rollback.exec();
+        cleanup.exec();
         ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Production"),
                                storedProcErrorLookup("postProduction", itemlocSeries),
```
ChristopherCotnoir: s/itemlocSeries/result.

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,270 @@ void issueLineToShipping::sIssue()
...
+                               tr("Inventory history not found")
+                               .arg(windowTitle()),__FILE__,__LINE__);
+          return;
+        }
       }
     }
+    else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Production for Job Item"),
+      prod, __FILE__,__LINE__))
+    {
+      rollback.exec();
```
ChristopherCotnoir: rollback before ErrorReporter.

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,270 @@ void issueLineToShipping::sIssue()
...
-                "WHERE ((invhist_series = :itemlocseries) "
-                " AND (invhist_transtype = 'RM')); ");
-      prod.bindValue(":itemlocseries" , itemlocSeries);
-      prod.exec();
-      if (prod.first())
-        invhistid = prod.value("invhist_id").toInt();
-      else
+      // If controlled item, get the inventory history from post production trans. 
+      // so we can create itemlocdist records for issue to shipping transaction and auto-distribute to them in postInvTrans.
+      if (_controlled)
```
ChristopherCotnoir: Should be issueIssue.value("woItemControlled").toBool()

guiclient/unpostedPoReceipts.cpp
```diff
@@ -274,7 +276,7 @@ void unpostedPoReceipts::sPost()
       closedPeriod.exec();
       if (!closedPeriod.first() || closedPeriod.value("period_closed").toBool())
       {
```
ChristopherCotnoir: failedReceipts.append(id) here.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -218,16 +224,40 @@ void issueWoMaterialBatch::sIssue()
...
         }
       }
       else
       {
         cleanup.exec();
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Records"),
-          parentItemlocdist, __FILE__, __LINE__);
-        return;
+        failedItems.append(items.value("item_number").toString());
+        errors.append(tr("Error Creating itemlocdist Records")
```
ChristopherCotnoir: Have a %1 for the .arg.

guiclient/postInvoices.cpp
```diff
@@ -210,18 +222,31 @@ void postInvoices::sPost()
       if (!parentItemlocdist.first())
       {
         cleanup.exec();
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Records"),
-                                parentItemlocdist, __FILE__, __LINE__);
-        return;
+        failedInvoiceNumbers.append(invoiceNumber);
+        errors.append(tr("Error Creating itemlocdist Record for item %1").arg(items.value("item_number").toString()));
+        break;
```
ChristopherCotnoir: continue, not break.

guiclient/salesOrderItem.cpp
```diff
@@ -3598,7 +3598,7 @@ void salesOrderItem::sPopulateOrderInfo()
   }
   
   if (_costmethod == "J")
-    _createSupplyOrder->setEnabled(false);
+    //_createSupplyOrder->setEnabled(false);
   
   _supplyOrderQtyCache = _supplyOrderQty->toDouble();
   _supplyOrderDueDateCache = _supplyOrderDueDate->date();
```
ChristopherCotnoir: Why is this file changed?

jcarlin: Crap. It was added by mistake. I had to change that to be able to create supply order line items.

guiclient/issueLineToShipping.cpp
```diff
@@ -234,126 +248,272 @@ void issueLineToShipping::sIssue()
...
+      parentItemlocdist.bindValue(":itemsite_id", issueIssue.value("wo_itemsite_id").toInt());
+      parentItemlocdist.bindValue(":orderitemId", issueIssue.value("wo_id").toInt());
+      parentItemlocdist.bindValue(":orderType", "WO");
+      parentItemlocdist.bindValue(":transType", "RM");
+      parentItemlocdist.bindValue(":qty", issueIssue.value("postprodqty").toDouble());
+    }
+  } // job item
+
+  // Create the itemlocdist record if controlled item and distribute detail if controlled or controlled backflush items
+  if (_controlled || issueIssue.value("woItemControlled").toBool() || hasControlledBackflushItems)
```
ChristopherCotnoir: issueIssue.value("woItemControlled").toBool() needs to be (issueIssue.value("woItemControlled").toBool() && issueIssue.value("itemsite_costmethod").toString() == "J" && _qtyToIssue->toDouble() > 0). Oth

guiclient/enterPoReceipt.cpp
ChristopherCotnoir: Derp, actually what is needed after adding the error is break.

guiclient/enterPoReceipt.cpp
```diff
@@ -593,7 +648,9 @@ void enterPoReceipt::sPost()
             {
```
ChristopherCotnoir: Rollback before ErrorReporter.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+                       " AND orderitem_id = :orderitemId;");
+  issueDetail.bindValue(":orderType", _order->type());
+  issueDetail.bindValue(":orderitemId", id);
+  issueDetail.exec();
+  if (issueDetail.first())
+  {
+    if (issueDetail.size() > 1)
+    {
+      ErrorReporter::error(QtCriticalMsg, this, tr("Retrieving from orderitem table returned more than one row."),
+        issueDetail, __FILE__, __LINE__);
```
ChristopherCotnoir: ErrorReporter with the 4th parameter as an XSqlQuery only throws an error if the XSqlQuery has an error. This cannot be the case here. Instead, the 3rd parameter should be the title and the 4th parameter should be the error description.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+      return false;
+    }
+
+    
+  }
+  else
+  {
+    ErrorReporter::error(QtCriticalMsg, this, tr("Could not find this orderitem_id in the orderitem table."),
+      issueDetail, __FILE__, __LINE__);
```
ChristopherCotnoir: There are two ways this can be called. The query could have an error, perhaps a user has a bad database that is missing roundQty or something. The query could also return no results, which appears to be what is being checked here. In the second case, ErrorReporter will do nothing for the reason described above. Either pass issueDetail.lastError().text() instead of issueDetail so the user will still be shown an error if nothing is found with just a title, or handle the cases separately.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+      "           THEN (womatl_qtyfxd + (roundQty(item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
+      "      ELSE womatl_qtywipscrap END)) AS qtyToIssue " // consumed
+      "FROM womatl, wo, itemsite, item "
+      "WHERE womatl_issuemethod IN ('L', 'M') "
+      " AND womatl_wo_id=wo_id "
+      " AND womatl_itemsite_id=itemsite_id "
+      " AND wo_ordid = :coitem_id "
+      " AND wo_ordtype = 'S' "
+      " AND itemsite_item_id=item_id "
+      " AND isControlledItemsite(itemsite_id);");
```
ChristopherCotnoir: This query has mismatched parentheses and incorrect roundQty calls. It should match the query in issueLineToShipping exactly.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+  issue.exec("BEGIN;");
+
+  // If Job item, postSoItemProduction
+  if (altId == 1)
+  {
     XSqlQuery prod;
-    prod.prepare("SELECT postSoItemProduction(:soitem_id, :ts) AS result;");
+    prod.prepare("SELECT postSoItemProduction(:soitem_id, :qty, now(), :itemlocSeries, TRUE) AS result;");
     prod.bindValue(":soitem_id", id);
-    prod.bindValue(":ts", _transDate->date());
```
ChristopherCotnoir: Don't remove this.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+    prod.bindValue(":qty", balance);
```
ChristopherCotnoir: postSoItemProduction multiplies this value by coitem_qty_invuomratio before passing to postProduction. calcIssueToShippingLineBalance also multiplies by coitem_qty_invuomratio, resulting in this happening twice here, which we don't want. Remove the coitem_qty_invuomratio multiplication from calcIssueToShippingLineBalance and put it inside the roundQty for postprodqty above, or something equivalent that ensures coitem_qty_invuomratio multiplication is done correctly.

jcarlin: I have already corrected this on my local branch. Thanks.

jcarlin: I have already corrected this on my local branch. Thanks.

guiclient/issueToShipping.cpp
jcarlin: I have already corrected this on my local branch. Thanks.

guiclient/issueToShipping.cpp
```diff
@@ -442,52 +444,221 @@ void issueToShipping::sIssueLineBalance()
...
+    ":itemlocSeries, NULL, NULL, :transType) AS result;");
+  parentItemlocdist.bindValue(":itemsite_id", itemsiteId);
+  parentItemlocdist.bindValue(":qty", balance * -1);
+  parentItemlocdist.bindValue(":orderitemId", id);
+  parentItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
+  parentItemlocdist.bindValue(":orderType", _order->type());
+  parentItemlocdist.bindValue(":transType", "SH");
+  
   // If this is a lot/serial controlled job item, we need to post production first
   if (altId == 1)
```
ChristopherCotnoir: && issueDetail.value("balance").toDouble() here and below.

guiclient/salesOrder.cpp
```diff
@@ -4295,179 +4295,298 @@ void salesOrder::sIssueLineBalance()
...
+          "           THEN (womatl_qtyfxd + (roundQty(item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
+          "      ELSE womatl_qtywipscrap END)) AS qtyToIssue " // consumed
+          "FROM womatl, wo, itemsite, item "
+          "WHERE womatl_issuemethod IN ('L', 'M') "
+          " AND womatl_wo_id=wo_id "
+          " AND womatl_itemsite_id=itemsite_id "
+          " AND wo_ordid = :coitem_id "
+          " AND wo_ordtype = 'S' "
+          " AND itemsite_item_id=item_id "
+          " AND isControlledItemsite(itemsite_id);");
```
ChristopherCotnoir: Fix mismatched parentheses and such. Should be the same query as in issueLineToShipping.cpp.

guiclient/salesOrder.cpp
```diff
@@ -4295,179 +4295,298 @@ void salesOrder::sIssueLineBalance()
...
+            }
+          }
         }
-        if (prod.first())
+
+        // If it's a controlled job item, set the relavant params
+        if (issueSales.value("woItemControlled").toBool())
         {
-          itemlocSeries = prod.value("result").toInt();
+          parentItemlocdist.bindValue(":itemsite_id", issueSales.value("wo_itemsite_id").toInt());
```
ChristopherCotnoir: Needs to be selected above.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -320,33 +559,49 @@ void unpostedPoReceipts::sPost()
...
+                              "WHERE recv_id = :recv_id ");
+      failedLineItems.bindValue(":recv_id", failedReceipts.at(i));
+      failedLineItems.exec();
+      if (failedLineItems.first())
+      {
+        _soheadid.removeAll(failedLineItems.value("cohead_id").toInt());
+      }
+    }
+
     // Ship any drop shipped orders
```
ChristopherCotnoir: Below, break instead of return so previous errors get reported.

guiclient/postInvoices.cpp
```diff
@@ -210,18 +222,33 @@ void postInvoices::sPost()
       if (!parentItemlocdist.first())
       {
         cleanup.exec();
-        ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist Records"),
-                                parentItemlocdist, __FILE__, __LINE__);
-        return;
+        failedInvoiceNumbers.append(invoiceNumber);
+        errors.append(tr("Error Creating itemlocdist Record for item %1").arg(items.value("item_number").toString()));
+        continue;
```
ChristopherCotnoir: This will actually just continue the innermost loop, when we want to continue the outer loop. Use a flag that is checked after the inner loop as a workaround.

guiclient/postInvoices.cpp
```diff
@@ -210,18 +222,33 @@ void postInvoices::sPost()
...
     if (items.size() > 0 && distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(), true)
       == XDialog::Rejected)
     {
       cleanup.exec();
-      QMessageBox::information( this, tr("Post Invoices"), tr("Error Posting Invoice Item Distribution") );
-      return;
+      if (QMessageBox::question(this,  tr("Post Invoices"),
+        tr("Posting distribution detail for invoice number %1 was cancelled but "
+           "there other invoices to Post. Continue posting the remaining invoices?")
+        .arg(items.value("item_number").toString()),
```
ChristopherCotnoir: Should be invoiceNumber.

guiclient/postInvoices.cpp
```diff
@@ -224,9 +225,13 @@ void postInvoices::sPost()
         cleanup.exec();
         failedInvoiceNumbers.append(invoiceNumber);
         errors.append(tr("Error Creating itemlocdist Record for item %1").arg(items.value("item_number").toString()));
-        continue;
+        invoiceLineFailed = true;
```
ChristopherCotnoir: break; after setting invoiceLineFailed = true;.

jcarlin: oh, right

guiclient/unpostedPoReceipts.cpp
```diff
@@ -180,90 +180,324 @@ void unpostedPoReceipts::sViewOrderItem()
...
+              if (distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(), QDate(),
+                true) == XDialog::Rejected)
+              {
+                cleanup.exec();
+                // If it's not the last item in the loop, ask the user to exit loop or continue
+                if (i != (selected.size() - 1))
+                {
+                  if (QMessageBox::question(this,  tr("Unposted Receipts"),
+                                tr("Posting distribution detail for item number %1 was cancelled but "
+                                  "there are more items to post. Continue posting the remaining receipts?"),
```
ChristopherCotnoir: Needs .arg for the %1.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -180,90 +180,324 @@ void unpostedPoReceipts::sViewOrderItem()
...
+                  {
+                    failedReceipts.append(id);
+                    failedItems.append(recvInfo.value("item_number").toString());
+                    errors.append("Detail Distribution Cancelled");
+                    continue;
+                  }
+                  else
+                  {
+                    tryagain = false;
+                    continue;
```
ChristopherCotnoir: Should be break. tryagain handles the do; while loop, not the for loop in this function.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -180,90 +180,324 @@ void unpostedPoReceipts::sViewOrderItem()
...
+                              "there are more items to post. Continue posting the remaining receipts?"),
+                            QMessageBox::Yes | QMessageBox::No, QMessageBox::No) == QMessageBox::Yes)
+              {
+                failedReceipts.append(id);
+                failedItems.append(recvInfo.value("item_number").toString());
+                errors.append("Detail Distribution Cancelled");
+                continue;
+              }
+              else
+                break;
```
ChristopherCotnoir: Need to append to lists first.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -320,33 +560,49 @@ void unpostedPoReceipts::sPost()
...
+      unpostedPost.exec("COMMIT;"); 
     } // for each selected line
 
+    // Remove sales orders from _soheadid list that had a line item fail so that we don't ship that sales order
+    for (int i = 0; i < failedReceipts.count(); i++)
+    {
+      XSqlQuery failedLineItems;
+      failedLineItems.prepare("SELECT cohead_id "
+                              "FROM recv "
+                              " JOIN poitem ON poitem_id=recv_orderitem_id "
```
ChristopherCotnoir: Also make sure recv_order_type = 'PO' here, otherwise you could have a TO/RA recv that fails whose orderitem_id just so happens to be the same as the orderitem_id of a PO recv that was dropshipped, which we wouldn't want affected.

guiclient/unpostedPoReceipts.cpp
```diff
@@ -370,20 +626,31 @@ void unpostedPoReceipts::sPost()
           rollback.exec();
```
ChristopherCotnoir: Remove rollback here and below, this part isn't in a transaction.

PR #1566, opened 2017-03-29
issue #29385 - only reset default filter on first showEvent

gilmoskowitz:
Converted the displayPrivate class to a QObject so it can
accept signal/slot connections. Then added a slot and
tracking variable so we can limit finding and setting the
filter to the first time a window instance is shown.

### DIFF
--- a/guiclient/display.cpp
+++ b/guiclient/display.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
  * version 1.0, the full text of which (including xTuple-specific Exhibits)
  * is available at www.xtuple.com/CPAL.  By using this software, you agree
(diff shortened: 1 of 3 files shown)


guiclient/display.cpp
```diff
@@ -454,7 +412,12 @@ void display::showEvent(QShowEvent * e)
 {
   XWidget::showEvent(e);
 
-  parameterWidget()->applyDefaultFilterSet();
+  // don't overwrite the user's filter when the window is minimized
+  if (! _data->_filterChanged)
+  {
+    parameterWidget()->applyDefaultFilterSet();
+    connect(parameterWidget(), SIGNAL(filterChanged()), _data, SLOT(sFilterChanged()));
```
ChristopherCotnoir: If the user repeatedly minimizes/unminimizes the window before changing the filter, this causes this connection to be made each time that happens. This means if a user opens the window, then minimizes/unminimizes it 10 times, this signal gets emitted 10 times every time the filter changes until the window is closed.

guiclient/display.cpp
```diff
@@ -28,161 +27,120 @@
 #include <parameter.h>
 #include <previewdialog.h>
 
-#include "../scriptapi/parameterlistsetup.h"
+#include "parameterlistsetup.h"
```
ChristopherCotnoir: Including this and parameter.h is redundant since they are already included in displayprivate.h, which is included here.

PR #1567, opened 2017-03-30
issue #29730 - reduce unnecessary db calls and window openings

gilmoskowitz:
By checking for non-zero quantity and valid item selection,
we can avoid a few database queries when there's nothing
worthwhile to do and avoid opening the Item Sources window.
Also combined two queries.

### DIFF
--- a/guiclient/salesOrderItem.cpp
+++ b/guiclient/salesOrderItem.cpp
@@ -2696,10 +2696,10 @@ void salesOrderItem::sHandleSupplyOrder()
                      "  FROM itemsrc"
                      "  JOIN vendinfo ON itemsrc_vend_id = vend_id"
                      "  LEFT OUTER JOIN pohead ON vend_id = pohead_vend_id"
-                     " WHERE itemsrc_id      = :itemsrc_id"
-                     "   AND pohead_status   = 'U'"
-                     "   AND pohead_dropship = :drop_ship"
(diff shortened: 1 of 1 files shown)


guiclient/salesOrderItem.cpp
```diff
@@ -2691,51 +2692,42 @@ void salesOrderItem::sHandleSupplyOrder()
...
-                      "FROM itemsrc LEFT OUTER JOIN vendinfo ON (vend_id = itemsrc_vend_id) "
-                      "WHERE (itemsrc_id=:itemsrc_id);" );
+        ordq.prepare("SELECT itemsrc_vend_id, vend_name, pohead_id"
+                     "  FROM itemsrc"
+                     "  JOIN vendinfo ON itemsrc_vend_id = vend_id"
+                     "  LEFT OUTER JOIN pohead ON vend_id = pohead_vend_id"
+                     " WHERE itemsrc_id      = :itemsrc_id"
+                     "   AND pohead_status   = 'U'"
+                     "   AND pohead_dropship = :drop_ship"
+                     "   AND (NOT pohead_dropship OR (pohead_cohead_id = :sohead_id));");
```
ChristopherCotnoir: These last 3 conditions should be in the LEFT OUTER JOIN, otherwise they cause an error to be thrown below if no such pohead exists.

PR #1570, opened 2017-04-04
issue #29447 Move message handlers to qt-client and modify login2 to allow messagi…

ChristopherCotnoir:
…ng to command line during login process


ChristopherCotnoir:
Associated with xtuple/updater#49.

PR #1571, opened 2017-04-07
issue #29852 Undo RETURNING simplification because it doesn't take after triggers …

ChristopherCotnoir:
…into account


harithamekala:
Performed UAT: Issue is Fixed

PR #1574, opened 2017-04-15
Qt 5 8 compile fixes

dpocock:
These are the fixes to make qt-client build on Qt 5.8, they do not impact support for earlier Qt versions.


davidbeauchamp:
I think this is missing the qtsetup.cpp changes for the hint removals

dpocock:
@davidbeauchamp it looks like somebody did the hint removals in another commit

dpocock:
@davidbeauchamp you were right, they are still required but the code had been rearranged in a recent commit causing a merge conflict when I tried to introduce that change.  Now it is in this pull request too.

PR #1576, opened 2017-04-18
Resolves issue #29934 - target item number is not displayed in Transform Trans

jcarlin:


guiclient/transformTrans.cpp
```diff
@@ -409,6 +410,8 @@ void transformTrans::sFillList()
   {
     return;
   }
+
+  sPopulateTarget();
```
ChristopherCotnoir: This shouldn't be necessary. The clear/populate above ensure that _target will emit newId, calling this already.

ChristopherCotnoir: Never mind, apparently this isn't occurring. The PR is good to merge.

PR #1580, opened 2017-04-20
Support for Sales Order (line item) > Return Stock sans locking

jcarlin:
https://github.com/xtuple/xtuple/pull/3015

In qt-client, all calls to returnItemShipments are wrapped in a transaction, followed by distributeInventory::SeriesAdjust. I propose that the sole purpose of this was to call postItemlocSeries https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/distributeInventory.cpp#L355. Any and all distribution detail occurs entirely in the database because all nested functions of returnItemShipments that call postInvTrans always pass the invhist_id. 

~~I added postItemlocSeries to returnItemShipments (pr linked above).~~ I added postItemlocSeries to postInvTrans https://github.com/xtuple/xtuple/pull/3015/files#diff-986ae11ec8bee1fa06e1ef93bf5affeaR245. This pr removes the transaction blocks from the return stock methods in:
- issueToShipping.cpp
- maintainShipping.cpp
- shippingInformation.cpp
- transferOrder.cpp
- salesOrder.cpp

*TODO for future pr:* The following files should have the postItemlocSeries call removed as it's now handled in postInvTrans anytime invhist_id is sent.
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/enterPoReceipt.cpp#L362
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/enterPoReceipt.cpp#L412
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/postProduction.cpp#L338
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/postProduction.cpp#L373
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/scrapWoMaterialFromWIP.cpp#L249
- https://github.com/xtuple/qt-client/blob/4_10_x/guiclient/unpostedPoReceipts.cpp#L319

### DIFF
--- a/guiclient/maintainShipping.cpp
+++ b/guiclient/maintainShipping.cpp
@@ -205,6 +205,8 @@ void maintainShipping::sViewOrder()
   {
     return;
   }
+
+  sFillList();
 }
 
 void maintainShipping::sPrintShippingForm()
--- a/guiclient/maintainShipping.cpp
+++ b/guiclient/maintainShipping.cpp
@@ -183,6 +183,8 @@ void maintainShipping::sReturnAllOrderStock()
                          maintainReturnAllOrderStock, __FILE__, __LINE__);
     return;
   }
+
+  sFillList();
 }
 
 void maintainShipping::sViewOrder()
@@ -205,8 +207,6 @@ void maintainShipping::sViewOrder()
   {
     return;
   }
-
-  sFillList();
 }
 
 void maintainShipping::sPrintShippingForm()
@@ -244,6 +244,8 @@ void maintainShipping::sReturnAllLineStock()
                          maintainReturnAllLineStock, __FILE__, __LINE__);
     return;
   }
+
+  sFillList();
 }
 
 void maintainShipping::sViewLine()
@@ -271,40 +273,17 @@ void maintainShipping::sViewLine()
 void maintainShipping::sReturnAllStock()
 {
   XSqlQuery maintainReturnAllStock;
-  XSqlQuery rollback;
-  rollback.prepare("ROLLBACK;");
-
-  maintainReturnAllStock.exec("BEGIN");
   maintainReturnAllStock.prepare("SELECT returnShipmentTransaction(:ship_id) AS result;");
   maintainReturnAllStock.bindValue(":ship_id", _ship->altId());
   maintainReturnAllStock.exec();
-  if (maintainReturnAllStock.first())
+  if (maintainReturnAllStock.lastError().type() != QSqlError::NoError)
   {
-    int result = maintainReturnAllStock.value("result").toInt();
-    if (maintainReturnAllStock.value("result").toInt() < 0)
-    {
-      rollback.exec();
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Shipment Information"),
-                             storedProcErrorLookup("returnShipmentTransaction", result),
-                             __FILE__, __LINE__);
-      return;
-    }
-    else if (distributeInventory::SeriesAdjust(result, this) == XDialog::Rejected)
-    {
-      rollback.exec();
-      QMessageBox::information( this, tr("Issue to Shipping"), tr("Return Canceled") );
-      return;
-    }    
-    maintainReturnAllStock.exec("COMMIT;"); 
-    sFillList();
-  }
-  else if (maintainReturnAllStock.lastError().type() != QSqlError::NoError)
-  {
-    rollback.exec();
     ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Shipment Information"),
                          maintainReturnAllStock, __FILE__, __LINE__);
     return;
   }
+
+  sFillList();
 }
 
 void maintainShipping::sFillList()
--- a/guiclient/maintainShipping.cpp
+++ b/guiclient/maintainShipping.cpp
@@ -174,36 +174,11 @@ void maintainShipping::sShipOrder()
 void maintainShipping::sReturnAllOrderStock()
 {
   XSqlQuery maintainReturnAllOrderStock;
-  XSqlQuery rollback;
-  rollback.prepare("ROLLBACK;");
-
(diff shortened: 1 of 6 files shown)


guiclient/salesOrder.cpp
```diff
@@ -4557,7 +4531,7 @@ void salesOrder::sIssueLineBalance()
 
       // issueToShipping instead of issueLineBalanceToShipping because we have already calculated the balance
       issueSales.prepare("SELECT issueToShipping('SO', :soitem_id, :qty, :itemlocseries, now(), "
-                         ":invhist_id, TRUE) AS result;");
+                         ":invhist_id, FALSE, TRUE) AS result;");
```
jcarlin: This was an mistake in the last merged pr. Before this change, inventory detail was not posting for Issue Line from Sales Order.

guiclient/maintainShipping.cpp
```diff
@@ -174,36 +174,11 @@ void maintainShipping::sShipOrder()
 void maintainShipping::sReturnAllOrderStock()
 {
   XSqlQuery maintainReturnAllOrderStock;
-  XSqlQuery rollback;
-  rollback.prepare("ROLLBACK;");
-
-  maintainReturnAllOrderStock.exec("BEGIN");
   maintainReturnAllOrderStock.prepare("SELECT returnCompleteShipment(:ship_id) AS result;");
```
ChristopherCotnoir: This function is different from returnItemShipments, so it needs postItemlocSeries added to it too.

guiclient/maintainShipping.cpp
```diff
@@ -174,36 +174,11 @@ void maintainShipping::sShipOrder()
...
-      return;
-    }
-    else if (distributeInventory::SeriesAdjust(result, this) == XDialog::Rejected)
-    {
-      rollback.exec();
-      QMessageBox::information( this, tr("Issue to Shipping"), tr("Return Canceled") );
-      return;
-    }
-    maintainReturnAllOrderStock.exec("COMMIT;"); 
-    sFillList();
```
ChristopherCotnoir: Don't remove this call, still call this before the function ends if no error.

guiclient/salesOrder.cpp
```diff
@@ -4213,48 +4213,22 @@ bool salesOrder::okToProcessCC()
...
-        return;
-      }
-
-    }
-    else if (returnSales.lastError().type() != QSqlError::NoError)
-    {
-      rollback.exec();
       ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Return Item Information"),
                            returnSales, __FILE__, __LINE__);
       return;
```
ChristopherCotnoir: If this is here and the transaction block is gone, if one function call fails but the others succeed, previous calls will stay but new ones will not be executed even if both would have succeeded, which is confusing to the user. Use continue like before.

guiclient/transferOrder.cpp
```diff
@@ -1766,46 +1766,20 @@ void transferOrder::viewTransferOrder( int pId )
...
-      }
-
-    }
-    else if (transferReturnStock.lastError().type() != QSqlError::NoError)
+  if (transferReturnStock.lastError().type() != QSqlError::NoError)
     {
-      rollback.exec();
       ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Return Item Information"),
                            transferReturnStock, __FILE__, __LINE__);
       return;
```
ChristopherCotnoir: continue here as well.

ChristopherCotnoir:
You forgot issueToShipping.cpp.

guiclient/maintainShipping.cpp
jcarlin: Done.

guiclient/maintainShipping.cpp
```diff
@@ -258,36 +235,11 @@ void maintainShipping::sIssueStock()
...
-      return;
-    }
-    else if (distributeInventory::SeriesAdjust(result, this) == XDialog::Rejected)
-    {
-      rollback.exec();
-      QMessageBox::information( this, tr("Issue to Shipping"), tr("Return Canceled") );
-      return;
-    }    
-    maintainReturnAllLineStock.exec("COMMIT;"); 
-    sFillList();
```
ChristopherCotnoir: Keep this too.

guiclient/maintainShipping.cpp
```diff
@@ -205,6 +205,8 @@ void maintainShipping::sViewOrder()
   {
     return;
   }
+
+  sFillList();
```
ChristopherCotnoir: This is in a different function...

jcarlin:
I need to remove locking from maintainShipping::sReturnAllStock() as well.

jcarlin:
I was looking at adding postItemlocSeries to returnshipmenttransaction to handle this: https://github.com/xtuple/qt-client/pull/1580/files#diff-d4d86427d88931100a5ce0b53158c038R278. I did some testing after adding PERFORM postItemlocSeries(...)` to the bottom of returnshipmenttransaction. Now postItemlocSeries is being called multiple times before the end of the distributions for various transactions. There were no errors in testing. I think I should undo all the postItemlocSeries additions in this pr and move it to postInvTrans. https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/functions/postinvtrans.sql#L222. 

I believe all these cases of sending invhist_id to postinvtrans will be OK if postItemlocSeries is called within. The legacy code that remains will call postItemlocSeries again but because of the logic in that function, nothing actually happens. The deletion of itemlocpost records in postItemlocSeries wont matter either because each call to postInvTrans inserts into itemlocpost.

ChristopherCotnoir:
Yeah, you can just pass pPreDistributed as TRUE to the postinvtrans calls and let it call postitemlocseries, it can be called multiple times. It was already doing that anyway (See returnShipmentTransaction below the Sales Order postinvtrans), and you are right that it is safe to do.

PR #1582, opened 2017-04-26
issue #22868 - Post Production without locking

jcarlin:
https://github.com/xtuple/xtuple/pull/3018
https://github.com/xtuple/private-extensions/pull/900

### DIFF
--- a/guiclient/postProduction.cpp
+++ b/guiclient/postProduction.cpp
@@ -188,11 +188,9 @@ bool postProduction::okToPost()
   XSqlQuery itemsite;
   itemsite.prepare("SELECT wo_itemsite_id, "
                    "  itemsite_costmethod, "
-                   "	item_type IN ('R', 'F') AS invalid_itemtype, "
-                   "  isControlledItemsite(wo_itemsite_id) AS controlled, item_type "
+                   "  isControlledItemsite(wo_itemsite_id) AS controlled "
                    "FROM wo "
                    "  JOIN itemsite ON wo_itemsite_id = itemsite_id "
-                   "  JOIN item ON itemsite_item_id = item_id "
                    "WHERE wo_id = :woId;");
   itemsite.bindValue(":woId", _wo->id());
   itemsite.exec();
@@ -208,13 +206,6 @@ bool postProduction::okToPost()
                           tr("<p>Work Orders Item Sites with Job cost method "
                              "can not be transferred to another warehouse. "
                              "Uncheck the Immediate Transfer to Site box to continue.") );
-        return false;
-      }
-
-      if (itemsite.value("invalid_itemtype").toBool())
-      {
-      	QMessageBox::critical( this, tr("Cannot Transfer"),
-                          tr("<p>Item Type %1 cannot be transfered.").arg(itemsite.value("item_type").toString()) );
         return false;
       }
     }
@@ -262,18 +253,6 @@ QString postProduction::updateWoAfterPost()
   return result;
 }
 
-QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
-{
-  QString result = QString::null;
-  if (distributeInventory::SeriesAdjust(itemlocSeries, this) == XDialog::Rejected)
-    result = tr("Transaction Canceled");
-
-  if (DEBUG)
-    qDebug("postProduction::handleSeriesAdjustAfterPost() returning %s",
-           qPrintable(result));
-  return result;
-}
-
 int postProduction::handleSeriesAdjustBeforePost()
 {
   XSqlQuery parentItemlocdist;
@@ -314,9 +293,8 @@ int postProduction::handleSeriesAdjustBeforePost()
       "        THEN (womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
       "        ELSE womatl_qtywipscrap END)) "
       " ELSE " // returnWoMaterial qty = expected * -1
-      "   ((womatl_qtyfxd + ((roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) * -1 " 
+      "   roundQty(woitem.item_fractional, :qty) * womatl_qtyper " 
       " END AS qty, "
-      "	roundQty(woitem.item_fractional, :qty) * womatl_qtyper AS return_qty, "
       "	womatl_id, womatl_wo_id, womatl_uom_id, womatl_qtyreq, womatl_qtyiss "
       "FROM womatl, wo "
 			"	JOIN itemsite AS woitemsite ON wo_itemsite_id=woitemsite.itemsite_id "
@@ -377,16 +355,16 @@ int postProduction::handleSeriesAdjustBeforePost()
       {
         // make sure the backflush item has relevant qty for returnWoMaterial
         if (backflushItems.value("womatl_qtyreq").toDouble() >= 0 ? 
-            backflushItems.value("womatl_qtyiss").toDouble() >= backflushItems.value("return_qty").toDouble() : 
-            backflushItems.value("womatl_qtyiss").toDouble() <= backflushItems.value("return_qty").toDouble())
+            backflushItems.value("womatl_qtyiss").toDouble() >= backflushItems.value("qty").toDouble() : 
+            backflushItems.value("womatl_qtyiss").toDouble() <= backflushItems.value("qty").toDouble())
         {
           hasControlledBackflushItems = true;
-          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, :womatl_uom_id, NULL, :qty), :qty) * -1, "
-          													"'WO', :wo_id, :itemlocSeries, NULL, NULL, 'IM') AS result;");
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, :womatl_uom_id, NULL, :qty), :qty), "
+          													"  'WO', :wo_id, :itemlocSeries, NULL, NULL, 'IM') AS result;");
           womatlItemlocdist.bindValue(":itemsite_id", backflushItems.value("itemsite_id").toInt());
           womatlItemlocdist.bindValue(":item_id", backflushItems.value("itemsite_item_id").toInt());
           womatlItemlocdist.bindValue(":womatl_uom_id", backflushItems.value("womatl_uom_id").toInt());
-          womatlItemlocdist.bindValue(":qty", backflushItems.value("return_qty").toDouble());
+          womatlItemlocdist.bindValue(":qty", backflushItems.value("qty").toDouble());
           womatlItemlocdist.bindValue(":wo_id", _wo->id());
           womatlItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
           womatlItemlocdist.exec();
@@ -453,6 +431,7 @@ int postProduction::handleTransferSeriesAdjustBeforePost()
 {
   int twItemlocSeries = 0;
   int toWhItemsiteId = 0;
+  int fromWhItemlocdistId = 0;
   bool toWhControlled = false;
 
   if (!_immediateTransfer->isChecked())
@@ -479,7 +458,7 @@ int postProduction::handleTransferSeriesAdjustBeforePost()
 
   // Get TO warehouse itemsite and control values
   XSqlQuery toWh;
-  toWh.prepare("SELECT itemsite_id, isControlled(itemsite_id) AS controlled "
+  toWh.prepare("SELECT itemsite_id, isControlledItemsite(itemsite_id) AS controlled "
   						 "FROM itemsite "
                "WHERE itemsite_warehous_id = :warehouseId "
                " AND itemsite_item_id = "
@@ -500,18 +479,28 @@ int postProduction::handleTransferSeriesAdjustBeforePost()
     return -1;
   }
 
-  if (!_controlled && !toWhControlled)
-    return twItemlocSeries;
-
-  XSqlQuery fromWhItemlocdist;
-  fromWhItemlocdist.prepare("SELECT createItemlocdistParent(:itemsiteId, :qty * -1, 'W', :orderitemId, "
-                            "   :itemlocSeries, NULL, NULL, 'TW') AS result;");
-  fromWhItemlocdist.bindValue(":itemsiteId", _itemsiteId);
-  fromWhItemlocdist.bindValue(":qty", _qty->toDouble());
-  fromWhItemlocdist.bindValue(":orderitemId", _wo->id());
-  fromWhItemlocdist.bindValue(":itemlocSeries", twItemlocSeries);
-  fromWhItemlocdist.exec();
-  if (fromWhItemlocdist.first() && toWhControlled)
+  if (_controlled)
+  {
+    XSqlQuery fromWhItemlocdist;
+    fromWhItemlocdist.prepare("SELECT createItemlocdistParent(:itemsiteId, :qty * -1, 'W', :orderitemId, "
+                              "   :itemlocSeries, NULL, NULL, 'TW') AS result;");
+    fromWhItemlocdist.bindValue(":itemsiteId", _itemsiteId);
+    fromWhItemlocdist.bindValue(":qty", _qty->toDouble());
+    fromWhItemlocdist.bindValue(":orderitemId", _wo->id());
+    fromWhItemlocdist.bindValue(":itemlocSeries", twItemlocSeries);
+    fromWhItemlocdist.exec();
+    if (fromWhItemlocdist.first())
+      fromWhItemlocdistId = fromWhItemlocdist.value("result").toInt();
+    else
+    {
+      cleanup.exec();
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Creating itemlocdist record for FROM Warehouse for interwarehousetransfer "),
+        fromWhItemlocdist, __FILE__, __LINE__);
+      return -1;
+    }
+  }
+  
+  if (toWhControlled)
   {
     XSqlQuery toWhItemlocdist;
     toWhItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, 'W', :orderitemId, "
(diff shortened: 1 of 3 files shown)


guiclient/postProduction.cpp
```diff
@@ -218,260 +218,459 @@ bool postProduction::okToPost()
...
-{
-  QString result = QString::null;
-  if (_immediateTransfer->isChecked())
+  int itemsiteId = 0;
+  int itemlocSeries = 0;
+  int twItemlocSeries = 0;
+  bool controlled = false;
+  bool hasControlledBackflushItems = false;
+
+  // Stage cleanup function to be called on error
```
jcarlin: I need another cleanup for the twItemlocSeries.

guiclient/postProduction.cpp
```diff
@@ -218,260 +218,459 @@ bool postProduction::okToPost()
...
+            return;
+          }
+        }
+        
+      }
+      // create the itemlocdist record for this controlled returnWoMaterial item
+      else if (_wo->method() == "D")
+      {
+        // make sure the backflush item has relevant qty for returnWoMaterial
+        if (backflushItems.value("womatl_qtyreq").toDouble() >= 0 ?
```
jcarlin: This is also in the sql below, remove one and make other related adjustments.

guiclient/postProduction.cpp
```diff
@@ -218,260 +218,459 @@ bool postProduction::okToPost()
...
+      {
+        // make sure the backflush item has relevant qty for returnWoMaterial
+        if (backflushItems.value("womatl_qtyreq").toDouble() >= 0 ? 
+            backflushItems.value("womatl_qtyiss").toDouble() > backflushItems.value("qty").toDouble() : 
+            backflushItems.value("womatl_qtyiss").toDouble() < backflushItems.value("qty").toDouble())
+        {
+          hasControlledBackflushItems = true;
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, womatl_uom_id, NULL, :qty), :qty) * -1, 'WO', womatl_wo_id, "
+                                    " :itemlocSeries, NULL, NULL, 'IM') AS result "
+                                    "FROM womatl "
```
jcarlin: This doesn't need to select from womatl again, that information can all come from the backflushItems sql.

guiclient/postProduction.cpp
```diff
@@ -181,28 +184,34 @@ bool postProduction::okToPost()
...
+    _itemsiteId = itemsite.value("wo_itemsite_id").toInt();
+    _controlled = itemsite.value("controlled").toBool();
+    if (_immediateTransfer->isChecked() && itemsite.value("itemsite_costmethod").toString() == "J")
+    {
+      QMessageBox::critical( this, tr("Can not Transfer"),
+                          tr("<p>Work Orders Item Sites with Job cost method "
+                             "can not be transferred to another warehouse. "
+                             "Uncheck the Immediate Transfer to Site box to continue.") );
+      return false;
+    }
```
ChristopherCotnoir: Where did this check come from?

jcarlin: I believe it was an oversight. https://github.com/xtuple/private-extensions/blob/4_11_x/source/inventory/foundation-database/public/functions/interwarehousetransfer.sql#L70 does nothing for Job items.... hmm, I should be checking that item type is not R or F as well? Also, I was wrong, now that I look at this line again, it checks for item type J not itemsite_costmethod. What is item type J? Is this from a time before Job was moved to a cost method instead of item type?

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+      "SELECT item_number, item_fractional, itemsite_id, itemsite_item_id, womatl_id, womatl_wo_id, womatl_uom_id, "
+      " CASE WHEN :qty > 0 THEN " //issueWoMaterial qty = noNeg(expected - consumed)
+      "   noNeg(((womatl_qtyfxd + ((:qty + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) - "
+      "   (womatl_qtyiss + "
+      "   CASE WHEN (womatl_qtywipscrap >  ((womatl_qtyfxd + (:qty + wo_qtyrcv) * womatl_qtyper) * womatl_scrap)) "
+      "        THEN (womatl_qtyfxd + (:qty + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
+      "        ELSE womatl_qtywipscrap END)) "
+      " ELSE " // returnWoMaterial qty = expected * -1
+      "   ((womatl_qtyfxd + ((:qty + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) * -1 " 
+      " END AS qty "
```
ChristopherCotnoir: The :qty here (except for the comparison with 0, where it is acceptable but not necessary) needs to be roundQty(item_fractional, :qty) FROM wo JOIN itemsite ON wo_itemsite_id=itemsite_id JOIN item ON itemsite_item_id=item_id WHERE wo_id=:wo_id. Note that this item_fractional is from wo, not womatl which is where the item_fractional directly selected is from.

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+        {
+          hasControlledBackflushItems = true;
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, roundQty(:item_fractional, itemuomtouom(:item_id, :womatl_uom_id, NULL, :qty)) * -1, 'WO', :wo_id, "
+                                    " :itemlocSeries, NULL, NULL, 'IM') AS result;");
+          womatlItemlocdist.bindValue(":itemsite_id", backflushItems.value("itemsite_id").toInt());
+          womatlItemlocdist.bindValue(":item_id", backflushItems.value("itemsite_item_id").toInt());
+          womatlItemlocdist.bindValue(":item_fractional", backflushItems.value("item_fractional").toBool());
+          womatlItemlocdist.bindValue(":wo_id", _wo->id());
+          womatlItemlocdist.bindValue(":qty", backflushItems.value("qty").toDouble());
+          womatlItemlocdist.bindValue(":itemlocSeries", itemlocSeries);
```
ChristopherCotnoir: :womatl_uom_id needs to be bound.

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+        }
+        
+      }
+      // create the itemlocdist record for this controlled returnWoMaterial item
+      else if (_wo->method() == "D")
+      {
+        // make sure the backflush item has relevant qty for returnWoMaterial
+        if (backflushItems.value("womatl_qtyreq").toDouble() >= 0 ? 
+            backflushItems.value("womatl_qtyiss").toDouble() > backflushItems.value("qty").toDouble() : 
+            backflushItems.value("womatl_qtyiss").toDouble() < backflushItems.value("qty").toDouble())
```
ChristopherCotnoir: Should be >= and <=.

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+        {
+          hasControlledBackflushItems = true;
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, womatl_uom_id, NULL, :qty), :qty) * -1, 'WO', womatl_wo_id, "
```
ChristopherCotnoir: Don't multiply by -1. It gets multiplied by -1 before being passed to postInvTrans, which then reverses it because the sense of 'IM' is negative.

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+        }
+      }
+    }
+  } // backflush handling
+
+  // If controlled item, createItemlocdistParent
+  if (_controlled)
+  {
+    // create the RM itemlocdist record
+    parentItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, :qty, 'WO', :orderitemId, "
```
ChristopherCotnoir: :qty should be roundQty(item_fractional, :qty).

guiclient/postProduction.cpp
```diff
@@ -252,7 +261,254 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+  }
+  else 
+  {
+    ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+      parentSeries, __FILE__, __LINE__);
+    return -1;
+  }
+
+  if (!_controlled)
+    return twItemlocSeries;
```
ChristopherCotnoir: This results in no itemlocdist entries being created if the source itemsite is not controlled, even if the target itemsite is controlled. Make sure to handle the two sites separately in that regard, like with enterPoReceipt.cpp.

guiclient/postProduction.cpp
ChristopherCotnoir: Yeah, check for those. Not sure what item_type 'J' is, but as long as we check for that, it shouldn't be an issue. interwarehousetransfer does nothing in that case, rather than throw an error, so I don't think we should throw an error for that case, but we should skip the itemlocdist creation.

guiclient/postProduction.cpp
```diff
@@ -252,7 +274,281 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+      {
+        // make sure the backflush item has relevant qty for issueWoMaterial
+        if (backflushItems.value("qty").toDouble() > 0)
+        {
+          hasControlledBackflushItems = true;
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, roundQty(:item_fractional, itemuomtouom(:item_id, :womatl_uom_id, NULL, :qty)) * -1, 'WO', :wo_id, "
+                                    " :itemlocSeries, NULL, NULL, 'IM') AS result;");
+          womatlItemlocdist.bindValue(":itemsite_id", backflushItems.value("itemsite_id").toInt());
+          womatlItemlocdist.bindValue(":item_id", backflushItems.value("itemsite_item_id").toInt());
+          womatlItemlocdist.bindValue(":item_id", backflushItems.value("womatl_uom_id").toInt());
```
jcarlin: womatl_uom_id

guiclient/postProduction.cpp
```diff
@@ -494,13 +807,14 @@ void postProduction::sPost()
...
     rollback.exec();
+    cleanup.exec();
     ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Production"),
                          postPost, __FILE__, __LINE__);
     return;
   }
 
   if (_captive)
-    done(itemlocSeries);
+    close();
```
jcarlin: The screen was not closing but I need to check to see what was expecting itemlocSeries return val, and test to verify if this was a latent bug or caused by other changes here.

jcarlin: @ChristopherCotnoir I believe this change is OK, along with https://github.com/xtuple/private-extensions/pull/900#discussion-diff-114246501R294.

guiclient/postProduction.h
```diff
@@ -26,11 +26,13 @@ class postProduction : public XDialog, public Ui::postProduction
 
     Q_INVOKABLE bool    captive() { return _captive; }
     Q_INVOKABLE QString handleSeriesAdjustAfterPost(int itemlocSeries);
```
jcarlin: I should probably remove this. I left it thinking that there might be other scripts that use it but I don't think there are.

guiclient/postProduction.cpp
```diff
@@ -252,7 +274,283 @@ QString postProduction::handleSeriesAdjustAfterPost(int itemlocSeries)
...
+  else 
+  {
+    ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve the Next itemloc_series_seq"),
+      parentSeries, __FILE__, __LINE__);
+    return -1;
+  }
+
+  // Get TO warehouse itemsite and control values
+  XSqlQuery toWh;
+  toWh.prepare("SELECT itemsite_id, isControlled(itemsite_id) AS controlled "
```
jcarlin: Should be `isControlledItemsite(...`.

guiclient/postProduction.cpp
```diff
@@ -186,24 +186,37 @@ bool postProduction::okToPost()
...
                              "Uncheck the Immediate Transfer to Site box to continue.") );
-      return false;
+        return false;
+      }
+
+      if (itemsite.value("invalid_itemtype").toBool())
+      {
+      	QMessageBox::critical( this, tr("Cannot Transfer"),
+                          tr("<p>Item Type %1 cannot be transfered.").arg(itemsite.value("item_type").toString()) );
+        return false;
```
ChristopherCotnoir: postProduction shouldn't be prevented in this case, this should only be used to prevent itemlocdist creation for interwarehousetransfer.

guiclient/postProduction.cpp
```diff
@@ -293,23 +306,28 @@ int postProduction::handleSeriesAdjustBeforePost()
...
+      "   CASE WHEN (womatl_qtywipscrap >  ((womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap)) "
+      "        THEN (womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
       "        ELSE womatl_qtywipscrap END)) "
       " ELSE " // returnWoMaterial qty = expected * -1
-      "   ((womatl_qtyfxd + ((:qty + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) * -1 " 
-      " END AS qty "
-      "FROM womatl, wo, itemsite, item "
+      "   ((womatl_qtyfxd + ((roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) * -1 " 
+      " END AS qty, "
+      "	roundQty(woitem.item_fractional, :qty) * womatl_qtyper AS return_qty, "
```
ChristopherCotnoir: This should be undone. All the usages of this were correct as they were before.

guiclient/postProduction.cpp
```diff
@@ -354,25 +373,24 @@ int postProduction::handleSeriesAdjustBeforePost()
...
-            backflushItems.value("womatl_qtyiss").toDouble() < backflushItems.value("qty").toDouble())
+            backflushItems.value("womatl_qtyiss").toDouble() >= backflushItems.value("return_qty").toDouble() : 
+            backflushItems.value("womatl_qtyiss").toDouble() <= backflushItems.value("return_qty").toDouble())
         {
           hasControlledBackflushItems = true;
-          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, womatl_uom_id, NULL, :qty), :qty) * -1, 'WO', womatl_wo_id, "
-                                    " :itemlocSeries, NULL, NULL, 'IM') AS result "
-                                    "FROM womatl "
-                                    "WHERE womatl_id = :womatl_id;");
+          womatlItemlocdist.prepare("SELECT createItemlocdistParent(:itemsite_id, COALESCE(itemuomtouom(:item_id, :womatl_uom_id, NULL, :qty), :qty) * -1, "
```
ChristopherCotnoir: Do not multiply by -1 here.

guiclient/postProduction.cpp
```diff
@@ -464,21 +511,16 @@ int postProduction::handleTransferSeriesAdjustBeforePost()
   fromWhItemlocdist.bindValue(":orderitemId", _wo->id());
   fromWhItemlocdist.bindValue(":itemlocSeries", twItemlocSeries);
   fromWhItemlocdist.exec();
```
ChristopherCotnoir: This should be skipped if not _controlled.

guiclient/postProduction.cpp
```diff
@@ -684,17 +726,15 @@ void postProduction::sPost()
...
   if (_wo->method() == "A")
     postPost.bindValue(":qty", _qty->toDouble());
   else
     postPost.bindValue(":qty", _qty->toDouble() * -1);
   postPost.bindValue(":backflushMaterials", QVariant(_backflush->isChecked()));
   postPost.bindValue(":itemlocSeries", itemlocSeries);
-  postPost.bindValue(":date",  _transDate->date() == QDate::currentDate()
-                               ? QDateTime::currentDateTime() : QDateTime(_transDate->date()));
-
+  postPost.bindValue(":date",  _transDate->date());
```
ChristopherCotnoir: Don't change this.

guiclient/postProduction.cpp
ChristopherCotnoir: No, scrapWoMaterialFromWIP.cpp uses the return val. Try and find out why it is not closing.

guiclient/postProduction.h
ChristopherCotnoir: Yes, that is safe to remove.

guiclient/postProduction.cpp
jcarlin: Done. But doesn't L729 change to the same?

guiclient/postProduction.cpp
jcarlin: Actually, there was a latent bug in which the entire qty was being returned instead of the qty proportionate to the Qty posting here.

guiclient/postProduction.cpp
ChristopherCotnoir: I don't know what you mean. You can remove the timestamp with time zone cast if that's what you mean, I don't know why you added that in the first place. Nothing about the datetime passing should change from previous behavior, it is completely unrelated to the locking problem.

guiclient/postProduction.cpp
ChristopherCotnoir: OK, then. In that case, it can just be moved to inside the case statement where expected was before, as that isn't being used anymore. Also, don't forget to multiply by -1.

jcarlin: Oh, right. I must have forgotten about that when testing, should have replaced it.

guiclient/postProduction.cpp
```diff
@@ -314,9 +293,8 @@ int postProduction::handleSeriesAdjustBeforePost()
       "        THEN (womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
       "        ELSE womatl_qtywipscrap END)) "
       " ELSE " // returnWoMaterial qty = expected * -1
-      "   ((womatl_qtyfxd + ((roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) * -1 " 
+      "   roundQty(woitem.item_fractional, :qty) * womatl_qtyper "
```
ChristopherCotnoir: Should be * -1 here.

jcarlin: `:qty` is (already) negative in this part of the case statement.

ChristopherCotnoir: * -1 to make it positive. See postProduction, if pQty is negative, return_qty is still multiplied by -1 before being passed to returnWoMaterial. This has to match that.

PR #1596, opened 2017-05-18
issue #22868 - Remove distribution locking from remaining classes 

jcarlin:
Except for `printMulticopyDocument.cpp`, this PR contains all the remaining classes to resolve issue #22868.

Interdependent on: https://github.com/xtuple/private-extensions/pull/904, https://github.com/xtuple/xtuple/pull/3033.

- correctProductionPosting.cpp
- dspAROpenItems.cpp
- issueWoMaterialBatch.cpp - _minor corrections from previous pr_
- issueWoMaterialItem.cpp - _remove unnecessary locking because issueWoMaterial() doesn't return (-)_
- postCreditMemos.cpp
- postInvoices.cpp
- postMiscProduction.cpp
- returnWoMaterialItem.cpp
- scrapWoMaterialFromWIP.cpp
- transferOrder.cpp
- transferTrans.cpp
- transformTrans.cpp
- unpostedCreditMemos.cpp
- unpostedInvoices.cpp
- workOrder.cpp


guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+
+  cleanup.bindValue(":itemlocSeries", itemlocSeries);
+
+  // If backflush materials, createItemlocdistParent for each (for returnWoMaterial() call)
+  if (_backFlush->isChecked())
+  {
+    XSqlQuery backflushMaterials;
+    backflushMaterials.prepare("SELECT wo_qtyrcv, womatl_id, womatl_uom_id, itemsite_id, "
+                               "  roundQty(item_fractional, "
+                               "    (CASE WHEN wo_qtyrcv - :qty > 0 THEN 0 ELSE womatl_qtyfxd END + :qty * womatl_qtyper) * (1 + womatl_scrap)) * -1 AS qty "
```
ChristopherCotnoir: Do not multiply by -1, IM is a negative transaction type so it cancels.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+    ErrorReporter::error(QtCriticalMsg, this, tr("Failed to Retrieve WO and Itemsite Info"),
+                            postInfo, __FILE__, __LINE__);
+    return;
+  }
 
-  correctCorrect.exec("BEGIN;");	// handle cancel of lot, serial, or loc distribution
-  correctCorrect.prepare("SELECT correctProduction(:wo_id, :qty, :backflushMaterials, 0, :date)"
-            "        AS result;");
+  itemlocSeries = distributeInventory::SeriesCreate(postInfo.value("wo_itemsite_id").toInt(),
+    postInfo.value("parent_qty").toDouble() * -1, "WO", "RM", _wo->id());
```
ChristopherCotnoir: Don't do this or the backflush itemlocdist creation if _qty->toDouble()==0.0, otherwise they will be left as orphaned itemlocdist records when correctProduction returns early.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+                               "  JOIN item ON (item_id=itemsite_item_id) "
+                               "WHERE wo_id=:wo_id "
+                               "  AND isControlledItemsite(womatl_itemsite_id);");
+    backflushMaterials.bindValue(":wo_id", _wo->id());
+    backflushMaterials.bindValue(":qty", postInfo.value("parent_qty").toDouble());
+    backflushMaterials.exec();
+    while (backflushMaterials.next())
+    {
+      if (backflushMaterials.value("qty").toDouble() > 0)
+      {
```
ChristopherCotnoir: returnWoMaterial exits before calling postinvtrans if

SELECT (
         CASE WHEN (womatl_qtyreq >= 0) THEN
           womatl_qtyiss < pQty
         ELSE
           womatl_qtyiss > pQty
         END )
       FROM womatl
       WHERE ( womatl_id=pWomatlid )

Make sure to check for this.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+        hasControlledMaterialItems = true;
+        int result = distributeInventory::SeriesCreate(backflushMaterials.value("itemsite_id").toInt(),
+          backflushMaterials.value("qty").toDouble(), "WO", "IM", _wo->id(), itemlocSeries);
```
ChristopherCotnoir: Should be itemuomtouom(itemsite_item_id, womatl_uom_id, NULL, qty) instead of just qty, where the itemsite_item_id comes from womatl_itemsite_id. Also, make sure that transformation is done after checking for the above case.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+        if (result != itemlocSeries)
+          return;
```
ChristopherCotnoir: Should call cleanup if this happens.

guiclient/distributeInventory.cpp
```diff
@@ -621,10 +677,7 @@ void distributeInventory::sFillList()
   distributeFillList.prepare( "SELECT itemsite_id, "
              "       COALESCE(itemsite_location_id,-1) AS itemsite_location_id,"
              "       formatlotserialnumber(itemlocdist_ls_id) AS lotserial,"
-             "       CASE WHEN itemlocdist_order_id IS NOT NULL "
-             "              AND itemlocdist_order_type IN ('SO', 'PO', 'TO', 'WO', 'RA', 'IN') " // these types are handled by formatOrderLineItemNumber
-             "            THEN (itemlocdist_order_type || ' ' || formatOrderLineItemNumber(itemlocdist_order_type, itemlocdist_order_id)) "
-             "            ELSE '' END AS order, "
+             "       itemlocdist_order_type || ' ' || formatOrderLineItemNumber(itemlocdist_order_type, itemlocdist_order_id) AS order, "
```
ChristopherCotnoir: Why is the check being removed? Should just add 'CM' to the list.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+    XSqlQuery backflushMaterials;
+    backflushMaterials.prepare("SELECT wo_qtyrcv, womatl_id, womatl_uom_id, itemsite_id, "
+                               "  roundQty(item_fractional, "
+                               "    (CASE WHEN wo_qtyrcv - :qty > 0 THEN 0 ELSE womatl_qtyfxd END + :qty * womatl_qtyper) * (1 + womatl_scrap)) * -1 AS qty "
+                               "FROM wo "
+                               "  JOIN womatl ON (womatl_wo_id=wo_id AND womatl_issuemethod='L') "
+                               "  JOIN itemsite ON (itemsite_id=womatl_itemsite_id) "
+                               "  JOIN item ON (item_id=itemsite_item_id) "
+                               "WHERE wo_id=:wo_id "
+                               "  AND isControlledItemsite(womatl_itemsite_id);");
```
ChristopherCotnoir: ORDER BY womatl_id here and in the corresponding query in correctProduction.sql to ensure the itemlocdist are created in the same order as postinvtrans is called.

guiclient/dspAROpenItems.cpp
```diff
@@ -643,40 +643,78 @@ void dspAROpenItems::sVoidCreditMemo()
...
+                  "WHERE cmhead_id=:cmheadId "
+                  " AND cmitem_qtyreturned <> 0 "
+                  " AND cmitem_updateinv "
+                  " AND isControlledItemsite(cmitem_itemsite_id) "
+                  "ORDER BY cmitem_id;");
+  cmitems.bindValue(":cmheadId", list()->currentItem()->id("docnumber"));
+  cmitems.exec();
+  while (cmitems.next())
+  {
+    if (distributeInventory::SeriesCreate(cmitems.value("itemsite_id").toInt(),
```
ChristopherCotnoir: itemsite_id isn't selected above. Should be cmitem_itemsite_id.

guiclient/dspAROpenItems.cpp
```diff
@@ -643,40 +643,78 @@ void dspAROpenItems::sVoidCreditMemo()
...
+                  " AND cmitem_updateinv "
+                  " AND isControlledItemsite(cmitem_itemsite_id) "
+                  "ORDER BY cmitem_id;");
+  cmitems.bindValue(":cmheadId", list()->currentItem()->id("docnumber"));
+  cmitems.exec();
+  while (cmitems.next())
+  {
+    if (distributeInventory::SeriesCreate(cmitems.value("itemsite_id").toInt(), 
+      cmitems.value("qty").toDouble(), "CM", "RS", cmitems.value("cmitem_id").toInt(), itemlocSeries) < 0)
+      return;
```
ChristopherCotnoir: Should cleanup here.

guiclient/dspAROpenItems.cpp
```diff
@@ -799,40 +837,80 @@ void dspAROpenItems::sVoidInvoiceDetails()
...
+    return;
+
+  XSqlQuery cleanup; // Stage cleanup function to be called on error
+  cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+  cleanup.bindValue(":itemlocSeries", itemlocSeries);
+  
+  // Cycle through credit memo items that are controlled and have qty returned, create an itemlocdist record for each
+  XSqlQuery invcitems;
+  invcitems.prepare("SELECT invcitem_id, itemsite_id, item_number, "
+                    " (invcitem_billed * invcitem_qty_invuomratio) * -1 AS qty "
```
ChristopherCotnoir: SH has negative sense, so don't multiply by -1 here.

guiclient/dspAROpenItems.cpp
```diff
@@ -799,40 +837,80 @@ void dspAROpenItems::sVoidInvoiceDetails()
...
+                    "WHERE invchead_id=:invcheadId "
+                    " AND isControlledItemsite(itemsite_id) "
+                    "ORDER BY invcitem_id;");
+  invcitems.bindValue(":invcheadId", list()->currentItem()->id("docnumber"));
+  invcitems.exec();
+  while (invcitems.next())
+  {
+    if (distributeInventory::SeriesCreate(invcitems.value("itemsite_id").toInt(), 
+      invcitems.value("qty").toDouble(), "IN", "SH", invcitems.value("invcitem_id").toInt(), itemlocSeries) < 0)
+    return;
```
ChristopherCotnoir: Should cleanup here.

guiclient/dspAROpenItems.cpp
```diff
@@ -1087,53 +1165,81 @@ void dspAROpenItems::sPostCreditMemo()
...
+                    " JOIN itemsite ON itemsite_id=cmitem_itemsite_id "
+                    " JOIN item ON item_id=itemsite_item_id "
+                    " JOIN costcat ON costcat_id=itemsite_costcat_id "
+                    "WHERE cmitem_qtyreturned <> 0 "
+                    " AND cmitem_updateinv "
+                    " AND cmhead_id=:cmheadId "
+                    " AND isControlledItemsite(itemsite_id) "
+                    " AND itemsite_costmethod != 'J' "
+                    "GROUP BY cmitem_id, itemsite_id, item_number "
+                    "ORDER BY cmitem_id;");
```
ChristopherCotnoir: Don't put cmitem_id in the group by. Order by itemsite_id.

guiclient/dspAROpenItems.cpp
```diff
@@ -1087,53 +1165,81 @@ void dspAROpenItems::sPostCreditMemo()
...
+  cmitems.bindValue(":cmheadId", id);
+  cmitems.exec();
+  while (cmitems.next())
+  {
+    if (distributeInventory::SeriesCreate(cmitems.value("itemsite_id").toInt(), 
+      cmitems.value("qty").toDouble(), "CM", "RS", cmitems.value("cmitem_id").toInt(), itemlocSeries) < 0)
+      return;
```
ChristopherCotnoir: Should call cleanup here.

guiclient/dspAROpenItems.cpp
```diff
@@ -1277,18 +1377,69 @@ void dspAROpenItems::sPostInvoice()
...
+                  " AND isControlledItemsite(itemsite_id) "
+                  " AND itemsite_costmethod != 'J' "
+                  "ORDER BY invcitem_id;");
+  invcitems.bindValue(":invcheadId", list()->currentItem()->id("docnumber"));
+  invcitems.exec();
+  while (invcitems.next())
+  {
+    if (distributeInventory::SeriesCreate(invcitems.value("itemsite_id").toInt(), 
+      invcitems.value("qty").toDouble(), "IN", "SH", invcitems.value("invcitem_id").toInt(), itemlocSeries) < 0)
+      return;
```
ChristopherCotnoir: Should cleanup here.

guiclient/issueWoMaterialBatch.cpp
```diff
@@ -176,16 +176,13 @@ void issueWoMaterialBatch::sIssue()
...
-  bool trynext = true;
   int succeeded = 0;
   QList<QString> failedItems;
   QList<QString> errors;
   while(items.next())
-  {
-    // Previous error and user did not want to continue posting remaining invoices. Do nothing for the rest of the loop.
-    if (!trynext)
+  { 
+    if (items.value("qty").toDouble() <= 0)
```
ChristopherCotnoir: This is a change in behavior. That should be handled in a different PR if it is really needed.

guiclient/issueWoMaterialItem.cpp
```diff
@@ -267,19 +262,15 @@ void issueWoMaterialItem::sIssue()
       lsdetail.exec();
       if (lsdetail.lastError().type() != QSqlError::NoError)
       {
-        rollback.exec();
```
ChristopherCotnoir: We need to be able to rollback the issueWoMaterial call if this INSERT fails.

guiclient/postCreditMemos.cpp
```diff
@@ -87,78 +88,179 @@ void postCreditMemos::sPost()
...
+                      "FROM cmhead JOIN cmitem ON cmitem_cmhead_id=cmhead_id "
+                      " JOIN itemsite ON itemsite_id=cmitem_itemsite_id "
+                      " JOIN item ON item_id=itemsite_item_id "
+                      " JOIN costcat ON costcat_id=itemsite_costcat_id "
+                      "WHERE cmhead_id=:cmheadId "
+                      " AND cmitem_qtyreturned <> 0 "
+                      " AND cmitem_updateinv "
+                      " AND isControlledItemsite(itemsite_id) "
+                      " AND itemsite_costmethod != 'J' "
+                      "GROUP BY cmitem_id, itemsite_id, item_number "
```
ChristopherCotnoir: Same problem as elsewhere. Don't GROUP BY cmitem_id.

guiclient/postCreditMemos.cpp
```diff
@@ -87,78 +88,179 @@ void postCreditMemos::sPost()
...
+        cmitemFail = true;
+        break;
+      }
+
+      hasControlledItems = true;
     }
-    else if (distributeInventory::SeriesAdjust(postPost.value("result").toInt(), this) == XDialog::Rejected)
+
+    // Don't continue on this credit memo because there was an issue with one of it's line items
+    if (cmitemFail)
```
ChristopherCotnoir: Should call cleanup here before continuing.

guiclient/postCreditMemos.cpp
```diff
@@ -87,78 +88,179 @@ void postCreditMemos::sPost()
...
+    // Distribute detail for the records created above
+    if (hasControlledItems && distributeInventory::SeriesAdjust(itemlocSeries, this, QString(), QDate(),
+      QDate(), true) == XDialog::Rejected)
     {
-      rollback.exec();
-      QMessageBox::information( this, tr("Post Sales Credits"), tr("Transaction Canceled") );
-      return;
+      cleanup.exec();
+      // If it's not the last credit memo, ask user if they want to continue
+      if (creditMemos.at() != creditMemos.size())
```
ChristopherCotnoir: Should be size() -1

guiclient/postCreditMemos.cpp
```diff
@@ -87,78 +88,179 @@ void postCreditMemos::sPost()
...
+          continue;
+        }
+        else
+        {
+          failedItems.append(creditMemoNumber);
+          errors.append(tr("Detail Distribution Cancelled"));
+          break;
+        }
+      }
+
```
ChristopherCotnoir: There should also be appending to lists and continue here, for if it is the last item. We still want to skip the rest of the operations and still want to report the error.

guiclient/postMiscProduction.cpp
```diff
@@ -91,71 +92,57 @@ void postMiscProduction::sPost()
   if (!okToPost())
     return;
 
-  XSqlQuery rollback;
-  rollback.prepare("ROLLBACK;");
+  if (!createwo())
```
ChristopherCotnoir: This needs to be rolled back if something later fails. It should stay in the transaction block.

guiclient/postMiscProduction.cpp
```diff
@@ -91,71 +92,57 @@ void postMiscProduction::sPost()
...
+    return;
 
-  postPost.exec("BEGIN;");	// because of possible lot, serial, or location distribution cancelations
+  int itemlocSeries = handleSeriesAdjustBeforePost();
+  int twItemlocSeries = _immediateTransfer->isChecked() ? handleTransferSeriesAdjustBeforePost() : 0;
```
ChristopherCotnoir: returntool() should have itemlocdist records created for it as well, but they seem to be missing.

guiclient/postMiscProduction.cpp
```diff
@@ -91,71 +92,57 @@ void postMiscProduction::sPost()
...
-    if (_immediateTransfer->isChecked())
-    {
-      if (!transfer())
-      {
-        rollback.exec();
-        return;
-      }
-    }
+    cleanup.exec();
+    closewo();
```
ChristopherCotnoir: Remove this, put createwo() in the transaction block instead. This isn't sufficient cleanup to truly erase changes.

guiclient/postMiscProduction.cpp
```diff
@@ -412,3 +386,196 @@ void postMiscProduction::clear()
...
+  XSqlQuery parentItemlocdist;
+  bool hasControlledBackflushItems = false;
+
+  // Stage cleanup function to be called on error
+  XSqlQuery cleanup;
+  cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+
+  // Get series
+  int itemlocSeries = distributeInventory::SeriesCreate(0, 0, QString(), QString());
+  if (itemlocSeries <= 0)
```
ChristopherCotnoir: Should give the user an error before returning.

guiclient/postMiscProduction.cpp
```diff
@@ -412,3 +386,196 @@ void postMiscProduction::clear()
...
+                           "FROM ( "
+                           "  SELECT matlitem.item_number, matlitem.item_id, womatl_itemsite_id, womatl_uom_id, womatl_id, "
+                           "    CASE WHEN :qty > 0 THEN "
+                           "      noNeg(((womatl_qtyfxd + ((roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper)) * (1 + womatl_scrap)) - "
+                           "        (womatl_qtyiss + "
+                           "          CASE WHEN (womatl_qtywipscrap >  ((womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap)) "
+                           "          THEN (womatl_qtyfxd + (roundQty(woitem.item_fractional, :qty) + wo_qtyrcv) * womatl_qtyper) * womatl_scrap "
+                           "          ELSE womatl_qtywipscrap "
+                           "          END)) "
+                           "    ELSE (roundQty(woitem.item_fractional, :qty) * womatl_qtyper) "
```
ChristopherCotnoir: Should be * -1.

guiclient/postMiscProduction.cpp
```diff
@@ -412,3 +386,196 @@ void postMiscProduction::clear()
...
+    backflushItems.bindValue(":qty", _qty);
+    backflushItems.exec();
+    while (backflushItems.next())
+    {
+      hasControlledBackflushItems = true;
+
+      int result = distributeInventory::SeriesCreate(backflushItems.value("womatl_itemsite_id").toInt(), backflushItems.value("qty").toDouble(), "WO", "IM", _woid, itemlocSeries);
+      if (result != itemlocSeries)
+      {
+        cleanup.exec();
```
ChristopherCotnoir: Raise an error here.

guiclient/postMiscProduction.cpp
```diff
@@ -412,3 +386,196 @@ void postMiscProduction::clear()
...
+      }
+    }
+  }
+
+  // Distribute detail
+  if (_controlled || hasControlledBackflushItems)
+  {
+    if (_controlled)
+    {
+      int result = distributeInventory::SeriesCreate(_itemsiteid, _qty, "WO", "RM", _woid, itemlocSeries);
```
ChristopherCotnoir: _qty should be roundQty(item_fractional, _qty), with the item from wo.

guiclient/postMiscProduction.cpp
```diff
@@ -367,26 +341,26 @@ bool postMiscProduction::closewo()
...
+  posttransfer.prepare("SELECT interWarehouseTransfer( :item_id, :from_warehous_id, :to_warehous_id,"
+                       "  :qty, 'W', COALESCE(:documentNumber, formatWoNumber(:wo_id)), 'Transfer from Misc. Production Posting', "
+                       "  :itemlocSeries, "
+                       "  now(), "
+                       "  TRUE, "
+                       "  true ) AS result;");
   posttransfer.bindValue(":item_id", _item->id());
   posttransfer.bindValue(":from_warehous_id", _warehouse->id());
   posttransfer.bindValue(":to_warehous_id", _transferWarehouse->id());
   posttransfer.bindValue(":qty", _qty * _sense);
```
ChristopherCotnoir: You removed the code that sets _sense, and given how you handle itemlocdist creation below, I presume you don't want to do that.

guiclient/postMiscProduction.cpp
```diff
@@ -412,3 +386,196 @@ void postMiscProduction::clear()
...
+                               tr("Error looking up itemlocdist info. Expected itemlocdist_id to be > 0, not %1 for itemlocdist_series %2")
+                               .arg(itemlocdistId).arg(twItemlocSeries) );
+        return -1;   
+      }
+
+      // Create the TO wh itemlocdist record
+      int result = distributeInventory::SeriesCreate(toWhItemsiteId,
+        // if disassembly, _qty is positive, make it negative because interWarehouseTransfer function will create invhist record with negative qty for TO wh
+        _disassembly->isChecked() ? _qty * -1 : _qty, 
+        "W", "TW", _woid, twItemlocSeries, itemlocdistId);
```
ChristopherCotnoir: This should be outside the if block, otherwise it doesn't occur if the source itemsite is not controlled even if the destination itemsite is. The if _controlled && toWhControlled check should only be used to determine if ItemlocdistId is passed, the series should be created regardless.

guiclient/returnWoMaterialBatch.cpp
```diff
@@ -99,59 +99,131 @@ void returnWoMaterialBatch::sReturn()
...
         XSqlQuery items;
-        items.prepare("SELECT womatl_id,"
-                      "       CASE WHEN wo_qtyord >= 0 THEN"
-                      "         womatl_qtyiss"
-                      "       ELSE"
-                      "         ((womatl_qtyreq - womatl_qtyiss) * -1)"
-                      "       END AS qty"
-                      "  FROM wo, womatl, itemsite"
+        items.prepare("SELECT womatl_id, womatl_uom_id, womatl_qtyreq, womatl_qtyiss, "
+                      " COALESCE(itemuomtouom(item_id, NULL, womatl_uom_id, "
```
ChristopherCotnoir: Should be womatl_uom_id, NULL, not the other way around.

guiclient/returnWoMaterialBatch.cpp
```diff
@@ -99,59 +99,131 @@ void returnWoMaterialBatch::sReturn()
...
+
+          // Stage distribution cleanup function to be called on error
+          XSqlQuery cleanup;
+          cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+
+          // Get the parent series id
+          itemlocSeries = distributeInventory::SeriesCreate(items.value("itemsite_id").toInt(),
+            items.value("qty").toDouble(), "WO", "IM", _wo->id());
+          if (itemlocSeries <= 0)
+            return;
```
ChristopherCotnoir: Should append to lists and continue.

guiclient/returnWoMaterialBatch.cpp
```diff
@@ -99,59 +99,131 @@ void returnWoMaterialBatch::sReturn()
...
+
+          cleanup.bindValue(":itemlocSeries", itemlocSeries);
+
+          // If controlled and backflush item has relevant qty for returnWoMaterial
+          if (items.value("controlled").toBool() && 
+             (items.value("womatl_qtyreq").toDouble() >= 0 ? 
+              items.value("womatl_qtyiss").toDouble() >= items.value("qty").toDouble() : 
+              items.value("womatl_qtyiss").toDouble() <= items.value("qty").toDouble()))
```
ChristopherCotnoir: This check should be used when determining whether to create an itemlocdist entry in the first place.

guiclient/returnWoMaterialBatch.cpp
```diff
@@ -99,59 +99,131 @@ void returnWoMaterialBatch::sReturn()
...
+                .arg(returnReturn.lastError().text()));
+              continue;
             }
           }
           else
           {
-            rollback.exec();
             ErrorReporter::error(QtCriticalMsg, this, tr("Error Returning Work Order Material Batch, W/O ID #%1")
                                  .arg(_wo->id()),returnReturn, __FILE__, __LINE__);\
             return;
```
ChristopherCotnoir: Should continue/append here.

guiclient/scrapWoMaterialFromWIP.cpp
```diff
@@ -205,67 +219,62 @@ void scrapWoMaterialFromWIP::sScrap()
         }
       }
     }
+    else // If not Post Production, handle detail distribution.
+    {      
+      // Generate the series and itemlocdist record (if controlled)
+      itemlocSeries = distributeInventory::SeriesCreate(itemsite.value("wo_itemsite_id").toInt(),
+        _topLevelQty->toDouble() * -1, "SI", "SI", _wo->id());
```
ChristopherCotnoir: Should skip this if the itemsite is job costed, because invScrap will return early in that case and it isn't caught by isControlledItemsite (because we still use lot/serial with job costing for things like issueToShipping).

guiclient/transferOrder.cpp
```diff
@@ -1857,31 +1858,73 @@ void transferOrder::sIssueLineBalance()
...
+          itemsite.lastError().databaseText(), __FILE__, __LINE__);
+        return;
+      }
+      
+      // Stage cleanup query
+      XSqlQuery cleanup;
+      cleanup.prepare("SELECT deleteitemlocseries(:itemlocSeries, TRUE);");
+
+      // Generate the series and itemlocdist record (if controlled)
+      int itemlocSeries = distributeInventory::SeriesCreate(itemsiteId, balance * -1, "TO", "IS", toitem->id());
```
ChristopherCotnoir: s/IS/SH/

guiclient/transferOrder.cpp
```diff
@@ -1857,31 +1858,73 @@ void transferOrder::sIssueLineBalance()
...
+      if (itemlocSeries <= 0)
+        return;
```
ChristopherCotnoir: Since we're in a loop here, let's do continue here and elsewhere and proper error reporting to fix this latent bug where some items will not be issued if they come after a failure.

guiclient/transferOrder.cpp
```diff
@@ -1857,31 +1858,73 @@ void transferOrder::sIssueLineBalance()
...
+      cleanup.bindValue(":itemlocSeries", itemlocSeries);
+
```
ChristopherCotnoir: Skip all the itemlocdist stuff if balance == 0, since that's what it did before when calling issueLineBalanceToShipping.

guiclient/transferOrder.cpp
```diff
@@ -1857,31 +1858,73 @@ void transferOrder::sIssueLineBalance()
...
         }
-        if (distributeInventory::SeriesAdjust(transferIssueLineBalance.value("result").toInt(), this) == XDialog::Rejected)
-        {
-          rollback.exec();
-          QMessageBox::information( this, tr("Issue to Shipping"), tr("Transaction Canceled") );
-          return;
-        }
 
         transferIssueLineBalance.exec("COMMIT;");
       }
```
ChristopherCotnoir: cleanup below.

guiclient/transferOrder.cpp
```diff
@@ -879,6 +879,7 @@ void transferOrder::sNew()
   params.append("orderDate",	_orderDate->date());
   params.append("taxzone_id",	_taxzone->id());
   params.append("curr_id",	_freightCurrency->id());
+  params.append("status", _statusTypes[_status->currentIndex()]);
```
ChristopherCotnoir: Gil says for this change, to prompt the user to open the Transfer Order if it is closed. So if the status is closed, ask the user if they want to open the transfer order before adding a line item, and if they agree set the transfer order status to open if it has been saved already and change the combobox accordingly (before passing it, so transferOrderItem gets passed "O"), and if they disagree, do nothing.

guiclient/transferTrans.cpp
```diff
@@ -205,50 +204,138 @@ void transferTrans::sPost()
...
+  if (itemsiteInfo.first())
+  {
+    fromItemsiteControlled = itemsiteInfo.value("from_controlled").toBool();
+    toItemsiteControlled = itemsiteInfo.value("to_controlled").toBool();
+    fromItemsiteId = itemsiteInfo.value("from_itemsite_id").toInt();
+    toItemsiteId = itemsiteInfo.value("to_itemsite_id").toInt();
+  }
+  else
+  {
+    cleanup.exec();
```
ChristopherCotnoir: This will unnecessarily throw another error because :itemlocSeries hasn't been bound yet. Don't cleanup here, there is nothing to cleanup yet.

guiclient/transferTrans.cpp
```diff
@@ -205,50 +204,138 @@ void transferTrans::sPost()
...
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Finding Itemlocdist Info"),
+        itemlocdist, __FILE__, __LINE__);
+      return;
+    }
+  }
+
+  if (toItemsiteControlled)
+  {
+    // FROM itemsite itemlocdist_id is valid, proceed to create itemlocdist record for TO itemsite
+    int result = distributeInventory::SeriesCreate(toItemsiteId, _qty->toDouble(), "TW", "TR", 0, itemlocSeries, itemlocdistId);
```
ChristopherCotnoir: Should be TW instead of TR.

guiclient/correctProductionPosting.cpp
```diff
@@ -140,49 +140,101 @@ void correctProductionPosting::clear()
...
+    return;
+  }
 
-  correctCorrect.exec("BEGIN;");	// handle cancel of lot, serial, or loc distribution
-  correctCorrect.prepare("SELECT correctProduction(:wo_id, :qty, :backflushMaterials, 0, :date)"
-            "        AS result;");
+  itemlocSeries = distributeInventory::SeriesCreate(postInfo.value("wo_itemsite_id").toInt(),
+    postInfo.value("parent_qty").toDouble() * -1, "WO", "RM", _wo->id());
+  if (itemlocSeries <= 0)
+    return;
```
ChristopherCotnoir: Give an error message here and elsewhere so the user knows it failed.

guiclient/transferTrans.cpp
```diff
@@ -205,50 +204,138 @@ void transferTrans::sPost()
...
+    if (result != itemlocSeries)
+      return;
```
ChristopherCotnoir: cleanup here.

guiclient/unpostedCreditMemos.cpp
```diff
@@ -188,7 +187,6 @@ void unpostedCreditMemos::sPost()
 		  "WHERE cmhead_id=:cmhead_id;");
 
   QList<XTreeWidgetItem*> selected = _cmhead->selectedItems();
-  QList<XTreeWidgetItem*> triedToClosed;
```
ChristopherCotnoir: We want to keep this functionality. Do what we did for unpostedPoReceipts.cpp.

guiclient/unpostedCreditMemos.cpp
```diff
@@ -210,80 +208,177 @@ void unpostedCreditMemos::sPost()
...
-  do {
-    for (int i = 0; i < selected.size(); i++)
+  int journalNumber = unpostedPost.value("result").toInt();
+
+
+  int succeeded = 0;
+  QList<QString> failedMemos;
+  QList<QString> errors;
+  for (int i = 0; i < selected.size(); i++)
+  {
```
ChristopherCotnoir: continue here if !checkSitePrivs(((XTreeWidgetItem*)(selected[i]))->id()). That check should not be removed.

guiclient/unpostedCreditMemos.cpp
```diff
@@ -210,80 +208,177 @@ void unpostedCreditMemos::sPost()
...
+                      " JOIN itemsite ON itemsite_id=cmitem_itemsite_id "
+                      " JOIN item ON item_id=itemsite_item_id "
+                      " JOIN costcat ON costcat_id=itemsite_costcat_id "
+                      "WHERE cmhead_id=:cmheadId "
+                      " AND cmitem_qtyreturned <> 0 "
+                      " AND cmitem_updateinv "
+                      " AND isControlledItemsite(itemsite_id) "
+                      " AND itemsite_costmethod != 'J' "
+                      "GROUP BY cmitem_id, itemsite_id, item_number "
+                      "ORDER BY cmitem_id;");
```
ChristopherCotnoir: Do not GROUP BY cmitem_id, ORDER BY itemsite_id instead.

guiclient/unpostedCreditMemos.cpp
```diff
@@ -210,80 +208,177 @@ void unpostedCreditMemos::sPost()
...
+    XSqlQuery memo;
+    memo.prepare("SELECT cmhead_number FROM cmhead WHERE cmhead_id=:cmheadId;");
+    memo.bindValue(":cmheadId", memoId);
+    memo.exec();
+    if (!memo.first())
     {
-      if (checkSitePrivs(((XTreeWidgetItem*)(selected[i]))->id()))
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Journal Number"),
+        memo.lastError().databaseText(), __FILE__, __LINE__);
+      return;
```
ChristopherCotnoir: continue here.

guiclient/unpostedInvoices.cpp
```diff
@@ -210,141 +210,208 @@ void unpostedInvoices::sPost()
...
+              " invchead_invcnumber "
+              "FROM invchead "
+              "WHERE invchead_id=:invchead_id;");
 
   XSqlQuery setDate;
   setDate.prepare("UPDATE invchead SET invchead_gldistdate=:distdate "
 		  "WHERE invchead_id=:invchead_id;");
 
   QList<XTreeWidgetItem*> selected = list()->selectedItems();
-  QList<XTreeWidgetItem*> triedToClosed;
```
ChristopherCotnoir: Same issue as unpostedCreditMemos, keep this functionality.

guiclient/unpostedInvoices.cpp
```diff
@@ -210,141 +210,208 @@ void unpostedInvoices::sPost()
...
   }
 
-  bool tryagain = false;
-  do {
-    for (int i = 0; i < selected.size(); i++)
+  int succeeded = 0;
+  QList<QString> failedInvoices;
+  QList<QString> errors;
+  for (int i = 0; i < selected.size(); i++)
+  {
```
ChristopherCotnoir: continue if !checkSitePrivs(((XTreeWidgetItem*)(selected[i]))->id())

guiclient/unpostedInvoices.cpp
```diff
@@ -210,141 +210,208 @@ void unpostedInvoices::sPost()
...
+      invoiceNumber = sum.value("invchead_invcnumber").toString();
+
+      if (sum.value("subtotal").toDouble() == 0 && QMessageBox::question(this, tr("Invoice Has Value 0"),
+              tr("Invoice #%1 has a total value of 0.\n"
+               "Would you like to post it anyway?")
+          .arg(selected[i]->text(0)),
+        QMessageBox::Yes,
+        QMessageBox::No | QMessageBox::Default)
+      == QMessageBox::No)
+        continue;
```
ChristopherCotnoir: Append to lists here and below.

guiclient/unpostedInvoices.cpp
```diff
@@ -210,141 +210,208 @@ void unpostedInvoices::sPost()
...
+    }
+    else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Invoice Information"),
+                                    sum, __FILE__, __LINE__))
+    {
```
ChristopherCotnoir: Append to lists here instead of raising an error.

guiclient/workOrder.cpp
```diff
@@ -1312,42 +1312,169 @@ void workOrder::sReturnMatlBatch()
...
+                    " CASE WHEN wo_qtyord >= 0 THEN womatl_qtyiss "
+                    "   ELSE ((womatl_qtyreq - womatl_qtyiss) * -1) "
+                    " END AS qty, "
+                    " itemsite_id, itemsite_item_id, item_number, "
+                    " isControlledItemsite(itemsite_id) AS controlled "
+                    "FROM wo, womatl, itemsite, item "
+                    "WHERE wo_id=womatl_wo_id "
+                    " AND womatl_itemsite_id=itemsite_id "
+                    " AND itemsite_item_id=item_id "
+                    " AND womatl_wo_id=:wo_id "
```
ChristopherCotnoir: Should also add AND (wo_qtyord < 0) OR (womatl_issuemethod IN ('S','M'))

guiclient/workOrder.cpp
```diff
@@ -1312,42 +1312,169 @@ void workOrder::sReturnMatlBatch()
...
+        if (items.value("wo_qtyord").toDouble() > 0 && (items.value("womatl_issuemethod").toString() == "L"))
         {
-          rollback.exec();
-          ErrorReporter::error(QtCriticalMsg, this, tr("Error Returning Work Order Materials: W/O #%1")
-                                                       .arg(_woIndentedList->id()),
-                                                       workReturnMatlBatch, __FILE__, __LINE__);
-          return;
+          failedItems.append(items.value("item_number").toString());
+          errors.append(tr("Skipping Item Number %1 because it's Issue Method is Pull.")
+              .arg(items.value("item_number").toString()));
```
ChristopherCotnoir: This isn't an 'error', and shouldn't be treated as such.

guiclient/workOrder.cpp
```diff
@@ -1312,42 +1312,169 @@ void workOrder::sReturnMatlBatch()
...
+        // If controlled and backflush item has relevant qty for returnWoMaterial
+        if (items.value("controlled").toBool() && 
+           (items.value("womatl_qtyreq").toDouble() >= 0 ? 
+            items.value("womatl_qtyiss").toDouble() >= items.value("qty").toDouble() : 
+            items.value("womatl_qtyiss").toDouble() <= items.value("qty").toDouble()))
+        {
+          // Create the parent itemlocdist record for each line item requiring distribution and call distributeInventory::seriesAdjust
+          XSqlQuery parentItemlocdist;
+          parentItemlocdist.prepare("SELECT createitemlocdistparent(:itemsite_id, "
+                                    " COALESCE(itemuomtouom(:itemsite_item_id, NULL, :womatl_uom_id, :qty), :qty), "
```
ChristopherCotnoir: NULL and :womatl_uom_id are in the wrong order.

guiclient/workOrder.cpp
```diff
@@ -1358,79 +1485,189 @@ void workOrder::sIssueMatlBatch()
...
-                "  FROM womatl, itemsite, item, whsinfo "
-                " WHERE ( (womatl_itemsite_id=itemsite_id) "
+                "  FROM wo, womatl, itemsite, item, whsinfo "
+                " WHERE ( (wo_id=womatl_wo_id) "
+                "   AND (womatl_itemsite_id=itemsite_id) "
                 "   AND (itemsite_item_id=item_id) "
                 "   AND (itemsite_warehous_id=warehous_id) "
                 "   AND (NOT ((item_type = 'R') OR (itemsite_controlmethod = 'N'))) "
-                "   AND ((itemsite_controlmethod IN ('L', 'S')) OR (itemsite_loccntrl)) "
-                "   AND (womatl_issuemethod IN ('S', 'M')) "
```
ChristopherCotnoir: This condition should stay, and the check for item_type and itemsite_controlmethod should go.

guiclient/workOrder.cpp
```diff
@@ -1358,79 +1485,189 @@ void workOrder::sIssueMatlBatch()
...
   {
-    if(!(issue.value("isqtyavail").toBool()))
+    if (items.value("womatl_issuemethod").toString() == "L")
     {
-      QMessageBox::critical(this, tr("Insufficient Inventory"),
-        tr("Item Number %1 in Site %2 is a Multiple Location or\n"
+      failedItems.append(items.value("item_number").toString());
+      errors.append(tr("Skipping Item Number %1 because it's Issue Method is Pull.")
+          .arg(items.value("item_number").toString()));
+      continue;
```
ChristopherCotnoir: Don't throw an error for this.

guiclient/distributeInventory.cpp
jcarlin: It was the easiest way to display relevant information in the `_order` field. There are many transactions that don't have a document number, IMO it is helpful display as much relevant info. as possible for those cases, even if it's just the order/trans type. Also, it's very common that we let logic (i.e. order types handled by formatOrderLineItemNumber) remain in the stored procedure, I don't agree that we need to check that it's an order type handled by the stored proc. before calling it.

guiclient/correctProductionPosting.cpp
jcarlin: Transformation?

guiclient/issueWoMaterialBatch.cpp
jcarlin: It's a correction from the last merged changes.

guiclient/correctProductionPosting.cpp
jcarlin: distributeInventory::SeriesCreate displays error message before returning -1.

guiclient/transferOrder.cpp
jcarlin: I don't believe it's possible to open a closed work order, only Open and Unreleased. I noticed when creating a TO and setting it's status to Open then adding line items, the line items' status were Unreleased. This change is part of that fix to simply match the new line item statuses to the header status. Should I strip all of this out, file an incident and submit a separate pr?

guiclient/workOrder.cpp
jcarlin: I included it because there's a scenario where nothing happens but no message to let the end user know, it's confusing. I thought it would be more transparent to include them in the 'error' messages.

guiclient/workOrder.cpp
jcarlin: Why not? https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/workOrder.cpp#L1381

guiclient/workOrder.cpp
gilmoskowitz: In that case it might we worth keeping 2 lists - hard errors and warnings. They could be presented together at the end:

**Errors:**
- A
- B

**Warnings:**
- C

Make sure not to show the header ("Errors" or "Warnings") if the pertinent list is empty.

jcarlin: I went down that route but stopped because there are many other classes where this could be added and it's too much work at this point.

guiclient/correctProductionPosting.cpp
jcarlin: Done (below).

guiclient/postCreditMemos.cpp
jcarlin: Done (below).

guiclient/workOrder.cpp
jcarlin: Done (below).

guiclient/postMiscProduction.cpp
jcarlin: The series is created on L509.

guiclient/postMiscProduction.cpp
jcarlin: `SeriesCreate` handles the errors.

guiclient/postMiscProduction.cpp
jcarlin: Handled in SeriesCreate.

guiclient/postMiscProduction.cpp
jcarlin: I don't know how to reproduce this feature of controlled tool items needing detail distribution after posting.

jcarlin: I reverted the changes to returntool. This will lock for whatever scenario it is that actually requires distribution.

davidbeauchamp:
jenkins - retest this please

guiclient/issueWoMaterialBatch.cpp
jcarlin: I copied https://github.com/xtuple/xtuple/blob/4_11_x/foundation-database/public/functions/issuewomaterial.sql#L130 here, except that I added `=`, and looking at this again, we allow calls to `issueWoMaterial.sql`/`postInvTrans.sql` with 0 qty??

gilmoskowitz: Except that `trynext` is set in response to user input. I agree with Chris. As for qty 0, let's not worry about that right now.

guiclient/issueWoMaterialItem.cpp
gilmoskowitz: agreed

guiclient/workOrder.cpp
gilmoskowitz: I agree with Chris, then - this message shouldn't be displayed because the system is behaving as configured/desired. However, the point about confusion is valid. Please create a bug that we should add warning list.

guiclient/workOrder.cpp
gilmoskowitz: Failing to issue a Pull item is desired behavior - the user has made that choice.  The comparison with the workOrder.cpp is invalid because the condition being checked has changed.

guiclient/distributeInventory.cpp
gilmoskowitz: I agree with Julian on this one.

gilmoskowitz:
I didn't review the changes carefully, just looked at the current state of comment/response. In that light this PR is very close - probably just another couple of hours from done and ready to merge.

guiclient/issueWoMaterialBatch.cpp
jcarlin: Setting `trynext` was obsolete because of the `break` that followed: https://github.com/xtuple/qt-client/pull/1596/files/c9d556330838d88ac35174df57bbd3f3f03e33ea#diff-b02cee9e9709ab76620b6879b112c827R241. I've removed the continue on qty <= 0.

guiclient/issueWoMaterialItem.cpp
jcarlin: Done https://github.com/xtuple/qt-client/pull/1596/files#diff-dabe31e201b12725fe1e6555057d4322R232/.

guiclient/postMiscProduction.cpp
jcarlin: For `SeriesCreate` qty I'm now using `itemlocdist_qty` from above which includes `* -1`.

guiclient/postMiscProduction.cpp
jcarlin: I added a deletewo() method instead.

PR #1600, opened 2017-05-19
issue #30126 Avoid foreign key error, add error check after save

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3036.

PR #1605, opened 2017-05-28
Issue #30008: Incorrect username displaying when user name overlaps

andersdd:
Been like this for 7+ years.  Could not find a reason why we were using a broad regexp on username.  Makes more sense to have discrete username matching.

### DIFF
--- a/widgets/usernameCluster.cpp
+++ b/widgets/usernameCluster.cpp
@@ -105,7 +105,7 @@ void UsernameLineEdit::setUsername(const QString & pUsername)
   XSqlQuery query;
   QString sql("SELECT usr_id, usr_username AS number "
                 "  FROM usr"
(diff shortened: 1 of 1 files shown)


widgets/usernameCluster.cpp
```diff
@@ -105,15 +105,15 @@ void UsernameLineEdit::setUsername(const QString & pUsername)
   XSqlQuery query;
   QString sql("SELECT usr_id, usr_username AS number "
                 "  FROM usr"
-                " WHERE ((usr_username ~* :username) ");
+                " WHERE ((usr_username = :username) ");
```
gilmoskowitz: The `~*` gives case-insensitive search and it would be nice to preserve that. How 'bout `upper(usr_username) = upper(:username)` instead?

PR #1607, opened 2017-05-30
Issue #30111: Lot selection on inventory history report

andersdd:
Lot is capitalised when created so applying this to filter

### DIFF
--- a/guiclient/displays/dspDetailedInventoryHistoryByLotSerial.cpp
+++ b/guiclient/displays/dspDetailedInventoryHistoryByLotSerial.cpp
@@ -226,7 +226,7 @@ bool dspDetailedInventoryHistoryByLotSerial::setParams(ParameterList &params)
   else
   {
     params.append("pattern", "t");
-    params.append("lotSerial", _lotSerialPattern->text().trimmed());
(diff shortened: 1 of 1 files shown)


guiclient/displays/dspDetailedInventoryHistoryByLotSerial.cpp
```diff
@@ -226,7 +226,7 @@ bool dspDetailedInventoryHistoryByLotSerial::setParams(ParameterList &params)
   else
   {
     params.append("pattern", "t");
-    params.append("lotSerial", _lotSerialPattern->text().trimmed());
+    params.append("lotSerial", _lotSerialPattern->text().trimmed().toUpper());
```
gilmoskowitz: What's the advantage of this solution vs. making the [ls_number search](https://github.com/xtuple/private-extensions/blob/4_11_x/source/inventory/foundation-database/public/functions/lshist.sql#L73) case-insensitive?

andersdd: No advantage/disadvantage, however when creating a lot/serial value then system always uppercases the code so the above logic will always work (until someone removes that forced uppercase).

PR #1617, opened 2017-06-13
issue #30291 - fix paren mismatch, convert to mql to reduce future risk

gilmoskowitz:


andersdd:
+1 Merge when tests pass

PR #1619, opened 2017-06-16
issue #30138 - fix inheritance so encrypted metrics save correctly

gilmoskowitz:
replaces #1618

### DIFF
--- a/common/metrics.h
+++ b/common/metrics.h
@@ -36,27 +36,27 @@ class Parameters : public QObject
     Parameters(QObject * parent = 0);
     virtual ~Parameters() {};
 
-    void load();
+    virtual void load();
 
-    QString value(const char *);
-    bool    boolean(const char *);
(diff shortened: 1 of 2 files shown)


gilmoskowitz:
@bendiy - the key was your statement:

> When Metricsenc::set() is called, it's calling the parent Parameters::set(), which then calls Parameters::_set(), not the child's Parametersenc::_set() because it's protected and the parent doesn't have access to it or the _key.

That means it's an inheritance bug. The point of `protected` is to allow classes in the same branch of the hierarchy to see these members but hide them from others. The point of `virtual` is to allow subclasses to override methods in their parents and have the parents behave appropriately.

bendiy:
@gilmoskowitz Thanks. I wasn't sure when the child overrides the parent's methods or the parent's method gets called instead.
Using `virtual` LGTM.

PR #1620, opened 2017-06-16
resolves issue #30337 - SO item number gray when left blank

jcarlin:


jcarlin:
Closing this, it's not the correct solution, sSaved shouldn't be called if no item has been entered.

PR #1623, opened 2017-06-21
Fixes #30138 Only try to load SSL cert extension files. Correctly create certs, dicts and ts for windows.

bendiy:


guiclient/guiclient.pro
```diff
@@ -1902,17 +1902,17 @@ include( hunspell.pri )
...
 
 macx {
-  EXTRASDIR=${DESTDIR}/xtuple.app/Contents/Resources
+  EXTRASDIR=${DESTDIR}xtuple.app/Contents/Resources
 } else {
   EXTRASDIR=${DESTDIR}
 }
 
-certificates.path = $${EXTRASDIR}/certificates
+certificates.path = ${EXTRASDIR}certificates
```
bendiy: @gilmoskowitz I don't know if this change will break MacOS or linux, but it works on Windows with `(ming32)make install`.

PR #1635, opened 2017-07-03
issue #25224 - improve perf of sales order entry

gilmoskowitz:
In particular, this change to the custCluster query improves performance of lookups
by about 10% in isolation, 20% in conjunction with xtuple/xtuple#3077


common/applock.cpp
```diff
@@ -46,75 +47,67 @@ class AppLockPrivate
...
+        XSqlQuery vq("SELECT compareversion('9.2.0') <= 0 AS isNew;");
+        if (vq.first())
+          _actPidCol = vq.value("isNew").toBool() ? "pid" : "procpid";
+        else
+          (void)ErrorReporter::error(QtCriticalMsg,
+                                     qobject_cast<QWidget*>(parent->parent()),
+                                     parent->tr("Locking Error"),
+                                     vq, __FILE__, __LINE__);
       }
+
```
ChristopherCotnoir: This is acceptable as is, but _actPidCol is no longer actually needed, since Postgres 9.2.0 and earlier is deprecated as of xtuple version 4.10.0, according to our compatibility matrix.

gilmoskowitz:
Pushed from the wrong version:-(

PR #1644, opened 2017-07-10
issue #30166 Compare with the right table in the right mode

ChristopherCotnoir:


guiclient/createLotSerial.cpp
```diff
@@ -494,17 +494,15 @@ void createLotSerial::sLotSerialSelected()
...
-                    "WHERE lsdetail_id=:itemloc_id; ");
+
+    if (!_preassigned)
+      itemloc.prepare("SELECT itemloc_expiration, itemloc_warrpurc "
+                      "FROM itemloc "
+                      "WHERE itemloc_id=:itemloc_id; ");
+    else
+      itemloc.prepare("SELECT COALESCE(itemloc_expiration, lsdetail_expiration), COALESCE(itemloc_warrpurc, lsdetail_warrpurc) "
+                      "FROM lsdetail LEFT OUTER JOIN itemloc ON (itemloc_itemsite_id=lsdetail_itemsite_id AND itemloc_ls_id=lsdetail_ls_id) "
+                      "WHERE lsdetail_id=:itemloc_id; ");
```
andersdd: Any reason for the IF (NOT true) ELSE sequence?  Why not reverse the statements and make IF true?

gilmoskowitz:
+1 after 4.11.0RC has been published

PR #1646, opened 2017-07-11
issue #22868 - remove LAST of remaining distribution locking (printMulticopyDocument.cpp)

jcarlin:
There are 6 `printMulticopyDocument.cpp` inherited classes that set `_distributeInventory = true`, which if set will `DistributeInventory::SeriesAdjust`. I added a new method `distributeInventory` to printMulticopyDocument and then call it if `_distributeInventory`.

### DIFF
--- a/guiclient/printMulticopyDocument.cpp
+++ b/guiclient/printMulticopyDocument.cpp
@@ -18,7 +18,6 @@
 #include <metasql.h>
 #include <openreports.h>
 
-#include "distributeInventory.h"
 #include "errorReporter.h"
 #include "storedProcErrorLookup.h"
 
@@ -277,14 +276,6 @@ bool printMulticopyDocument::sPostOneDoc(XSqlQuery *docq)
                              __FILE__, __LINE__);
         return false;
       }
-      if (_distributeInventory &&
-          (distributeInventory::SeriesAdjust(result, this) == XDialog::Rejected))
-      {
-        rollback.exec();
-        QMessageBox::information(this, tr("Posting Canceled"),
-                                 tr("Transaction Canceled") );
-        return false;
-      }
     }
     else if (postq.lastError().type() != QSqlError::NoError)
     {
@@ -322,6 +313,11 @@ void printMulticopyDocument::sPrint()
 
     // This indirection allows scripts to replace core behavior - 14285
     emit aboutToStart(&docinfoq);
+
+    // Distribute inventory detail
+    if (_distributeInventory && !distributeInventory(&docinfoq))
+      return;
+
     emit timeToPrintOneDoc(&docinfoq);
     emit timeToMarkOnePrinted(&docinfoq);
     emit timeToPostOneDoc(&docinfoq);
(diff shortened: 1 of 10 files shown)


guiclient/printMulticopyDocument.cpp
```diff
@@ -322,9 +335,21 @@ void printMulticopyDocument::sPrint()
...
     emit aboutToStart(&docinfoq);
+
+
+    // Distribute inventory detail
+    int itemlocSeries = 0;
+
+    if (_distributeInventory)
+    {
+      itemlocSeries = distributeInventory(&docinfoq);
+      if (itemlocSeries <= 0)
```
ChristopherCotnoir: Raise an error here and cleanup before returning.

guiclient/printMulticopyDocument.cpp
```diff
@@ -322,9 +335,21 @@ void printMulticopyDocument::sPrint()
...
+        return;
+    }
+
```
ChristopherCotnoir: Would probably make the most sense to do the above right before emitting timeToPostOneDoc, after emitting timeToMarkOnePrinted.

guiclient/printMulticopyDocument.cpp
jcarlin: The error should have already been reported to the user and if the itemlocseries is <= 0, cleanup will do nothing. I will make sure all of the new `distributeInventory` functions raise errors appropriately.
Looking at `printInvoice.cpp`: 
- https://github.com/xtuple/qt-client/pull/1646/files#diff-eb1d6ba4c75cf7bba15efed3cea1addbR94
(Error reported in `distributeInventory::SeriesCreate`)
- https://github.com/xtuple/qt-client/pull/1646/files#diff-eb1d6ba4c75cf7bba15efed3cea1addbR128
(missing error here)
- https://github.com/xtuple/qt-client/pull/1646/files#diff-eb1d6ba4c75cf7bba15efed3cea1addbR140
(QMessageBox)

guiclient/printMulticopyDocument.cpp
jcarlin: I did it this way because the user can cancel and in that case, nothing should print.

jcarlin: On that note about canceling, this `while` loop would be horrible if user changed their mind and there were many documents with distribution detail... `Cancel Cancel Cancel...` But it's always been this way and not the time to change it.

guiclient/printMulticopyDocument.cpp
jcarlin: After reviewing, I don't see any missing error handling or cleanup execution.

guiclient/printMulticopyDocument.cpp
ChristopherCotnoir: The posting option is kind of separate from the printing, isn't it? It seems to me even if the posting fails or is canceled, it should still print.

guiclient/printMulticopyDocument.cpp
ChristopherCotnoir: Ah, I missed the messages in SeriesCreate. OK, this is fine.

PR #1648, opened 2017-07-13
issue #30523 Rollback PR 1642 for now

ChristopherCotnoir:


ChristopherCotnoir:
https://github.com/xtuple/qt-client/pull/1736 is used to resolve issue #30469 without causing issue #30523 instead. This PR is no longer necessary.

PR #1659, opened 2017-07-17
issue #30529 Replaced 'Routings' metric on Work Order screen with a check for the …

ChristopherCotnoir:
…boohead table

### DIFF
--- a/guiclient/workOrder.cpp
+++ b/guiclient/workOrder.cpp
@@ -116,7 +116,27 @@ workOrder::workOrder(QWidget* parent, const char* name, Qt::WindowFlags fl)
     _warehouse->hide();
   }
 
-  if (!_metrics->boolean("Routings"))
+  static QVariant _booEnabled = QVariant();
+
+  if (_booEnabled.isNull())
+  {
+    XSqlQuery boohead;
+    boohead.prepare("SELECT 1 "
+                    "  FROM pg_class "
+                    " WHERE relname='boohead';");
+    boohead.exec();
+
+    if (boohead.first())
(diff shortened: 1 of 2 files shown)


guiclient/workOrder.cpp
```diff
@@ -116,7 +116,27 @@ workOrder::workOrder(QWidget* parent, const char* name, Qt::WindowFlags fl)
...
-  if (!_metrics->boolean("Routings"))
+  static QVariant _booEnabled = QVariant();
+
+  if (_booEnabled.isNull())
+  {
+    XSqlQuery boohead;
+    boohead.prepare("SELECT 1 "
+                    "  FROM pg_class "
+                    " WHERE relname='boohead';");
+    boohead.exec();
```
gilmoskowitz: The declaration, `prepare()`, and `exec()` can all [be combined](https://github.com/xtuple/openrpt/blob/f301a85b07b1b3ddc2c95d6866580af86b98c56e/common/xsqlquery.cpp#L135-L142).

We also have to beware of multiple objects named `boohead`, so this should be limited to just tables (see [PG docs](https://www.postgresql.org/docs/9.3/static/catalog-pg-class.html) and search for `relkind`).

guiclient/workOrder.h
```diff
@@ -84,6 +84,7 @@ protected slots:
   
 
 private:
+    QVariant _booEnabled;
```
gilmoskowitz: The change to the header is unnecessary because the `_booEnabled` in the .cpp is declared as a `static` local variable and hides the class member.

PR #1668, opened 2017-07-20
resolves issue #19021 - inconsistency in RAs (receiving site)

jcarlin:
1. Instead of disabling Recv/Ship Site comboboxes on edit, disable them in populate _IF_ qty credited, qty to receive, qty shipped > 0 https://github.com/xtuple/qt-client/pull/1668/files#diff-3ac3613b6d2a8f0f75c982a136e5a64cR954.
2. If Sales Order warehouse does not match the Recv/Ship on RA header, ask user if they'd like to update them to match the SO before proceeding. The default is Yes.

For 2 above, more details about this decision:
- I didn't want to make the combo boxes `allowNull` because Original Sales Order is optional and user should not always be expected to set the WH.
- Without changing to allowNull, there's no way to know if the WH(s) set previous to original SO should take precedence over the SO warehouse. I like asking the User with the default being Yes (update to match).

### DIFF
--- a/guiclient/returnAuthorization.cpp
+++ b/guiclient/returnAuthorization.cpp
@@ -625,24 +625,6 @@ void returnAuthorization::sOrigSoChanged()
   {
     if (_origso->isValid())
     {
-      if ( !_warehouse->isValid() )
-      {
-        QMessageBox::warning( this, tr("Invalid Receiving Site"),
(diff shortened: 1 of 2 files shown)


jcarlin:
It would be very helpful if the `Original Order # | New Order # | Project #` fields were at the top of the screen.

PR #1682, opened 2017-07-25
issue #27513 - possible to make changes to order while being edited by another user

jcarlin:
Order locking in our application has many feature gaps. This pr adds locking to all Purchase Order list actions and utilities so that if a user is editing the PO it can not be modified elsewhere. If this approach is OK it can be extended to other order types.

### DIFF
--- a/guiclient/unpostedPurchaseOrders.cpp
+++ b/guiclient/unpostedPurchaseOrders.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
  * version 1.0, the full text of which (including xTuple-specific Exhibits)
  * is available at www.xtuple.com/CPAL.  By using this software, you agree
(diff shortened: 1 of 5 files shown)


guiclient/unpostedPurchaseOrders.cpp
```diff
@@ -168,8 +168,13 @@ void unpostedPurchaseOrders::sDelete()
                                       question,
                                       QMessageBox::Yes,
                                       QMessageBox::No | QMessageBox::Default) == QMessageBox::No)
-			  continue;
-		  }
+			        continue;
+		      }
+
+          // If locked, don't proceed
+          if (!_lock.acquire("pohead", ((XTreeWidgetItem*)(selected[i]))->id(), AppLock::Interactive))
```
gilmoskowitz: There should be a corresponding `unlock` after the order has been deleted.

guiclient/unpostedPurchaseOrders.cpp
```diff
@@ -247,7 +252,11 @@ void unpostedPurchaseOrders::sRelease()
     if ((selected[i]->rawValue("pohead_status").toString() == "U")
       && (_privileges->check("ReleasePurchaseOrders"))
       && (checkSitePrivs(((XTreeWidgetItem*)(selected[i]))->id())))
-      {
+    {
+      // If locked, don't proceed
+      if (!_lock.acquire("pohead", ((XTreeWidgetItem*)(selected[i]))->id(), AppLock::Interactive))
```
gilmoskowitz: There should be a corresponding `unlock` after the order has been released.

guiclient/unpostedPurchaseOrders.cpp
```diff
@@ -287,6 +297,10 @@ void unpostedPurchaseOrders::sUnrelease()
         && (_privileges->check("UnreleasePurchaseOrders"))
         && (checkSitePrivs(((XTreeWidgetItem*)(selected[i]))->id())))
     {
+      // If locked, don't proceed
+      if (!_lock.acquire("pohead", ((XTreeWidgetItem*)(selected[i]))->id(), AppLock::Interactive))
```
gilmoskowitz: There should be a corresponding `unlock` after the order has been unreleased.

PR #1701, opened 2017-07-31
guierrorcheck refactor - round 1

NicholasBellucci:
Refactor to support Gui Error Check.

### DIFF
--- a/guiclient/glSeries.cpp
+++ b/guiclient/glSeries.cpp
@@ -201,6 +201,13 @@ enum SetResponse glSeries::set(const ParameterList &pParams)
 void glSeries::sNew()
 {
 
+  QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(!_date->isValid(), _date,
+                           tr("You must enter a Distribution Date for this Series."))
+    ;
+    if (GuiErrorCheck::reportErrors(this, tr("Cannot Maintain G/L Series"), errors))
+      return;
+
   ParameterList params;
   params.append("mode", "new");
   params.append("doctype", _doctype->currentText());
@@ -223,6 +230,13 @@ void glSeries::sNew()
 
 void glSeries::sEdit()
 {
+
+  QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(!_date->isValid(), _date,
+                           tr("You must enter a Distribution Date for this Series."))
+    ;
+    if (GuiErrorCheck::reportErrors(this, tr("Cannot Maintain G/L Series"), errors))
+      return;
 
   ParameterList params;
   params.append("mode", "edit");
--- a/guiclient/glSeries.cpp
+++ b/guiclient/glSeries.cpp
@@ -18,6 +18,7 @@
 #include "storedProcErrorLookup.h"
 #include "submitAction.h"
 #include "errorReporter.h"
+#include "guiErrorCheck.h"
 
 #define cPostStandardJournal 0x10
 
@@ -199,13 +200,6 @@ enum SetResponse glSeries::set(const ParameterList &pParams)
 
 void glSeries::sNew()
 {
-  if(!_date->isValid())
-  {
-    QMessageBox::information( this, tr("Cannot Maintain G/L Series"),
-                              tr("<p>You must enter a Distribution Date for this Series.") );
-    _date->setFocus();
-    return;
-  }
 
   ParameterList params;
   params.append("mode", "new");
@@ -229,13 +223,6 @@ void glSeries::sNew()
 
 void glSeries::sEdit()
 {
-  if(!_date->isValid())
-  {
-    QMessageBox::information( this, tr("Cannot Maintain G/L Series"),
-                              tr("<p>You must enter a Distribution Date for this Series.") );
-    _date->setFocus();
-    return;
-  }
 
   ParameterList params;
   params.append("mode", "edit");
@@ -272,21 +259,15 @@ void glSeries::sDelete()
 bool glSeries::update()
 {
   XSqlQuery glupdate;
-  if(!_date->isValid())
-  {
-    QMessageBox::information( this, tr("Cannot Post G/L Series"),
-                              tr("<p>You must enter a Distribution Date for this Series.") );
-    _date->setFocus();
-    return false;
-  }
 
-  if(_metrics->boolean("MandatoryGLEntryNotes") && _notes->toPlainText().trimmed().isEmpty())
-  {
-    QMessageBox::information( this, tr("Cannot Post G/L Series"),
-                              tr("<p>You must enter some Notes to describe this transaction.") );
-    _notes->setFocus();
-    return false;
-  }
+  QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(!_date->isValid(), _date,
+                           tr("You must enter a Distribution Date for this Series."))
+          << GuiErrorCheck(_metrics->boolean("MandatoryGLEntryNotes") && _notes->toPlainText().trimmed().isEmpty(), _notes,
+                           tr("You must enter some Notes to describe this transaction."))
+    ;
+    if (GuiErrorCheck::reportErrors(this, tr("Cannot Post G/L Series"), errors))
+      return false;
 
 // Do not save notes when posting std journal
   if (_mode == cPostStandardJournal)
--- a/guiclient/glTransaction.cpp
+++ b/guiclient/glTransaction.cpp
@@ -18,6 +18,7 @@
 #include "glcluster.h"
 #include <openreports.h>
 #include "errorReporter.h"
+#include "guiErrorCheck.h"
 
 glTransaction::glTransaction(QWidget* parent, const char* name, bool modal, Qt::WindowFlags fl)
     : XDialog(parent, name, modal, fl)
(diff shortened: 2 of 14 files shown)


guiclient/glSeries.cpp
```diff
@@ -199,13 +200,6 @@ enum SetResponse glSeries::set(const ParameterList &pParams)
...
 {
-  if(!_date->isValid())
-  {
-    QMessageBox::information( this, tr("Cannot Maintain G/L Series"),
-                              tr("<p>You must enter a Distribution Date for this Series.") );
-    _date->setFocus();
-    return;
-  }
 
   ParameterList params;
```
gilmoskowitz: This is a fine place to use GuiErrorCheck. _Removing_ the check causes a problem - it's here to ensure that `newdlg.set(params);` a few lines later gets its required input parameter.

guiclient/glSeries.cpp
```diff
@@ -229,13 +223,6 @@ void glSeries::sNew()
 
 void glSeries::sEdit()
 {
-  if(!_date->isValid())
-  {
-    QMessageBox::information( this, tr("Cannot Maintain G/L Series"),
-                              tr("<p>You must enter a Distribution Date for this Series.") );
-    _date->setFocus();
-    return;
-  }
```
gilmoskowitz: Same here - this is an important check for valid parameter passing to a new dialog.

guiclient/glTransaction.cpp
```diff
@@ -98,37 +99,21 @@ enum SetResponse glTransaction::set(const ParameterList &pParams)
...
+
+  QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(_amount->isZero(), _amount,
+                           tr("You must enter an amount for this G/L Transaction before you may Post it."))
+          << GuiErrorCheck(! _debit->isValid(), _debit,
+                           tr("You must select a Debit Account for this G/L Transaction before you may Post it."))
+          << GuiErrorCheck(! _credit->isValid(), _credit,
+                           tr("You must select a Credit Account for this G/L Transaction before you may Post it."))
+          << GuiErrorCheck(!_metrics->boolean("IgnoreCompany") && _credit->companyId() != _debit->companyId(), _credit,
+                           tr("The Accounts must belong to the same Company to Post this transaciton."))
```
gilmoskowitz: There's a typo in the original - `transaciton` - that is being propagated here.

guiclient/glTransaction.cpp
```diff
@@ -98,37 +99,21 @@ enum SetResponse glTransaction::set(const ParameterList &pParams)
...
+          << GuiErrorCheck(_metrics->boolean("MandatoryGLEntryNotes") && _notes->toPlainText().trimmed().isEmpty(), _notes,
+                           tr("You must enter some Notes to describe this transaction."))
```
gilmoskowitz: 👍 You found the intermediate style!

guiclient/incident.cpp
```diff
@@ -353,35 +354,18 @@ bool incident::save(bool partial)
...
-      _description->setFocus();
+    QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(_crmacct->id() == -1, _number,
+                           tr("You must specify the Account that this incident is for."))
+          << GuiErrorCheck(_cntct->id() <= 0 && _cntct->name().simplified().isEmpty(), _cntct,
+                           tr("You must specify a Contact for this Incident."))
+          << GuiErrorCheck(_description->text().trimmed().isEmpty(), _description,
+                           tr("You must specify a description for this incident report."))
+          << GuiErrorCheck(_status->currentIndex() == 3 && _assignedTo->username().isEmpty(), _assignedTo,
+                           tr("You must specify an assignee when the status is assigned."))
```
gilmoskowitz: Nice catch on the wrong focus widget.

guiclient/invoice.cpp
```diff
@@ -540,30 +541,20 @@ void invoice::sSave()
     QString	msg;
     QWidget*	widget;
   } error[] = {
-    { _cust->id() <= 0,
-      tr("<p>You must enter a Customer for this Invoice before saving it."),
-      _cust
-    },
     // TODO: add more error checks here?
     { true, "", NULL }
   };
```
gilmoskowitz: The whole `struct` definition can be removed.

guiclient/miscCheck.cpp
```diff
@@ -103,49 +104,21 @@ void miscCheck::sSave()
...
-			     "expensed payment.") );
-    _expcat->setFocus();
-    return;
-  }
+  QList<GuiErrorCheck> errors;
+    errors<< GuiErrorCheck(!_date->isValid(), _date,
+                           tr("You must enter a date for this payment."))
+          << GuiErrorCheck(_amount->isZero(), _amount,
+                           tr("You must enter an amount for this payment."))
+          << GuiErrorCheck((_applytocm->isChecked()) && (_cmCluster->id() == -1), _expcat,
```
gilmoskowitz: Hmmm... Try using `_cmCluster` as the focus widget. It would be a better choice here; I don't know why the old code had `_expcat`.

guiclient/miscCheck.cpp
```diff
@@ -103,49 +104,21 @@ void miscCheck::sSave()
...
-    _expcat->setFocus();
-    return;
-  }
- 
-  if (_applytocm->isChecked() && _cmCluster->isValid() && (_amt > _aropenamt))
-  {
-    QMessageBox::warning( this, tr("Invalid Amount"),
-                            tr("<p>You must enter an amount less than or equal to the  "
-	   		                   "credit memo selected.") );
-    _amount->setLocalValue(_aropenamt);
```
gilmoskowitz: There's no good way to carry through this functionality to the new style. Thoughts?

guiclient/reconcileBankaccount.cpp
```diff
@@ -185,49 +186,26 @@ bool reconcileBankaccount::sSave(bool closeWhenDone)
...
-  }
 
-  if(!_datesAreOK)
-  {
-    QMessageBox::critical( this, tr("Dates already reconciled"),
-                tr("The date range you have entered already has "
-                   "reconciled dates in it. Please choose a different "
-                   "date range.") );
-    _startDate->setFocus();
+  if(!_datesAreOK){
```
gilmoskowitz: Is there a reason to keep the `if`?

guiclient/standardJournal.cpp
```diff
@@ -137,14 +132,16 @@ void standardJournal::sSave()
...
-                             "Please enter in a different Name for this Standard Journal."));
-    _name->setFocus();
+
+  QList<GuiErrorCheck> errors;
+  errors<< GuiErrorCheck(_name->text().length() == 0, _name,
+                         tr("You must enter a valid Name.")),
+  errors<< GuiErrorCheck(standardSave.first(), _name,
+                         tr("The Name you have entered for this Standard Journal already exists. "
+                           "Please enter in a different Name for this Standard Journal."))
+  ;
```
gilmoskowitz: Nice catch with `standardSave.first()`.

NicholasBellucci:
Gil, 

I have added thoughts and comments in line with the previous email. These comments are in green. Let me know if you can’t see them.

-Nicholas Bellucci

gilmoskowitz:
@NicholasBellucci - your comments are hard to find because the formatting from your email client got lost. The best way to respond to a code review is through the github browser interface; reply to each comment independently. It's also not necessary to add `Fixed this` comments - the code and github UI will show that after the next push :-)

guiclient/miscCheck.cpp
NicholasBellucci: I'm confused with this statement. I currently have it using the new functionality. Do we not want this?

guiclient/reconcileBankaccount.cpp
NicholasBellucci: I kept the if statement because in the previous setup if !_datesAreOK was a much longer function and at the end set datesAreOK to false so in order to keep that consistent I kept the if statement.

guiclient/miscCheck.cpp
gilmoskowitz: The code transformation was done correctly but with a loss of functionality - the new version does not change the value of the `_amount` widget to `_aropenamt`.

That raises two questions: 
1. Is changing the value of `_amount` a feature worth preserving?
2. Was removing it a conscious decision?

To answer the first question, "No, this is not worth preserving." Rationale: the client as a rule does not change erroneous values to other potentially erroneous values during validation. Doing so loses information that might be important to the user; in this case it might be a simple typo like an extra trailing 0.

It is, however, a change from previous behavior that `git blame` says goes back to Nov 2007. A behavior change like this should be thought through carefully.

NicholasBellucci: I did not notice the setLocalValue. I do believe I could get this functionality to work with the new setup if we want to preserve the feature.

guiclient/reconcileBankaccount.cpp
gilmoskowitz: The resulting code is this:
```
  if(!_datesAreOK){
    _datesAreOK = false;
  }
```
which does nothing useful. The old code wasn't any better but the superfluous line was hidden by the bulk of the old code (going back to 2008).

guiclient/miscCheck.cpp
gilmoskowitz: Let's keep the new code here as is. I think the "feature" does more harm than good.

PR #1708, opened 2017-08-02
guierrorcheck refactor - round 2

NicholasBellucci:


guiclient/setup.cpp
```diff
@@ -506,10 +507,14 @@ void setup::setCurrentIndex(XTreeWidgetItem* item)
...
-                                tr("<p>There was an error loading the UI Form "
-                                   "from the database."));
+
+        QList<GuiErrorCheck> errors;
+        errors<< GuiErrorCheck(!uiFile.open(QIODevice::ReadOnly), _tree,
+                           tr("There was an error loading the UI Form from the database."))
+        ;
+        if (GuiErrorCheck::reportErrors(this, tr("Could not load UI"), errors))
+          return;
+
```
gilmoskowitz: Please revert changes to `setup.cpp`. This is a case where the same kind of code serves a different purpose that's not well suited to `GuiErrorCheck`.

PR #1711, opened 2017-08-03
issue #24452 Prevent SQL Injection using bindValue

ChristopherCotnoir:


guiclient/display.cpp
```diff
@@ -305,8 +305,11 @@ bool displayPrivate::setParams(ParameterList &params)
     param = params.value(column, &valid);
     if (valid)
     {
-      QStringList list = param.toStringList();
-      clauses.append(QString("charass_alias%1.charass_value IN ('%2') ").arg(_charidslist.at(i).toString()).arg(list.join("','")));
+      QStringList list;
+      for (int j = 0; j < param.toStringList().count(); j++)
```
gilmoskowitz: Most of the loops here can be made easier to read using `foreach`.

PR #1712, opened 2017-08-03
Issue #29234: Special character handling when saving lot characteristics

andersdd:


guiclient/lotSerialUtils.cpp
```diff
@@ -89,7 +89,7 @@ void LotSerialUtils::updateLotCharacteristics(int ls_id, const QList<QWidget *>
         XSqlQuery q;
         bool success = q.exec(QString("INSERT INTO charass "
                               "(charass_value, charass_target_type, charass_target_id, charass_char_id) "
-                              "VALUES ('%1', 'LS', %2, %3)").arg(char_text).arg(ls_id).arg(_charIds.at(i)));
+                              "VALUES ($$%1$$, 'LS', %2, %3)").arg(char_text).arg(ls_id).arg(_charIds.at(i)));
```
gilmoskowitz: This should use bound variables, not string interpolation, for safety.

PR #1715, opened 2017-08-04
Issue #14206: Credit Card Transaction screen

andersdd:
Added label but could not see how this date field was ever being populated.  Should the widget be removed from the screen?


gilmoskowitz:
The date field was apparently never populated but the value [is passed in](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/externalccprocessor.cpp#L186-L187) by the caller.

gilmoskowitz:
Please set the date field using the value passed in from the caller.

PR #1721, opened 2017-08-07
Issue #29291: Additional checks and notifications when deleting terms

andersdd:
Provide user message instead of foreign key SQL error

### DIFF
--- a/guiclient/termses.cpp
+++ b/guiclient/termses.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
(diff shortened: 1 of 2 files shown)


guiclient/termses.cpp
```diff
@@ -123,6 +123,48 @@ void termses::sDelete()
...
+                              "as there are one or more Sales Orders or A/R documents "
+                              "assigned to it." ) );
+    return;
+  }
+  else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving A/P Information"),
+                                termsesDelete, __FILE__, __LINE__))
+  {
+    return;
+  }
+
```
gilmoskowitz: Instead of adding these 2 queries and their associated messages and error checks (4 if you want to be more precise in error reporting), let ErrorReporter do the work. Uncomment the [reference to the `cohead_cohead_terms_id_fkey`](https://github.com/xtuple/qt-client/blob/4_11_x/common/errorReporter.cpp#L288) and fill in the error message text. Same with the purchase order f-key. You need to add lines to the structure for `aropen` and `apopen`.

andersdd: pohead and cohead - yes.  There's no foreign key on apopen and aropen and we cannot add them as the system is using -1 throughout so still need these separate checks.

PR #1726, opened 2017-08-07
issue #30679 Don't make cntctsel persistent

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3121.

ChristopherCotnoir:
There shouldn't be much of a real difference that I can see. If it were dropped it would just be recreated if the user reopened the window in the same session, which seems like extra work.

PR #1731, opened 2017-08-08
issue #30703 Let function handle check

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3126.

PR #1733, opened 2017-08-08
issue #30685 Workaround apparent Qt bug

ChristopherCotnoir:


andersdd:
Good spotting to fix this issue.  Banged my head for some time wondering why it was not working.

PR #1735, opened 2017-08-08
issue #30008 - backport to 4.10.2

gilmoskowitz:
See https://github.com/xtuple/qt-client/pull/1737 for an alternative fix - bigger and more disruptive but perhaps better long-term.


widgets/usernameCluster.cpp
```diff
@@ -100,20 +101,20 @@ void UsernameLineEdit::setUsername(const QString & pUsername)
...
-  query.bindValue(":username", QString(pUsername).prepend("^"));
+  MetaSQLQuery usrm("SELECT usr_id"
+                    "  FROM usr"
+                    " WHERE UPPER(usr_username) = UPPER(<? value('username') ?>)"
+                    "  <? if exists('isActive') ?>  AND usr_active    <? endif ?>"
+                    "  <? if exists('isInactive') ?>AND NOT usr_active<? endif ?>"
+                    ";");
+  ParameterList params;
+  params.append("username", pUsername);
+  if (UsersActive == _type)        params.append("isActive",   true);
```
jcarlin: New line below `if` for `params.append("isActive",   true);`. Same below.

gilmoskowitz:
Replaced by #1737

PR #1736, opened 2017-08-09
issue #30469 Make sure JS DateTime objects coming from QDate have the timezone of …

ChristopherCotnoir:
…the postgres server

Based off of https://github.com/xtuple/qt-client/pull/1642, which was rolled back due to breaking scripts that used JS DateTime functions on incoming QDate objects. This alternative solution keeps the script object coming from QDate objects as a JS DateTime, but adjusts the timezone to match the postgres server so that the date will not change when used in a postgres query.

### DIFF
--- a/scriptapi/qdateproto.cpp
+++ b/scriptapi/qdateproto.cpp
@@ -17,9 +17,12 @@ QScriptValue QDateToScriptValue(QScriptEngine *engine, const QDate &in)
   XSqlQuery getTZ("SELECT EXTRACT(TIMEZONE FROM now()) AS offset;");
   if (getTZ.first())
     return engine->newDate(QDateTime(in, QTime(), QTimeZone(getTZ.value("offset").toInt())));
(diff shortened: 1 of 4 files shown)


scriptapi/qdateproto.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+QScriptValue QDateToScriptValue(QScriptEngine *engine, const QDate &in)
+{
+  XSqlQuery getTZ("SELECT EXTRACT(TIMEZONE FROM now()) AS offset;");
+  if (getTZ.first())
+    return engine->newDate(QDateTime(in, QTime(), QTimeZone(getTZ.value("offset").toInt())));
+  else if (ErrorReporter::error(QtCriticalMsg, 0, "Failed to fetch server time zone",
+                                getTZ, __FILE__, __LINE__))
+    return engine->newDate(QDateTime(in));
+}
```
gilmoskowitz: This function has no `return` if `getTZ` fails to return a row but there's no database error. It's hard to see how that would happen but the compiler should have raised a warning because it doesn't know the semantics.

PR #1737, opened 2017-08-09
issue #30008 - revamp usernameCluster and virtualCluster

gilmoskowitz:
This is an alternative to xtuple/qt-client#1735, which seems to ignore partial matches in the completer. This is different than other VirtualCluster subclasses.

The change here pushes most of the responsibility for the username cluster back up to the parent virtual cluster class. The virtual cluster handling of active vs. inactive records has been enhanced by making inactive records a bit more obvious and putting them at the end of the sort order.


widgets/virtualCluster.cpp
```diff
@@ -864,7 +867,8 @@ void VirtualClusterLineEdit::sSearch()
       numQ.prepare(_query + _numClause +
                    (_extraClause.isEmpty() || !_strict ? "" : " AND " + _extraClause) +
                    ((_hasActive && ! _showInactive) ? _activeClause : "" ) +
-                   QString("ORDER BY %1;").arg(_numColName));
+                   QString(" ORDER BY %1 %2 LIMIT 1;")
```
jcarlin: Why `LIMIT 1`?

widgets/virtualCluster.cpp
```diff
@@ -316,8 +316,7 @@ VirtualClusterLineEdit::VirtualClusterLineEdit(QWidget* pParent,
         _completer->setCompletionColumn(1);
         _completer->setModelSorting(QCompleter::CaseInsensitivelySortedModel);
         connect(this, SIGNAL(textEdited(QString)), this, SLOT(sHandleCompleter()));
-        connect(_completer, SIGNAL(highlighted(QString)), this, SLOT(setText(QString)));
```
jcarlin: This is handled automatically and is no longer needed I assume.

gilmoskowitz: This isn't handled "automatically". The `activated` signal below should cover more cases.

PR #1756, opened 2017-09-07
issue #30813 Still call applyDefaultFilterSet if sFillList is called before the wi…

ChristopherCotnoir:
…ndow is shown after modifying the default filter for the sFillList call


ChristopherCotnoir:
It's actually the opposite. applyDefaultFilterSet must be called after the setDefault calls. The setDefault calls do not actually modify the current filter set, they modify the default values for all filter sets for this instance of the window, however the new default values will not take effect until a filter set is applied. applyDefaultFilterSet is what causes the parameters passed into set to function at all, which is why removing it caused problems.

PR #1757, opened 2017-09-11
issue #30496 Make xtsettings available to scripting and properly call the applock …

ChristopherCotnoir:
…setup which was previously orphaned


ChristopherCotnoir:
Associated with xtuple/connect#99.

PR #1759, opened 2017-09-12
issue #29523 Remove transactions from screen

ChristopherCotnoir:


ChristopherCotnoir:
It wasn't intended to. I don't think there is any reasonable way to implement that, and it would likely be better if that portion of the window was treated separately as a permanent change. There are other places in the application that do this (usually XTreeWidgets).

gilmoskowitz:
Bug 30974 captures the discussion above.

PR #1760, opened 2017-09-12
issue #30781 - fix completer bugs

gilmoskowitz:
also bring code closer to conforming with coding standards

### DIFF
--- a/widgets/ordercluster.cpp
+++ b/widgets/ordercluster.cpp
@@ -358,7 +358,7 @@ void OrderLineEdit::sParse()
         XSqlQuery countQ;
         countQ.prepare("SELECT COUNT(*) AS count FROM orderhead WHERE true " + _numClause +
                         (_extraClause.isEmpty() || !_strict ? "" : " AND " + _extraClause) +
-                        (_hasActive ? _activeClause : "" ) + QString(";"));
+                        ((_hasActive && ! _showInactive) ? _activeClause : "" ) + QString(";"));
         countQ.bindValue(":number", numQ.value("number").toString());
         countQ.exec();
         countQ.first();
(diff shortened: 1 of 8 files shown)


widgets/ordercluster.cpp
```diff
@@ -326,68 +328,71 @@ void OrderLineEdit::sNewId(const int p)
...
+  {
+    setId(_completerId);
+    _completerId = 0;
+  }
+  else if (! _parsed)
+  {
+    QString stripped = text().trimmed().toUpper();
+    if (stripped.isEmpty())
     {
       _parsed = true;
```
ChristopherCotnoir: This will only set _parsed if stripped.isEmpty(), whereas before _parsed was always set.

widgets/ordercluster.cpp
```diff
@@ -326,68 +328,71 @@ void OrderLineEdit::sNewId(const int p)
...
-                    (_hasActive ? _activeClause : "" ) +
-                    QString("ORDER BY %1 LIMIT 1;").arg(_numColName));
-        numQ.bindValue(":number", "^" + stripped);
-        numQ.exec();
-        if(numQ.first())
+        // Check for other active orders with the same number {
+        XSqlQuery countQ;
+        countQ.prepare("SELECT COUNT(*) AS count FROM orderhead WHERE true " + _numClause +
+                        (_extraClause.isEmpty() || !_strict ? "" : " AND " + _extraClause) +
+                        (_hasActive ? _activeClause : "" ) + QString(";"));
```
ChristopherCotnoir: if _showInactive was added above, shouldn't it be added here too for consistency?

widgets/ordercluster.cpp
gilmoskowitz: This should be addressed by https://github.com/xtuple/qt-client/pull/1760/files#diff-a9231029cbc9e49a67b6123039504711R397

PR #1787, opened 2017-10-05
Issue #18973: Handle partially entered months for some locale formats

andersdd:
Resolves entering single month digit (<10) when locale expects MM

### DIFF
--- a/widgets/datecluster.cpp
+++ b/widgets/datecluster.cpp
@@ -206,6 +206,17 @@ void XDateEdit::parseDate()
              qPrintable(dateFormatStr));
 
     QDate tmp = QDate::fromString(dateString, dateFormatStr);
+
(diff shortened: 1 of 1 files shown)


widgets/datecluster.cpp
```diff
@@ -206,6 +206,17 @@ void XDateEdit::parseDate()
              qPrintable(dateFormatStr));
 
     QDate tmp = QDate::fromString(dateString, dateFormatStr);
+
+    if (tmp.month() < 10 && dateFormatStr.indexOf(QRegExp("M{2}")) >= 0) // handle partial entered months
```
gilmoskowitz: The regular expression isn't right. It will match 2 `M`s in a row but will also match 3 or more `M`s. The regular expression has to exclude non-`M`s, taking into account that the month may come at the beginning or end of the string.

gilmoskowitz: @andersdd Do you have a few minutes to look into the `M{2}` vs. strict number of `M`s concern?

PR #1795, opened 2017-10-12
issue #26635 Add XDocumentWindow class that includes all subclasses in a QScrollVi…

ChristopherCotnoir:
…ew, make SalesOrder a subclass

### DIFF
--- a/guiclient/salesOrder.cpp
+++ b/guiclient/salesOrder.cpp
@@ -116,7 +116,7 @@ salesOrder::salesOrder(QWidget *parent, const char *name, Qt::WindowFlags fl)
   connect(_orderNumber,         SIGNAL(textChanged(const QString &)),           this,         SLOT(sSetUserEnteredOrderNumber()));
   connect(_save,                SIGNAL(clicked()),                              this,         SLOT(sSave()));
   connect(_saveAndAdd,          SIGNAL(clicked()),                              this,         SLOT(sSaveAndAdd()));
-  connect(_close,               SIGNAL(clicked()),                              this,         SLOT(closeDocument()));
+  connect(_close,               SIGNAL(clicked()),                              this,         SLOT(close()));
   connect(_shippingCharges,     SIGNAL(newID(int)),                             this,         SLOT(sHandleShipchrg(int)));
   connect(_shipToAddr,          SIGNAL(changed()),                              this,         SLOT(sConvertShipTo()));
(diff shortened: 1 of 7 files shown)


guiclient/xdocumentwindow.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+class XDocumentWindowPrivate : public Ui::XDocumentWindow
+{
+public:
+  XDocumentWindowPrivate(::XDocumentWindow * parent) : _parent(parent)
+  {
+    setupUi(_parent);
+  }
+
+private:
```
gilmoskowitz: Since this class is intended as the keeper of `XDocumentWindow`'s private data, its members should all be public, `XDocumentWindow` and `XDocumentWindowPrivate` should be `friend`s, or both.

guiclient/salesOrder.cpp
```diff
@@ -87,7 +87,8 @@ salesOrder::salesOrder(QWidget *parent, const char *name, Qt::WindowFlags fl)
     _taxzoneidCache(-1),
     _crmacctid     (-1)
 {
-  setupUi(this);
+  setupUi(widget());
```
gilmoskowitz: It's unclear why we `setupUi(_widget)`, the frame inside the scroll area, instead of `setupUi(XDocumentWindow)`.

guiclient/xdocumentwindow.h
```diff
@@ -0,0 +1,35 @@
...
+
+  public:
+    XDocumentWindow(QWidget * parent = 0, Qt::WindowFlags flags = 0);
+    XDocumentWindow(QWidget * parent, const char * name, Qt::WindowFlags flags = 0);
+    ~XDocumentWindow();
+
+    Q_INVOKABLE QWidget * widget();
+
+  public slots:
+    void closeDocument();
```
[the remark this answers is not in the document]
ChristopherCotnoir: I found that if I don't do this workaround, close() called on salesOrder will only clear the ui elements inside the QScrollView instead of closing the window itself, because it is only closing salesOrder, not XDocumentWindow, which is now top-level.

guiclient/salesOrder.cpp
ChristopherCotnoir: This tells setupUi where to put all of salesOrder.ui's elements: inside the QScrollView. See display subclasses for a comparison.

PR #1796, opened 2017-10-12
issue #26730 - Add utility for viewing and killing processes/locks

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3181.

ChristopherCotnoir:
This tool is meant to kill processes that aren't properly terminated. It's a last-resort tool, like all Task Killers. We should allow the Admin to kill whatever they need to and assume they know what they are doing.

Don't merge yet, though. Awaiting review from Gil and Perry.

xtpclark:
👍 Looks good to me.

PR #1801, opened 2017-10-17
issue #31191 Prevent adding an item as a component to itself on a work order

ChristopherCotnoir:


guiclient/woMaterialItem.cpp
```diff
@@ -255,6 +255,32 @@ void woMaterialItem::sSave()
...
+        QMessageBox::warning(this, tr("Cannot Create W/O Material Requirement"),
+                             tr("Cannot add an Item as a Material to itself."));
+        _item->setId(-1);
+        _item->setFocus();
+        return;
+      }
+
+      woSave.bindValue(":wo_id", woSave.value("id").toInt());
+      woSave.exec();
+    }
```
gilmoskowitz: Use a recursive query instead of repeatedly `exec()`ing.

xtuple-ci-user:
Build finished.

PR #1816, opened 2017-11-02
issue #31315 Use full widget path for preferences

ChristopherCotnoir:


widgets/parameterwidget.cpp
```diff
@@ -133,14 +133,15 @@ ParameterWidget::~ParameterWidget()
...
-    result = parent()->objectName();
-    if (result.isEmpty())
-      result = parent()->metaObject()->className();
+    if (!child->objectName().isEmpty())
+      result.prepend(child->objectName() + "/");
+    child = child->parentWidget();
   }
 
-  result += "/" + objectName();
+  result += objectName();
```
gilmoskowitz: What about this?
```C++
QString result;
for (QWidget *w = this; w; w = w->parentWidget())
  result.prepend((w->objectName().isEmpty() ? w->metaObject()->className()
                                            : w->objectName())             + "/");
return result;
// or
QStringList result;
for (QWidget *w = this; w; w = w->parentWidget())
  result << w->objectName().isEmpty() ? w->metaObject()->className() : w->objectName();
return result.join("/");
```
The handling of all objects in the chain would be uniform and none get skipped.

PR #1823, opened 2017-11-09
issue #31446 Expose QMimeType and QMimeDatabase to scripting

ChristopherCotnoir:


gilmoskowitz:
We need to test this to make sure the script engine properly distinguishes between the various overloads.

PR #1825, opened 2017-11-10
issue #31445 Add signals/slots for saving interaction with scripts

ChristopherCotnoir:


gilmoskowitz:
Eek. We'll have to revisit this `sSave()` in the future and reorganize it.

PR #1827, opened 2017-11-14
issue #19108 Load scripts in search path order everywhere

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3197.

PR #1831, opened 2017-11-27
issue #31217 - add create, edit, view prospect actions to addresses and contacts lists

jcarlin:
https://github.com/xtuple/xtuple/pull/3198

### DIFF
--- a/guiclient/addresses.cpp
+++ b/guiclient/addresses.cpp
@@ -24,6 +24,7 @@
 #include "storedProcErrorLookup.h"
 #include "parameterwidget.h"
 #include "errorReporter.h"
+#include "prospect.h"
 
 addresses::addresses(QWidget* parent, const char*, Qt::WindowFlags fl)
   : display(parent, "addresses", fl)
@@ -55,6 +56,7 @@ addresses::addresses(QWidget* parent, const char*, Qt::WindowFlags fl)
     connect(list(), SIGNAL(itemSelected(int)), this, SLOT(sView()));
   }
 
+  _crmacctid = -1;
 }
 
 void addresses::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem*, int)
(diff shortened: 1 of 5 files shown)


guiclient/addresses.cpp
```diff
@@ -69,6 +71,39 @@ void addresses::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem*, int)
...
   menuItem->setEnabled(_privileges->check("MaintainAddresses"));
+
+  // Create, Edit, View Prospect:
+  // Temporary sql hack to determine crmacct. Replace this post crm enhancements project.
+  XSqlQuery sql;
+  sql.prepare("SELECT crmacct_id, crmacct_prospect_id"
+              "  FROM cntct"
+              "  JOIN crmacct ON cntct_crmacct_id=crmacct_id"
+              " WHERE cntct_addr_id=:addr_id"
+              "   AND crmacct_cust_id IS NULL;");
```
gilmoskowitz: This query may return multiple rows because an address may be shared among several contacts. Suggestion: sort descending by `crmacct_prospect_id IS NULL` then `LIMIT 1` to reduce ambiguity or sort descending by prospect id with NULLs last to eliminate it entirely.

If multiple addresses are selected then there is no single CRM account, so this won't return the right results. Consider moving this query into the sNew/sEdit/sView methods and using a slightly different one here to decide which menu items to use.

guiclient/addresses.cpp
```diff
@@ -131,3 +166,49 @@ void addresses::sDelete()
     return;
   }
 }
+
+void addresses::sNewProspect()
+{
+  QList<XTreeWidgetItem*> selected = list()->selectedItems();
+  for (int i = 0; i < selected.size(); i++)
```
gilmoskowitz: Consider using `foreach`:
```
foreach (XTreeWidgetItem *item, list()->selectedItems())
{
  ...
}
```
The desirability of this will become apparent if `item` gets used inside the loop.

guiclient/contacts.cpp
```diff
@@ -356,5 +383,54 @@ void contacts::sOpen()
...
+  {
+    ParameterList params;
+    params.append("mode", "new");
+    params.append("cntct_id", ((XTreeWidgetItem*)(selected[i]))->id());
+    params.append("crmacct_id", ((XTreeWidgetItem*)(selected[i]))->rawValue("crmacct_id"));
+
+    prospect *newdlg = new prospect();
+    newdlg->set(params);
+    omfgThis->handleNewWindow(newdlg);
+  }
```
gilmoskowitz: `foreach` will simplify this loop and the 2 below.

PR #1835, opened 2017-12-01
issue #31600 Rework Item Sources screens

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3200, associated with xtuple/private-extensions#978.

guiclient/displays/dspItemSourcePrices.cpp
```diff
@@ -0,0 +1,157 @@
...
+  setExpandVisible(true);
+  setCollapseVisible(true);
+
+  QString qrySite = QString("SELECT 0, '%1' AS name, '' AS code "
+                    "UNION "
+                    "SELECT warehous_id, warehous_code, warehous_code "
+                    "FROM site() "
+                    "WHERE warehous_active "
+                    "ORDER BY code;")
+                    .arg(tr("[ Not Site Restricted ]"));
```
gilmoskowitz: - Coding standard - SQL formatting.
- It would be nice if we could use a warehouse combobox with allowNull set to true and `tr("[ Not Site Restricted ]")` as the null text. [Site](https://github.com/xtuple/qt-client/blob/4_11_x/widgets/parameterwidget.h#L37) is defined as one of the `ParameterWidgetTypes`, implying it shouldn't be too difficult.

guiclient/displays/dspItemSources.cpp
```diff
@@ -266,24 +260,20 @@ void dspItemSources::sBuyCard()
...
+  item.exec();
+  if (item.first())
+    itemid = item.value("item_id").toInt();
+  else if(ErrorReporter::error(QtCriticalMsg, this, tr("Error fetching item id"),
+                               item, __FILE__, __LINE__))
+    return;
 
   dspPoItemReceivingsByItem *newdlg = new dspPoItemReceivingsByItem();
-  newdlg->set(params);
+  newdlg->findChild<ItemCluster*>("_item")->setId(itemid);
```
gilmoskowitz: `dspPoItemReceivingsByItem ` inherits indirectly from `XWidget` so we could override `set()` there as we do elsewhere in the application.

guiclient/displays/dspItemSources.cpp
```diff
@@ -266,24 +260,20 @@ void dspItemSources::sBuyCard()
 
 void dspItemSources::sReceipts()
 {
-  ParameterList params;
-  params.append("itemsrc_id", list()->id());
+  XSqlQuery item;
+  int itemid = -1;
+  item.prepare("SELECT item_id "
+               "  FROM item "
+               " WHERE item_number=:item_number;");
```
gilmoskowitz: If the MetaSQL query that drives this display retrieved the item_id and stuffed it into `_list` as the item number column's ID role, this query wouldn't be necessary.

PR #1836, opened 2017-12-01
Fixes #31445 The DocAttach dialog will need modifications to enhance its scriptability

bendiy:


widgets/docAttach.cpp
```diff
@@ -547,7 +577,15 @@ void docAttach::sSave()
...
-  emit saveBeforeCommit();
+  if (_urlid > 0)
+  {
+    emit saveBeforeCommit(_urlid);
+  }
+  else if (newDocass.first() && newDocass.value("docass_id").toInt() > 0)
+  {
+    emit saveBeforeCommit(newDocass.value("docass_id").toInt());
+  }
+
```
gilmoskowitz: To create a uniform set of signals, please don't add "the" id to the signal signature. We don't know what information various scripts will need from the various objects that emit the signal. The id and type will be needed in this case so expose both of those to scripting and let the script decide what to do with the available information.

widgets/docAttach.h
```diff
@@ -27,26 +28,35 @@ class docAttach : public QDialog, public Ui::docAttach
     docAttach(QWidget* parent = 0, const char* name = 0, bool modal = false, Qt::WindowFlags fl = 0);
     ~docAttach();
 
+    Q_INVOKABLE virtual QString mode() { return _mode; }
+    Q_INVOKABLE virtual QString purpose() { return _purpose; }
+    Q_INVOKABLE virtual int sourceId() { return _sourceid; }
+    Q_INVOKABLE virtual QString sourceType() { return _sourcetype; }
+    Q_INVOKABLE virtual int targetId() { return _targetid; }
+    Q_INVOKABLE virtual QString targetType() { return _targettype; }
+    Q_INVOKABLE virtual int urlId() { return _urlid; }
```
gilmoskowitz: Please do not put code in header files, even for simple functions like these. If you see instances of it when editing files, move the code to the `.cpp`.

widgets/docAttach.h
```diff
@@ -27,26 +28,35 @@ class docAttach : public QDialog, public Ui::docAttach
...
-    virtual void sHandleButtons();
-    virtual void sSave();
-    virtual void sHandleNewId(int);
-    inline virtual void setSaveStatus(SaveStatus status) { _saveStatus = status; };
+    void set( const ParameterList & pParams );
+    void sFileList();
+    void sHandleButtons();
+    void sSave();
+    void sHandleNewId(int);
+    inline void setSaveStatus(SaveStatus status) { _saveStatus = status; };
```
gilmoskowitz: The `virtual` keyword is important. Please do not remove it.

widgets/scriptablewidget.cpp
```diff
@@ -30,6 +30,37 @@
...
+      QList<QVariant> cpplist;
+      for (int i = 0;  i < it.value().property("length").toInt32(); i++)
+        cpplist.append(it.value().property(i).toVariant());
+      params.append(it.name(), cpplist);
+    }
+    else
+      params.append(it.name(), it.value().toVariant());
+  }
+}
+
```
gilmoskowitz: These two functions already exist in [parameterlistsetup.cpp](https://github.com/xtuple/qt-client/blob/4_11_x/scriptapi/parameterlistsetup.cpp) and don't need to be duplicated here.

widgets/scriptablewidget.cpp
```diff
@@ -161,3 +192,17 @@ void ScriptableWidget::loadScriptEngine()
...
+  if(engine() && engine()->globalObject().property("setParams").isFunction())
+  {
+    QScriptValue paramArg = ParameterListToScriptValue(engine(), params);
+    QScriptValue tmp = engine()->globalObject().property("setParams").call(QScriptValue(), QScriptValueList() << paramArg);
+    ret = ret && tmp.toBool();
+    params.clear();
+    ParameterListFromScriptValue(paramArg, params);
+  }
+  return ret;
+}
```
gilmoskowitz: If this is here, consider removing or replacing the [code it was copied from](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/display.cpp#L513-L525).

widgets/docAttach.h
bendiy: I saw warnings from the compiler about it. That why I removed them. Are those warnings benign?

bendiy: `docAttach.h:42: Warning: Signals cannot be declared virtual`

bendiy:
@gilmoskowitz fixed.

widgets/docAttach.cpp
```diff
@@ -547,7 +613,17 @@ void docAttach::sSave()
...
+  {
+    _targetid = _urlid;
+    emit saveBeforeCommit();
+  }
+  else if (newDocass.first() && newDocass.value("docass_id").toInt() > 0)
+  {
+    _targetid = newDocass.value("docass_id").toInt();
+    emit saveBeforeCommit();
+  }
+
```
gilmoskowitz: If the `UPDATE url` is changed to `RETURNING url_id AS docass_id` then this becomes uniform and 10 lines go to 2:
```C++
_targetid = newDocass.value("docass_id").toInt();
emit saveBeforeCommit();
```

bendiy: I tried that originally, but the `url` view uses Rules and doesn't currently have a `RETURNING` statement. To further complicate things, conditional Rules cannot have RETURNING clauses. I think to do this right, I need to drop the Rules and switch to triggers on the `url` view. Should I make that change in `xtuple/xtuple`?

gilmoskowitz: No, don't bother. Thanks.

PR #1837, opened 2017-12-04
issue #31636 Expand to more windows, fix New Contact issue

ChristopherCotnoir:


guiclient/projects.cpp
```diff
@@ -554,112 +378,207 @@ void projects::sView()
...
+            break;
+          case -5:
+            errmsg = tr("One or more Purchase order Items refer to this project.");
+            break;
+          case -6:
+            errmsg = tr("One or more Invoices refer to this project.");
+            break;
+          default:
+            errmsg = tr("Error #%1 encountered while trying to delete project.").arg(result);
+        }
```
gilmoskowitz: Move to storedProcErrorLookup() so it'll be handled well now and when we rewrite `deleteProject()`.

guiclient/projects.cpp
```diff
@@ -554,112 +378,207 @@ void projects::sView()
...
 
 void projects::sDelete()
 {
-  XSqlQuery projectsDelete;
-  projectsDelete.prepare("SELECT deleteProject(:prj_id) AS result");
-  projectsDelete.bindValue(":prj_id", list()->id());
-  projectsDelete.exec();
-  if(projectsDelete.first())
+  QList<XTreeWidgetItem*> selected = list()->selectedItems();
+  for (int i = 0; i < selected.size(); i++)
```
gilmoskowitz: Here and elsewhere:
```
XSqlQuery qry;
qry.prepare("SELECT deleteWhatever(:id) AS result;");
foreach (XTreeWidgetItem* item, list()->selectedItems())
{
  // check, bind, exec
}
```

guiclient/projects.cpp
```diff
@@ -554,112 +378,207 @@ void projects::sView()
...
-  if (list()->id() == -1)
+  QList<XTreeWidgetItem*> selected = list()->selectedItems();
+  for (int i = 0; i < selected.size(); i++)
   {
-    QMessageBox::information(this, tr("Project Copy"), tr("Please select a project to copy first"));
-    return;
-  }
+    XTreeWidgetItem* item = (XTreeWidgetItem*)(selected[i]);
+
+    if (item->altId() != 1 || !getPriv(cEdit, item))
```
gilmoskowitz: Here and elsewhere, magic numbers for `altId()` aren't good. It would be nicer to define constants at the top of the file and use them, something like this:
```c++
static const int INVOICE = 35;
...
if (item->altId() == INVOICE && getPriv(cEdit, item))
  ...
```

guiclient/projects.cpp
```diff
@@ -554,112 +378,207 @@ void projects::sView()
...
+             (omfgThis->username() == item->rawValue("prj_username") &&
+              _privileges->check("ViewPersonalProjects")) ||
+             _privileges->check("ViewAllProjects");
+
+
+  if(item->altId() == 15 || item->altId() == 17)
+    if (mode==cEdit)
+      return _privileges->check("MaintainQuotes");
+    else
+      return _privileges->check("MaintainQuotes") || _privileges->check("ViewQuotes");
```
gilmoskowitz: Simplify:
```c++
return _privileges->check("MaintainQuotes ViewQuotes");
```

guiclient/prospects.cpp
```diff
@@ -78,42 +80,56 @@ void prospects::sNew()
...
 {
-  ParameterList params;
-  params.append("prospect_id", list()->id());
-  params.append("mode", "edit");
+  QList<XTreeWidgetItem*> selected = list()->selectedItems();
+  for (int i = 0; i < selected.size(); i++)
+  {
+    ParameterList params;
+    params.append("prospect_id", ((XTreeWidgetItem*)(selected[i]))->id());
+    params.append("mode", "edit");
```
gilmoskowitz: Many of these loops would be simpler if we had a `ParameterList::replace()` or faked it here with a `remove(); append();`

PR #1839, opened 2017-12-06
issue #31642 Remove Work Order Item characteristic usage

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3204.

guiclient/workOrder.cpp
```diff
@@ -112,7 +112,7 @@ workOrder::workOrder(QWidget* parent, const char* name, Qt::WindowFlags fl)
   ItemCharacteristicDelegate * delegate = new ItemCharacteristicDelegate(this);
   _itemcharView->setItemDelegate(delegate);
 
-  _wocharView->setType("WO");
+  _wocharView->setType("W");
```
andersdd: Not sure this is correct.  We are trying to display Item chars on a WO in one widget, and the "pure" WO chars in a second widget.  Something else is needed to achieve this I would suggest.

ChristopherCotnoir: This parallels how it works with Sales Order Items. There are two widgets on that screen as well, and there is no 'Sales Order Item Item' source. Both widgets use the same charuse_target_type. Same is true for everything else. Having it different here is inconsistent. It won't show up in the 'pure' WO chars widget if the 'Item' checkbox on the Characteristic screen is checked anyway.

PR #1848, opened 2017-12-15
issue #31730 - handle vendor group instead of cluster

gilmoskowitz:
the ap workbench uses a vendorgroup so the embedded vendor history
must do the same to keep pace.

requires xtuple/xtuple#3206

### DIFF
--- a/guiclient/apWorkBench.cpp
+++ b/guiclient/apWorkBench.cpp
@@ -75,15 +75,10 @@ apWorkBench::apWorkBench(QWidget* parent, const char* name, Qt::WindowFlags fl)
     _history = new dspVendorAPHistory(this, "dspVendorAPHistory", Qt::Widget);
     _apHistoryTab->layout()->addWidget(_history);
     _history->setCloseVisible(false);
-    _history->findChild<QWidget*>("_vendGroup")->hide();
     _history->findChild<DateCluster*>("_dates")->setStartNull(tr("Earliest"), omfgThis->startOfTime(), true);
     _history->findChild<DateCluster*>("_dates")->setEndNull(tr("Latest"),     omfgThis->endOfTime(),   true);
     _history->show();
-    VendorGroup *vend = _history->findChild<VendorGroup*>("_vend");
-    _vendorgroup->synchronize(vend);
-    connect(_vendorgroup, SIGNAL(newVendId(int)),          _history, SLOT(sFillList()));
-    connect(_vendorgroup, SIGNAL(newVendTypeId(int)),      _history, SLOT(sFillList()));
-    connect(_vendorgroup, SIGNAL(newTypePattern(QString)), _history, SLOT(sFillList()));
+    _vendorgroup->synchronize(_history->findChild<VendorGroup*>("_vend"));
   }
   else
     _apHistoryTab->setEnabled(false);
--- a/guiclient/apWorkBench.cpp
+++ b/guiclient/apWorkBench.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2017 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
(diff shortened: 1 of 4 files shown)


guiclient/apWorkBench.cpp
```diff
@@ -77,10 +77,13 @@ apWorkBench::apWorkBench(QWidget* parent, const char* name, Qt::WindowFlags fl)
     _history->setCloseVisible(false);
     _history->findChild<QWidget*>("_vendGroup")->hide();
```
ChristopherCotnoir: The synchronize function hides the synchronized VendorGroup widget. With some potential ui file simplification, it may be possible to remove this line now.

guiclient/apWorkBench.cpp
```diff
@@ -77,10 +77,13 @@ apWorkBench::apWorkBench(QWidget* parent, const char* name, Qt::WindowFlags fl)
...
     _history->findChild<DateCluster*>("_dates")->setStartNull(tr("Earliest"), omfgThis->startOfTime(), true);
-    _history->findChild<DateCluster*>("_dates")->setEndNull(tr("Latest"),	  omfgThis->endOfTime(),   true);
-    VendorCluster *histvend = _history->findChild<VendorCluster*>("_vend");
-    connect(histvend, SIGNAL(newId(int)), _history,      SLOT(sFillList()));
-    connect(_vendorgroup,  SIGNAL(newVendId(int)), histvend,      SLOT(setId(int)));
+    _history->findChild<DateCluster*>("_dates")->setEndNull(tr("Latest"),     omfgThis->endOfTime(),   true);
+    _history->show();
+    VendorGroup *vend = _history->findChild<VendorGroup*>("_vend");
+    _vendorgroup->synchronize(vend);
```
ChristopherCotnoir: These two lines can be combined to match above.

guiclient/apWorkBench.cpp
```diff
@@ -77,10 +77,13 @@ apWorkBench::apWorkBench(QWidget* parent, const char* name, Qt::WindowFlags fl)
...
+    connect(_vendorgroup, SIGNAL(newVendId(int)),          _history, SLOT(sFillList()));
+    connect(_vendorgroup, SIGNAL(newVendTypeId(int)),      _history, SLOT(sFillList()));
+    connect(_vendorgroup, SIGNAL(newTypePattern(QString)), _history, SLOT(sFillList()));
```
ChristopherCotnoir: It would be better to connect the updated signal instead of each of these. Additionally, this might be better moved to the dspVendorAPHistory window itself.

guiclient/displays/dspVendorAPHistory.cpp
```diff
@@ -66,7 +66,11 @@ enum SetResponse dspVendorAPHistory::set(const ParameterList &pParams)
 
   param = pParams.value("vend_id", &valid);
   if (valid)
-    _vend->setId(param.toInt());
+    _vend->setVendId(param.toInt());
+
+  param = pParams.value("vendtype_id", &valid);
+  if (valid)
+    _vend->setVendTypeId(param.toInt());
```
ChristopherCotnoir: Is the addition of this parameter really necessary here? I don't believe it is used.

guiclient/displays/dspVendorAPHistory.cpp
```diff
@@ -161,9 +165,7 @@ bool dspVendorAPHistory::setParams(ParameterList &params)
   if (isVisible())
   {
     QList<GuiErrorCheck> errors;
-    errors<< GuiErrorCheck(_vend->isVisible() && !_vend->isValid(), _vend,
-              tr("Please select a valid Vendor."))
```
ChristopherCotnoir: This is a good check. VendorGroup has a special isValid function, so there is no need to remove this, though the error text may need adjusting.

guiclient/apWorkBench.cpp
gilmoskowitz: The `updated()` signal is emitted whenever the user changes his/her selection of what the VendorGroup will show. Connecting to `VendorGroup::updated()` would requery whenever the user changes from all vendors to a single vendor, before s/he has had a chance to pick a vendor.

ChristopherCotnoir: I believe that is correct behavior. It is how all the other widgets function, if we don't want them updated for that, the VendorGroup widget itself needs changing. All widgets using VendorGroup rely on the updated signal to tell when to repopulate.

PR #1878, opened 2018-03-08
issue #31879 Cancel if save fails

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3236.

guiclient/group.cpp
```diff
@@ -236,7 +236,10 @@ void group::sSave()
      ;
 
   if (GuiErrorCheck::reportErrors(this, tr("Cannot Save Role"), errors))
+  {
+    sCancel();
     return;
+  }
```
gilmoskowitz: This will delete the group that was created in `set()`, which will then cascade to failures saving later after correcting the problem.

PR #1882, opened 2018-03-09
Fixes #31915 cc decline can delete cash receipt

bendiy:


guiclient/forwardUpdateAccounts.cpp
```diff
@@ -74,5 +75,10 @@ void forwardUpdateAccounts::sUpdate()
               " WHERE (NOT COALESCE(company_external,false));");
   
   forwardUpdate.exec();
+
+  if (!ErrorReporter::error(QtCriticalMsg, this, tr("Error updating accounts"),
+                            forwardUpdate, __FILE__, __LINE__))
+    QMessageBox::information(this, tr("Accounts Updated"),
+                             tr("All accounts were successfully updated."));
 }
```
bendiy: Not sure how this got in here...

guiclient/cashReceipt.cpp
```diff
@@ -164,6 +165,55 @@ cashReceipt::~cashReceipt()
...
+{
+  return _cashrcptid;
+}
+
+int cashReceipt::setId(int cashrcptId)
+{
+  _cashrcptid = cashrcptId;
+
+  return _cashrcptid;
+}
```
gilmoskowitz: Arbitrarily changing the internal id of a window managing a document can easily lead to data corruption.

bendiy: Removed.
I was going to use this to set id to `-1` before calling `close()`, but I found a workaround that allows me to call `sSave()` directly now.

PR #1883, opened 2018-03-12
updates to build with qt 5.10, refactor scriptapitest

scottminter:
DO NOT MERGE YET!


scottminter:
DO NOT MERGE

scriptapi/scriptapi.pro
```diff
@@ -274,30 +310,28 @@ SOURCES += setupscriptapi.cpp \
...
+              qwebsecurityoriginproto.h \
+              qwebsettingsproto.h \
+              qwebviewproto.h
+  SOURCES += qwebelementcollectionproto.cpp \
+              qwebelementproto.cpp \
+              qwebframeproto.cpp \
+              qwebpageproto.cpp \
+              qwebsecurityoriginproto.cpp \
+              qwebsettingsproto.cpp \
+              qwebviewproto.cpp
```
gilmoskowitz: - This looks like a duplication of the content at lines 165+ above. We only need one or the other.
- In a separate PR on the 4_11_x branch, please modify the functions in the deprecated .cpp files listed here to [report their deprecated status](https://github.com/xtuple/qt-client/blob/4_11_x/scriptapi/setupscriptapi.cpp#L320-L340) like the [script toolbox](https://github.com/xtuple/qt-client/blob/4_11_x/guiclient/scripttoolbox.cpp#L296) now does.

scriptapi/setupscriptapi.cpp
```diff
@@ -135,28 +136,44 @@
...
+
+#if QT_VERSION <= 0x050600
+  #include "qwebelementcollectionproto.h"
+  #include "qwebelementproto.h"
+  #include "qwebframeproto.h"
+  #include "qwebpageproto.h"
+  #include "qwebsecurityoriginproto.h"
+  #include "qwebsettingsproto.h"
+  #include "qwebviewproto.h"
+#endif
```
gilmoskowitz: `0x050600` is always less than `0x050900` so we only need one of the checks above. Make sure it matches the similar check in the `.pro` file, whichever way you decide to go.

scriptapi/setupscriptapi.cpp
```diff
@@ -135,28 +136,44 @@
...
+
+#if QT_VERSION > 0x050900
```
gilmoskowitz: This should be `>=`

PR #1886, opened 2018-03-13
issue #25224 - client-side metasql statement cache

gilmoskowitz:
- this reduces database load from requerying for known mql statements
- this introduces an abstract templated qhash class to build more caches

### DIFF
--- a/guiclient/guiclient.cpp
+++ b/guiclient/guiclient.cpp
@@ -820,7 +820,7 @@ void GUIClient::showEvent(QShowEvent *event)
     // START script code
       ParameterList params;
       params.append("jsonlist", "{\"1\": \"initMenu\"}");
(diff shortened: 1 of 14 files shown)


guiclient/guiclient.cpp
```diff
@@ -816,7 +820,7 @@ void GUIClient::showEvent(QShowEvent *event)
     // START script code
       ParameterList params;
       params.append("jsonlist", "{\"1\": \"initMenu\"}");
-      MetaSQLQuery mql = mqlLoad("scripts", "fetch");
+      MetaSQLQuery mql = _mqlhash->value("scripts", "fetch");
```
ChristopherCotnoir: value returns a QString, not a MetaSQLQuery. Should be

MetaSQLQuery mql(_mqlhash->value("scripts", "fetch");

Similar issues below.

PR #1891, opened 2018-03-16
issue #32496, issue #17968 - sort on user strings and preserve underlying raw codes

gilmoskowitz:
requires https://github.com/xtuple/xtuple/pull/3245

### DIFF
--- a/widgets/xtreewidget.cpp
+++ b/widgets/xtreewidget.cpp
@@ -994,6 +994,14 @@ bool XTreeWidgetItem::operator<(const XTreeWidgetItem &other) const
     QVariant v1 = data(sort.first, Xt::RawRole);
     QVariant v2 = other.data(sort.first, Xt::RawRole);
 
+    // bugs 17968 & 32496: sort strings according to user expectations AND preserve raw role
+    // here and case QVariant::String below: why not trust QVariant::toDouble()?
+    if (v1.toString().toDouble() == 0.0 && v2.toString().toDouble() == 0.0)
+    {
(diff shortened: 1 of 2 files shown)


widgets/xtreewidget.cpp
```diff
@@ -991,9 +991,23 @@ bool XTreeWidgetItem::operator<(const XTreeWidgetItem &other) const
...
     QVariant v2 = other.data(sort.first, Xt::RawRole);
 
+    // bugs 17968 & 32496: sort strings according to user expectations AND preserve raw role
+    if (v1.type() == QVariant::String && v1.toDouble(&ok) == 0.0 && ! ok &&
+        v2.type() == QVariant::String && v2.toDouble(&ok) == 0.0 && ! ok)
+    {
+      QVariant d1 = data(sort.first,       Qt::DisplayRole);
+      QVariant d2 = other.data(sort.first, Qt::DisplayRole);
+      if (d1.type() == QVariant::String && d1.toDouble(&ok) == 0.0 && ! ok &&
+          d2.type() == QVariant::String && d2.toDouble(&ok) == 0.0 && ! ok) {
```
ChristopherCotnoir: Coding Standards say brace should go on new line. It might be better to execute toDouble(&ok) on its own outside the if statements even though it requires multiple ok variables (ok1, ok2, etc.) simply because it makes it clear that comparing the result to 0.0 isn't actually a meaningful statement. Also, I recommend checking toString().toDouble(&ok) as well even though it really should be the same just in case, since both forms are used below. We can remove that for 4.12.

PR #1896, opened 2018-03-21
issue #32135 Allow user to change apply date

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3249.

PR #1897, opened 2018-03-22
issue #32190 Allow users to save orders over credit limit when customer is not yet…

ChristopherCotnoir:
… on warning/hold

### DIFF
--- a/guiclient/salesOrder.cpp
+++ b/guiclient/salesOrder.cpp
@@ -886,38 +886,18 @@ bool salesOrder::save(bool partial)
     }
 
 //  S/O Credit Check
-    if (_saving && _metrics->boolean("CreditCheckSOOnSave"))
+    if (_saving &&
+        _metrics->boolean("CreditCheckSOOnSave") && !creditLimitCheck() && _holdType->code() != "C")
     {
-      if (!creditLimitCheck())
-      {
-        if (_privileges->check("CreateSOForHoldCustomer"))
-        {
-          if(_holdType->code() != "C")
-          {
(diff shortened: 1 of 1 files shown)


gilmoskowitz:
This change enforces the distinction between creating the order that pushes the customer over the threshold and creating an order after the customer has already exceeded it. The goal is to avoid customer service reps spending time to create an order only to have their time wasted.

guiclient/salesOrder.cpp
```diff
@@ -886,38 +886,18 @@ bool salesOrder::save(bool partial)
...
-                                   "this order. You will need to edit the order to ensure "
-                                   "it falls within the credit limit or obtain a payment first." ) );
-        }
-      }
+      if (QMessageBox::question(this, tr("Sales Order Credit Check"),
+                      tr("<p>The customer has exceeded their credit limit "
+                         "and this order will be placed on Credit Hold.\n"
+                         "Do you wish to continue saving the order?"),
+                      QMessageBox::Yes,
+                      QMessageBox::No | QMessageBox::Default) == QMessageBox::Yes)
```
gilmoskowitz: This is a deprecated call to `question()`.

andersdd:
If I remember correctly, the original issue was to prevent orders being raised that would put a customer over their credit limit.  Unfortunately we cannot determine that until the order is ready to be saved and the order value is known.

gilmoskowitz:
@andersdd You're right - that was the original intent. We're finding in practice that people view the `CreditCheckSOOnSave` and `CreateSOForHoldCustomer` privileges differently than we anticipated and that this change more closely matches expectations.

PR #1898, opened 2018-03-22
Issue #27306 - credit card preflight checks

bendiy:
Add a new method to C++ ErrorReporter::getQueryStoredProcErrorMsg() that extracts our stored procedure RAISE EXCEPTION format string and calls storedProcErrorLookup() on it. This method support being called from the older style errorMsg() code in creditcardprocessor.cpp that does not use ErrorReporter currently.

Call the new canProcessPaymentGatewayTransaction() function before any payment gateway transaction is processed in creditcardprocessor.cpp.

Needs: https://github.com/xtuple/xtuple/pull/3253

### DIFF
--- a/common/storedProcErrorLookup.cpp
+++ b/common/storedProcErrorLookup.cpp
@@ -64,116 +64,93 @@ const struct {
                             "The order type is not valid for issue to shipping."),     0, "" },
 
   { "canProcessPaymentGatewayTransaction",
-    -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Invaid Customer or Payment Method."),
+    -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Invalid Customer or Payment Method."),
                           0, "" },
   { "canProcessPaymentGatewayTransaction",
-    -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Payment Method Bank Account is not mapped. "
-                          "Ensure as assignment is configured at: System > Setup > Creidt Cards"),
+    -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Payment Method Bank Account is not set up. "
+                          "Ensure an assignment is configured at: System > Setup > Credit Cards"),
                           0, "" },
   { "canProcessPaymentGatewayTransaction",
-    -3, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Payment Method Bank Account is not setup. "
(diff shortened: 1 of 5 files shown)


common/storedProcErrorLookup.cpp
```diff
@@ -60,9 +60,122 @@ const struct {
   { "calculateFreightDistribution", -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "The Voucher you are attempting to distribute freight "
                                     "to has incorrect item totals for the desired distribution method. Please check that Voucher"),     0, "" },
 
-  { "calcIssueToShippingLineBalance", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", 
+  { "calcIssueToShippingLineBalance", -1, QT_TRANSLATE_NOOP("storedProcErrorLookup",
                             "The order type is not valid for issue to shipping."),     0, "" },
 
+  { "canProcessPaymentGatewayTransaction",
+    -1, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Invaid Customer or Payment Method."),
```
gilmoskowitz: Typo - missing `l` in `Invalid`

common/storedProcErrorLookup.cpp
```diff
@@ -60,9 +60,122 @@ const struct {
...
+                          0, "" },
+  { "canProcessPaymentGatewayTransaction",
+    -2, QT_TRANSLATE_NOOP("storedProcErrorLookup", "Payment Method Bank Account is not mapped. "
+                          "Ensure as assignment is configured at: System > Setup > Creidt Cards"),
```
gilmoskowitz: Typos here and below
- `Ensure as`?
- `Creidt`

guiclient/creditcardprocessor.cpp
```diff
@@ -237,6 +238,9 @@ static struct {
   { -99, QT_TRANSLATE_NOOP("CreditCardProcessor", "The CVV value is not valid.")				},
   {-100, QT_TRANSLATE_NOOP("CreditCardProcessor", "No approval code was received:\n%1\n%2\n%3")		},
 
+  // accounting setup errors
+  {-110, QT_TRANSLATE_NOOP("CreditCardProcessor", "Credit Card Processing Accounting Configuration Error: %1") },
```
gilmoskowitz: No need to state `Credit Card Processing` here as many of the configuration problems are not specific to this case.

PR #1899, opened 2018-03-22
Update code to compile with msvc

mmicr001:
Some code changes that enable qt-client to be built using MSVC


guiclient/guiclient.pro
```diff
@@ -23,6 +25,8 @@ INCLUDEPATH += ../scriptapi \
                ../../xtuple-build-desktop/widgets/tmp/lib \
                ../../xtuple-build-desktop/guiclient . \
                $(CSVIMP_HEADERS)/csvimpcommon $(CSVIMP_HEADERS)/plugin
+			   C:\QtNew\qt-everywhere-opensource-src-5.5.1\qtbase\src\3rdparty\zlib
+			   $$[QT_INSTALL_HEADERS]/QtZlib
```
gilmoskowitz: This is not going to work:

- no backslash at the end of line 27 means the next two lines aren't added to `INCLUDEPATH`
- everyone may have a different Qt installation directory
- paths will be messed up on non-Windows builds

guiclient/guiclient.pro
```diff
@@ -64,7 +68,7 @@ OBJECTS_DIR = tmp/obj
 win32 {
   win32-msvc*:LIBS += -lshell32
   RC_FILE = rcguiclient.rc
-  LIBS += -lz
+  #LIBS += -lz
```
gilmoskowitz: The implication is that we don't need to link the compression library on Windows. That doesn't sound right.

guiclient/moc/mocinclude.tmp
```diff
@@ -0,0 +1,64 @@
+-IC:/QtNew/qt-everywhere-opensource-src-5.5.1/qtbase/mkspecs/win32-msvc2015
```
gilmoskowitz: This file should not be committed.

guiclient/openSalesOrders.cpp
```diff
@@ -43,7 +43,7 @@ openSalesOrders::openSalesOrders(QWidget* parent, const char*, Qt::WindowFlags f
   optionsWidget()->hide();
 
   _dates->setStartNull(tr("Earliest"), omfgThis->startOfTime(), true);
-  int confDays = _metrics->value("DefaultOrderStartDays").toInt() ? : -90;
+  int confDays = _metrics->value("DefaultOrderStartDays").toInt() ? NULL : -90;
```
gilmoskowitz: Wow. That's a surprising bug! That line shouldn't have compiled for the past year. If we have a default, let's use it:
```c++
int confDays = _metrics->value("DefaultOrderStartDays").toInt()
             ? _metrics->value("DefaultOrderStartDays").toInt() : -90;
```

scriptapi/qsslerrorproto.cpp
```diff
@@ -113,7 +113,7 @@ QScriptValue constructQSslError(QScriptContext *context, QScriptEngine  *engine)
   }
   else if (context->argumentCount() == 1) {
     QScriptValue arg = context->argument(0);
-    if (arg.isString() or arg.isNumber()) {
+    if (arg.isString() || arg.isNumber()) {
```
gilmoskowitz: And this shouldn't have built for 2 years.

gilmoskowitz:
OK to merge after 4.11.3 is tagged.

PR #1904, opened 2018-03-29
Issue #32686 - Lookup -16 for Tax Hold error.

bendiy:


guiclient/issueToShipping.cpp
```diff
@@ -941,7 +941,7 @@ void issueToShipping::sFillList()
       else if (issueFillList.value("holdtype").toString() == "T")
       {
         QMessageBox::critical( this, tr("Cannot Issue Stock"),
-                              storedProcErrorLookup("issuetoshipping", -15));
+                              storedProcErrorLookup("issuetoshipping", -16));
```
bendiy: Change `-15` to `-16` to match:
https://github.com/xtuple/qt-client/blob/4_11_x/common/storedProcErrorLookup.cpp#L1098

PR #1910, opened 2018-04-06
issue #32648 Properly set empty state

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3265.

PR #1912, opened 2018-04-16
issue #32823 Save after changing Credit By

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/private-extensions#1005.

PR #1916, opened 2018-04-17
issue #30503 Handle error when contact is used on incident

ChristopherCotnoir:


ChristopherCotnoir:
It can't be, it's not a foreign key that is being violated, it's a not null constraint. See the deletecntct function.

PR #1919, opened 2018-04-19
issue #30503 Add bulk replace to contacts list

ChristopherCotnoir:


guiclient/contacts.cpp
```diff
@@ -191,6 +195,227 @@ void contacts::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem *, int)
...
+      if (qry.first())
+      {
+        label = "New " + qry.value("char_name").toString() + ":";
+        chartype = (characteristic::Type)(qry.value("char_type").toInt());
+      }
+      else if (ErrorReporter::error(QtCriticalMsg, this, tr("Characteristic lookup failed"),
+                                    qry, __FILE__, __LINE__))
+        return;
+    }
+  }
```
gilmoskowitz: Can't this just use the header item's text? Then there's less translation to worry about, no `char` query required, and less code to maintain when something changes. If we want to make the label explicit, which this code does via `"New " + stuff`, then handle translation with something like this:
```
label = tr("New value for %1").arg(header[column]->text);
```

guiclient/contacts.cpp
```diff
@@ -191,6 +195,227 @@ void contacts::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem *, int)
...
+      {
+        replace = newdlg.getDate();
+        ok = true;
+      }
+    }
+  }
+
+  if (ok && QMessageBox::warning(this, tr("Bulk Replace?"),
+                                 tr("<p>The selected value will be replaced for all selected "
+                                    "items. Are you sure you want to do this?"),
```
gilmoskowitz: It would be nice to include the number of selected items here.

guiclient/contacts.cpp
```diff
@@ -191,6 +195,227 @@ void contacts::sPopulateMenu(QMenu *pMenu, QTreeWidgetItem *, int)
...
+                  "                   WHERE charass_target_type = 'CNTCT' "
+                  "                     AND charass_target_id = cntct_id "
+                  "                     AND charass_char_id = :char_id) "
+                  "  AND cntct_id IN (";
+      updateStr = "UPDATE charass "
+                  "   SET charass_value=:replace "
+                  " WHERE charass_target_type='CNTCT' "
+                  "   AND charass_char_id = :char_id "
+                  "   AND charass_target_id IN (";
+    }
```
gilmoskowitz: MetaSQL with `<? foreach ?>` for `INSERT` and `UPDATE` should make the code more compact and readable here as well.

guiclient/xclusterinputdialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+      _cluster = new CRMAcctCluster(this, "_cluster");
+      this->findChildren<QHBoxLayout*>()[0]->insertWidget(0, _cluster);
+    } 
+    else if (param.toString() == "addr")
+    {
+      _cluster = new AddressCluster(this, "_cluster");
+      this->findChildren<QHBoxLayout*>()[0]->insertWidget(0, _cluster);
+    } 
+  }
```
gilmoskowitz: This would be easier to maintain if the `type` were the full classname or if we could pass in an instance of a VirtualCluster from the caller. With the full classname, we might even be able to [construct them on the fly](http://doc.qt.io/qt-5/qmetaobject.html#newInstance).

PR #1921, opened 2018-04-20
issue #30503 Add contact marking

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3274.

PR #1939, opened 2018-05-16
[31859] Resolve SQL error when cancelling POs

andersdd:
Plus refresh the Open PO list to clear temporary entry


gilmoskowitz:
jenkins - retest this please

PR #1947, opened 2018-05-24
[32856] Properly replace system locale with user locale if language is English

ChristopherCotnoir:


guiclient/main.cpp
```diff
@@ -330,13 +330,18 @@ int main(int argc, char *argv[])
...
   foreach (f, transfile)
   {
     foreach (QString l, lang)
     {
-      if (translator->load(translationFile(l, f.first, f.second)))
+      if (f.first == "xTuple" && l.split("_")[0] == "en")
+      {
+        app.removeTranslator(translatorSys);
+        break;
+      }
```
gilmoskowitz: - The `en` check isn't sufficient as folks might want to "translate" American English into Canadian or British or ... English
- There is a rare use case where people replace our terminology into their own site- or company-specific jargon. It would be good to find a way to retain that use case while still addressing the bug as reported.

PR #1955, opened 2018-05-29
[30500] Prevent orphaned SO headers, plus refresh order list on delete

andersdd:
Prevent scenario where user can delete SO items and leave the SO in a closed, orphaned state

### DIFF
--- a/guiclient/salesOrder.cpp
+++ b/guiclient/salesOrder.cpp
@@ -3138,7 +3138,7 @@ bool salesOrder::deleteForCancel()
       return false;
   }
 
-  if ((_mode == cNew && !_captive) || (ISORDER(_mode) && _soitem->topLevelItemCount() == 0))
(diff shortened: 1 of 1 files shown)


guiclient/salesOrder.cpp
```diff
@@ -3138,8 +3138,7 @@ bool salesOrder::deleteForCancel()
       return false;
   }
 
-  if (_mode == cNew &&
-      !_captive)
+  if ((_mode == cNew && !_captive) || (!_metrics->boolean("EnableRentals") && ISORDER(_mode) && _soitem->topLevelItemCount() == 0))
```
gilmoskowitz: Referring to an extension from the core should be avoided.

andersdd: Normally I would agree, however rentals works a certain way which allows for no lines on an order and is already accommodated in the Sales Order a little further up the code in the same manner.   How would you suggest we handle this situation?

PR #1959, opened 2018-06-01
[29289] Allow editing of BOM Item Sequence numbers

andersdd:


guiclient/bomItem.cpp
```diff
@@ -352,6 +357,9 @@ void bomItem::sSave()
   bomitem.bindValue(":bomitem_issuemethod", _issueMethods[_issueMethod->currentIndex()]);
   bomitem.bindValue(":bomitem_schedatwooper", false);
 
+  if (_mode == cEdit)
+    bomitem.bindValue(":bomitem_seqnumber", _seqnumber->toDouble());
```
gilmoskowitz: `toDouble()` should be `toInt()`, or `QSpinBox::value()`.

guiclient/bomItem.cpp
```diff
@@ -125,6 +126,8 @@ enum SetResponse bomItem::set(const ParameterList &pParams)
     if (param.toString() == "new")
     {
       _mode = cNew;
+      _seqnumberLit->setVisible(false);
+      _seqnumber->setVisible(false);
```
gilmoskowitz: Hiding these when creating a new BOM item seems unnecessary. If a trigger would overwrite these values on `INSERT` then we should fix the trigger to only supply a value when `NEW.bom_seqnumber IS NULL`.

andersdd: Sort of agree but due to the introduction of a `QSpinBox` that would require introducing a repeat of the numbering mechanism in the client itself before saving.  So the client automatically assigns the sequence number on `INSERT` but you can modify either using edit or the move up/down buttons

guiclient/bomItem.ui
```diff
@@ -517,7 +480,7 @@ to be bound by its terms.</comment>
           <enum>Qt::StrongFocus</enum>
          </property>
          <property name="type">
-          <enum>Comments::BOMItem</enum>
+          <number>Comments::BOMItem</number>
```
gilmoskowitz: The element name here should be `enum`

(and — separate project [33606] — we _really_ need to fix the widgets to make this problem go away)

PR #1964, opened 2018-06-11
[33238] Prevent ability to receipt into inactive item sites.

andersdd:
Also prevents amending an item type if the item site validation checks would fail

requires xtuple/xtuple#3313

### DIFF
--- a/guiclient/item.cpp
+++ b/guiclient/item.cpp
@@ -482,6 +482,7 @@ void item::saveCore()
 void item::sSave()
 {
   XSqlQuery itemSave;
+  XSqlQuery checks;
   QList<GuiErrorCheck> errors;
   QString sql;
   QString itemNumber = _itemNumber->text().trimmed().toUpper();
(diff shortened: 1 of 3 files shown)


guiclient/item.cpp
```diff
@@ -693,7 +694,76 @@ void item::sSave()
...
+            "  FROM planord "
+            "  JOIN itemsite ON itemsite_id=planord_itemsite_id "
+            " WHERE itemsite_item_id=<? value('item_id') ?>"
+            "UNION "
+            "SELECT planreq_id "
+            "  FROM planreq "
+            "  JOIN itemsite ON itemsite_id=planreq_itemsite_id "
+            " WHERE itemsite_item_id=<? value('item_id') ?>"
+            "<? endif ?>"
+            "LIMIT 1; ";
```
gilmoskowitz: This might better be encapsulated in a function, or at least a metasql table record, so we can update it more easily based on edition & extensions. It also removes the need for the [itemsite query above](https://github.com/xtuple/qt-client/pull/1964/files#diff-525a15918e79bd9ee5a12412a09dbcddR690).

PR #1968, opened 2018-06-18
Freight and Misc handling

ChristopherCotnoir:


ChristopherCotnoir:
It's to tell the application that the value put in that field is intended to be an order level discount, which apparently some customers have been doing and which the xtuple code needs to know as that requires special handling. See the calculatetax function in the xtuple PR.

PR #1976, opened 2018-06-21
issue #32336 -  correct sort order indicator in xtreewidget

mmicr001:
Just a simple swap of the 2 arrow png files in xtreewidget.cpp

### DIFF
--- a/widgets/xtreewidget.cpp
+++ b/widgets/xtreewidget.cpp
@@ -1061,7 +1061,7 @@ bool XTreeWidgetItem::operator<(const XTreeWidgetItem &other) const
         else if (v1.toString().toDouble() == 0.0 && v2.toDouble(&ok2)) //v1 is string, v2 is number
           returnVal = false; //the number should always be treated as greater than a string
         else if (v1.toDouble(&ok1) && v2.toString().toDouble() == 0.0)
-          returnVal = true
+          returnVal = true;
         else
           returnVal = (v1.toDouble() < v2.toDouble()); 
         break;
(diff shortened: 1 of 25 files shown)


ChristopherCotnoir:
This switches the arrows completely, which is a change in behavior from how it has always worked and thus something we do not want. The bug states that the default has changed, meaning the direction the arrow is on first sort, that is the regression that needs to be fixed.

widgets/xtreewidget.cpp
```diff
@@ -1052,13 +1052,19 @@ bool XTreeWidgetItem::operator<(const XTreeWidgetItem &other) const
...
         else if (v1.toString().toDouble() == 0.0 && v2.toDouble(&ok2)) //v1 is string, v2 is number
-          returnVal = false; //the number should always be treated as greater than a string
+		{qDebug() << "S #";
+		returnVal = false;} //the number should always be treated as greater than a string
         else if (v1.toDouble(&ok1) && v2.toString().toDouble() == 0.0)
-          returnVal = true;
+		{returnVal = true; qDebug() << "# S";}
         else
-          returnVal = (v1.toDouble() < v2.toDouble());
+		{returnVal = (v1.toDouble() < v2.toDouble()); qDebug() << "# #";}
```
gilmoskowitz: 3 inadvertent changes that add `qDebug()` calls

widgets/xtreewidget.cpp
```diff
@@ -1717,7 +1723,8 @@ void XTreeWidget::sHeaderClicked(int column)
   if (_roles.size() <= 0)
   {
     _sort.clear();
-    _sort.append(qMakePair(column, header()->sortIndicatorOrder()));
+	
+	_sort.append(qMakePair(column, header()->sortIndicatorOrder()));
```
gilmoskowitz: Here and elsewhere - be careful about changes to whitespace. The indentation here doesn't match the surrounding code and may lead to problems down the road.

PR #1987, opened 2018-07-11
[32453] Add option to suppress automatic supply order creation, fix issue whe…

ChristopherCotnoir:
…re only first supply order is created from places other than sales order item screen


ChristopherCotnoir:
Requires xtuple/xtuple#3337.

PR #2004, opened 2018-07-25
Working port of QtWebkit to QtWebEngine

davidbeauchamp:
As the title says, this PR removes our dependence on QtWebkit and replaces it with QtWebEngine. The classes that no longer exist have been removed and the webengine equivalents of the other bits we used have been exposed to script. ~~This PR is incomplete in the sense that I have not gone over the WebEngine classes to expose any new functionality yet, just removed the things that Webkit used to have but WebEngine doesn't. I will review the documentation for them and see if there are new things to expose.~~ I have completed exposing the rest of the three QWebEngine classes we've used up to this point and cleaned up the Webkit stuff that doesn't exist anymore. I suspect this is more of a 5_0_x PR but I leave that up to you.

[This part in xtdesktop](https://github.com/xtuple/xtdesktop/blob/4_0_x/resources/client/scripts/initMenu.js#L79) just becomes QWebEngineView instead of QWebView. The only code xtdesktop uses that doesn't exist anymore is [this enum](https://github.com/xtuple/xtdesktop/blob/4_0_x/resources/client/scripts/initMenu.js#L107) to delegate all links, which I don't think even matters as chromium tends to open pages well on its own.


scriptapi/qwebenginepageproto.cpp
```diff
@@ -0,0 +1,608 @@
...
+
+QString QWebEnginePageProto::selectedText() const
+{
+  QWebEnginePage *item = qscriptvalue_cast<QWebEnginePage*>(thisObject());
+  if (item)
+    return item->selectedText();
+  return QString();
+}
+
+void QWebEnginePageProto::setAudioMuted(bool muted)
```
gilmoskowitz: Here and elsewhere (~13 * 2 places in this file alone) — one mistake we've made in the past with these prototypes is adding wrappers for the getter and setter methods for Qt properties. It would be really nice if we could let simple properties remain simple properties. Thus:

```JavaScript
var page = new QWebEnginePage(mwindow);
page.setAudioMuted(true);     // bad
page.audioMuted = true;       // good
...
if (! page.isAudioMuted()) { ... } // bad
if (! page.audioMuted)     { ... } // good
```

If we can do that here, there should be less code here, fewer scripts to be remediated in the future, and a model to follow in the other `*proto.cpp`s.

This may be really easy, as we might have overlooked [one documented path](http://doc.qt.io/qt-5/qscriptengine.html#newQMetaObject) a long time ago.

davidbeauchamp: Take a look at [qwebenvineviewproto.h](https://github.com/xtuple/qt-client/blob/88593d55e4b42771d701ee35842680d3b47b6f14/scriptapi/qwebengineviewproto.h) and [qwebengineviewproto.cpp](https://github.com/xtuple/qt-client/blob/88593d55e4b42771d701ee35842680d3b47b6f14/scriptapi/qwebengineviewproto.cpp) -- I switched to using the QMetaObject there. Basically anything thats QObject related, such as the properties, slots and signals, are exposed automatically using the QMetaObject, and things like read/write properties work, and you don't need the to/from scriptvalue functions anymore. 

However, anything that is a public function, i.e. things that we need to mark as `Q_INVOKABLE`, still need a wrapper to the base class, which you can apply as the default prototype for the QWebEngineView after its constructor is registered with the script engine. 

Overall it does cut out quite a bit of boilerplate code.

davidbeauchamp:
I will resolve the conflicts tomorrow.. this PR was switched to 5_0_x and I did not rebase my local branch before making a new commit.

davidbeauchamp:
Gave QWebEnginePage the same treatment, a bit more apparent how much code we can trim out.

gilmoskowitz:
Outstanding! Thank you for following through on this. Our scripting API will be much more consistent and easier to manage with this information.

davidbeauchamp:
hrm, merging in 5_0_x made some changes to the project files I need to correct. Doing so now.

PR #2010, opened 2018-07-27
[33100] Create WO checkbox in Sales Order Line was not working.  

andersdd:
Reinstate logic to create WOs. Also added hard check to prevent duplicate WOs from being created from SO lines (there was another issue raised about this).  This hard check needs to be confirmed.

requires xtuple/xtuple#3362

### DIFF
--- a/guiclient/salesOrderItem.cpp
+++ b/guiclient/salesOrderItem.cpp
@@ -2770,15 +2770,45 @@ void salesOrderItem::sHandleSupplyOrder()
       }
 
       // check _supplyOrderType to determine type of order
-      if (_supplyOrderType == "W" && (_supplyOrderId > 0) && _item->isConfigured())
+      if (_supplyOrderType == "W")
       {
-        XSqlQuery implodeq;
-        implodeq.prepare( "SELECT implodeWo(:wo_id, true) AS result;" );
-        implodeq.bindValue(":wo_id", _supplyOrderId);
-        implodeq.exec();
+        ordq.prepare( "SELECT createWo(:orderNumber, itemsite_id, :qty, itemsite_leadtime, :dueDate,"
+                      "                :comments, :parent_type, :parent_id ) AS result, itemsite_id "
+                      "  FROM itemsite "
+                      " WHERE itemsite_item_id=:item_id"
+                      "   AND itemsite_warehous_id=:warehous_id "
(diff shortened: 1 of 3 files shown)


andersdd:
@ChristopherCotnoir can you please review this change.  I had to revert some changes you made in 4.11.  Some logic in your change was flawed and would never work, but the reason for removal of WO creation was not clear and caused this issue.

ChristopherCotnoir:
The supply order creation code was removed from the C++ code because it already exists on the coitem trigger. The supply order is created by saving the coitem with a coitem_order_id of -1 which causes the trigger to create the supply order. The issue is this was only done for INSERT and not UPDATE so unchecking and rechecking isn't working correctly. Simply fixing the UPDATE bindings should fix the problem.

andersdd:
@ChristopherCotnoir updated trigger as suggested so both INSERT/UPDATE creates supply orders as necessary.  Removed some extraneous code and SQL calls.

Had to do a bit of work around Supply Order Type as this was only determined once when initially selecting the Item, and was NULLed when updating the order line (if supply order unchecked).  Code now retains the supply order and relies on supplyorderid setting to determine whether to recreate the order.

requires updated trigger in xtuple/xtuple#3362

guiclient/salesOrderItem.cpp
```diff
@@ -1133,7 +1133,8 @@ void salesOrderItem::sSave(bool pPartial)
                "       :soitem_cos_accnt_id, :soitem_rev_accnt_id, :soitem_dropship "
                "FROM itemsite "
                "WHERE ( (itemsite_item_id=:item_id)"
-               " AND (itemsite_warehous_id=:warehous_id) );" );
+               " AND (itemsite_warehous_id=:warehous_id) ) "
+               "RETURNING * ;" );
```
ChristopherCotnoir: If we do this, PO and PR creation will have to be moved into the before trigger, otherwise the RETURNING statement will return -1 for coitem_order_id since RETURNING does not take after trigger changes into account. Also avoid using *, just get what we need.

andersdd:
Reworked much of this process to get the timings correct and to handle Supply Order creation on both New Sales Orders and after editing.  Reworked workflow of process to ensure SO screen is updated immediately for the user.  Unchecking Create Supply Order and rechecking now recreates the supply order appropriately.  Fixed several latent bugs found when testing this PR

PR #2016, opened 2018-08-01
[10370] Add Search and Filters to CRM Address List

andersdd:
Some of this is already in version 5.0 so we need to watch for conflicts when merging 4.12 into 5.0

requires xtuple/xtuple#3367


gilmoskowitz:
OK to merge when xtuple/xtuple#3367 is merged.

PR #2026, opened 2018-08-07
[33559] fix designer crashes

gilmoskowitz:
Replaced qobject_casts that always succeeded with qobject_cast calls and
guarded dereferencing of the resulting pointers. Tested by creating a
uiform with the embedded designer that contains every xTuple custom widget

### DIFF
--- a/widgets/scriptablewidget.cpp
+++ b/widgets/scriptablewidget.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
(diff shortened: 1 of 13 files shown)


widgets/scriptablewidget.cpp
```diff
@@ -73,6 +73,12 @@ QScriptEngine *ScriptableWidget::engine()
 
 void ScriptableWidget::loadScript(const QStringList &list)
 {
+  if (! _guiClientInterface)
+  {
+    qDebug() << "No GUIClientInterface to ask for scripts";
+    return;
+  }
+
```
andersdd: In testing - this change to the `scriptableWidget.cpp` file was all that was necessary to prevent the internal xTuple screen editor from segfaulting the entire application.

PR #2035, opened 2018-08-09
[11525] Encryption key generation feature.  Creates and populates initial key

andersdd:
create initial key with a hash.

Also adds information about distribution of the key and the consequences of altering the key.


andersdd:
@gilmoskowitz If you read the Reference Guide we suggest very specific key text requirements such as at least 16 chars, mix of upper/lower/numbers and no special characters.  I would argue my automatic hash meets those requirements (and exceeds the length) and we are gaining nothing by letting some user enter it in the client except complexity.  I can easily add a QInputDialog for user entry, but I dont think it is necessary.  They _can_ still edit the text file if the _really_ want their own value in the file.

PR #2073, opened 2018-09-06
[29272] Streamline lot/serial traceability on Transfer Orders. 

andersdd:
Also resolves a bunch of bugs around lot/serial on TO Receipts.
Streamlines the process and makes receiving lots on TOs much easier for end-users.
Fixed characteristic/expiry bug
Locks down Lot receipt so you cannot receive more than shipped

requires xtuple/private-extensions#1034

### DIFF
--- a/guiclient/createLotSerial.cpp
+++ b/guiclient/createLotSerial.cpp
@@ -120,8 +120,20 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
       {
         // First check for "pre-assigned/associated" lots. A couple example use cases:
         // 1) Receiving TO - get lot(s) from issue to shipping trans. 2) Receiving RA - get lot(s) from original SO (if exists).
+        // Only display lots with qty remaining available so we dont double allocate
         XSqlQuery origTransLots;
-        origTransLots.prepare("SELECT ls_id, ls_number, ls_number FROM ls WHERE ls_id IN (SELECT UNNEST(getAssocLotSerialIds(:itemlocdist_id)));");
+        origTransLots.prepare("SELECT ls_id, ls_number, ls_number, qtys - "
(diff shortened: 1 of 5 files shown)


guiclient/createLotSerial.cpp
```diff
@@ -120,8 +120,20 @@ enum SetResponse createLotSerial::set(const ParameterList &pParams)
...
+                              "            WHERE itemlocdist_source_id=:itemlocdist_id "
+                              "              AND ls_id=itemlocdist_ls_id), 0) AS qty "
+                              "  FROM ls "
+                              "  JOIN (SELECT * FROM getassoclotserialids(:itemlocdist_id)) foo "
+                              "  ON ls_id=foo.ids "
+                              " WHERE qtys - "
+                              "       COALESCE((SELECT itemlocdist_qty FROM itemlocdist "
+                              "                  WHERE itemlocdist_source_id=:itemlocdist_id "
+                              "                    AND ls_id=itemlocdist_ls_id), 0) > 0 "
+                              " ORDER BY ls_number;" );
```
gilmoskowitz: Simplify:
```SQL
WITH dist AS (SELECT itemlocdist_ls_id, itemlocdist_qty
                FROM itemlocdist
               WHERE itemlocdist_source_id = :itemlocdist_id)
SELECT ls_id, ls_number, ls_number, qtys - itemlocdist_qty
  FROM ls
  JOIN getAssocLotSerialIds(:itemlocdist_id) ON ls_id = ids
  JOIN dist ON ls_id = itemlocdist_ls_id
 WHERE qtys > itemlocdist_qty
 ORDER BY ls_number;
```

andersdd: Your suggestion wont work (returns no rows).  I have to compare across rows so need to use a sub-query.

PR #2075, opened 2018-09-06
[11131] Add CSV support to Export Data Functionality

mmicr001:
Added:
- functionality so that user can select any delimiter that he wants
- an editable combobox to  System > Preferences window, with a list of delimiter values from which the user must select a default.
- a window that prompts user to select export file type and delimiter (when applicable)
- ability to manually enter any file extension (in cases where user wants a .tsv or other extension not offered in our list)

### DIFF
--- a/guiclient/userPreferences.cpp
+++ b/guiclient/userPreferences.cpp
@@ -1,7 +1,7 @@
 /*
  * This file is part of the xTuple ERP: PostBooks Edition, a free and
  * open source Enterprise Resource Planning software suite,
- * Copyright (c) 1999-2014 by OpenMFG LLC, d/b/a xTuple.
+ * Copyright (c) 1999-2018 by OpenMFG LLC, d/b/a xTuple.
  * It is licensed to you under the Common Public Attribution License
  * version 1.0, the full text of which (including xTuple-specific Exhibits)
  * is available at www.xtuple.com/CPAL.  By using this software, you agree
@@ -29,6 +29,7 @@
 #include "timeoutHandler.h"
 #include "translations.h"
 #include "dictionaries.h"
+#include "xtreewidget.h"
 
 extern QString __password;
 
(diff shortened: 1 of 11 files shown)


widgets/xtreewidget.h
```diff
@@ -211,7 +212,8 @@ class XTUPLEWIDGETS_EXPORT XTreeWidget : public QTreeWidget
     Q_INVOKABLE XTreeWidgetItem         *findXTreeWidgetItemWithId(const XTreeWidgetItem *ptreeitem, const int pid);
 
     Q_INVOKABLE QString toTxt() const;
-    Q_INVOKABLE QString toCsv() const;
+    Q_INVOKABLE QString toSV(QString pSep) const;
+    Q_INVOKABLE QString toCsv() {return toSV(QString(','));} ;
```
gilmoskowitz: Please define (put the code for) `toCsv()` [in the .cpp file, not the header](https://github.com/xtuple/qt-client/wiki/xTuple-Coding-Standards#c). This is a change we're propagating throughout the C++ code so there are still exceptions to it in some header files. Let's fix 'em when we see 'em and avoid adding more.

widgets/xtreewidget.cpp
```diff
@@ -2880,3 +2889,54 @@ void XTreeWidget::keyPressEvent(QKeyEvent* e)
...
+    else if (notEncouraged.contains(delim))
+    {
+      QMessageBox msgBox;
+      msgBox.setText("Be careful with your selection of delimiter");
+      msgBox.setInformativeText("We suggest you avoid using these delimiters : @#$%&*_=-+/? ");
+      msgBox.exec();
+    }
+  } 
+  return invalid;
+}
```
gilmoskowitz: Good!
Now consolidate:
```C++
bool invalidDelim(QString delim)
{
  QString msg; // note that this starts out as a null string
  QString disallowed  = ...;
  QString discouraged = ...;
  bool   valid = false;
  if (delim.length() > 1 && delim != "{tab}")
    msg = tr("Avoid delimiters more than 1 char long."); // tr() requires a QObject context; the end result may look different
  else if (disallowed.contains(delim))
    msg = tr("%1 is not permitted as a delimiter. The following characters are not allowed: %2").arg(delim, disallowed);
  else if (discouraged.contains(delim))
  {
    msg = tr("%1 may cause problems. We suggest avoiding the following: %2").arg(delim, discouraged);
    valid = true;
  }
  else
    valid = true;

  if (! msg.isEmpty())   // this reduces repetition above and also
  {                      // makes it easier to replace the message handler
    QMessageBox msgBox;  // if necessary
    ....
  }
  return valid;
}
```

guiclient/userPreferences.cpp
```diff
@@ -354,23 +354,36 @@ void userPreferences::sSave(bool close)
 
   _pref->set("IdleTimeout", _idleTimeout->value());
   omfgThis->_timeoutHandler->setIdleMinutes(_idleTimeout->value());
-
-  if(ExportHelper::validDelim(_delimiter->currentText()))
+  
+  //////// check delimiter /////////////////////////////////////////
+  enum delimCheck{valid=0, tooLong=1, disallowed=2, disencouraged=3};
```
gilmoskowitz: Define the enum in `common/exporthelper.h` as part of the `ExportHelper` class. Then it can be returned by `ExportHelper::validDelim()`, used here, and used again in `ExportOptions::sSave()` without duplication.

PR #2077, opened 2018-09-10
Merge4_12_x to 5_0_x 20180910

gilmoskowitz:


gilmoskowitz:
accompanies xtuple/xtuple#3413 and xtuple/private-extensions#1035

PR #2083, opened 2018-09-14
[33786] Add multi-select support to the Billing Approvals list.

bendiy:


guiclient/dspBillingSelections.cpp
```diff
@@ -102,45 +102,52 @@ void dspBillingSelections::sPost()
...
   {
-    int result = dspPost.value("result").toInt();
-    if (result < 0) {
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Posting Billing Selection(s)"),
-                               storedProcErrorLookup("createInvoice", result),
-                               __FILE__, __LINE__);
+    dspPost.prepare("SELECT cobmisc_cohead_id AS sohead_id "
+              "FROM cobmisc "
+              "WHERE (cobmisc_id = :cobmisc_id)");
+    dspPost.bindValue(":cobmisc_id", item->id());
```
gilmoskowitz: Please remove this query and the `soheadid` variable. This query sets the variable which is then not used again.

guiclient/dspBillingSelections.cpp
```diff
@@ -102,45 +102,52 @@ void dspBillingSelections::sPost()
...
-
-    omfgThis->sInvoicesUpdated(result, true);
-    omfgThis->sSalesOrdersUpdated(soheadid);
-    omfgThis->sBillingSelectionUpdated(soheadid, true);
-
-    sFillList();
+    }
+
+    dspPost.prepare("SELECT createInvoice(:cobmisc_id) AS result;");
+    dspPost.bindValue(":cobmisc_id", item->id());
```
gilmoskowitz: Please `dspPost.prepare()` before the loop. This saves one db call and one query planning step per loop iteration.

guiclient/dspBillingSelections.cpp
bendiy: The `soheadid` variable is used below by:
```
omfgThis->sSalesOrdersUpdated(soheadid);
```

I'll combine the two queries.

PR #2104, opened 2018-10-02
[33833] Changing Scheduled Dates on Supply Order S/Os caused SQL error

andersdd:
Caused by missing ';'s between SQL calls.  Updated code to split out queries per coding standards.


gilmoskowitz:
jenkins - retest this please
the qt directory shouldn't have disappeared like that

PR #2109, opened 2018-10-08
[24368] Add scroll ability to characteristics in customer window in View mode

mmicr001:
The customer window did not make use of the characteristicswidget. I got rid of most of the old code, but still need to check if I maybe deleted more than I should have or not enough.
Once the characteristicswidget is used, scrolling in view mode is done using setReadOnly(true).


mmicr001:
Customer Characteristics are different enough than other characteristics, such that the characteristicswidget cannot be used

PR #2110, opened 2018-10-09
Expose QSettings to script

davidbeauchamp:


davidbeauchamp:
This is the only thing stopping me from removing our reliance on trojitas gui

scriptapi/qsettings.cpp
```diff
@@ -0,0 +1,295 @@
...
+             context->argument(2).isString() &&
+             context->argument(3).isString() &&
+             context->argument(4).isObject())
+  {
+    obj = new QSettings(qscriptvalue_cast<QSettings::Format>(context->argument(0).toInt32()),
+                        qscriptvalue_cast<QSettings::Scope>(context->argument(1).toInt32()),
+                        context->argument(2).toString(),
+                        context->argument(3).toString(),
+                        qobject_cast<QObject*>(context->argument(4).toQObject()));
+  }
```
gilmoskowitz: There are actually 11 constructors for [QSettings](http://doc.qt.io/qt-5/qsettings.html). For example, the [const QString &, const QString &, QObject*](http://doc.qt.io/qt-5/qsettings.html#QSettings) constructor is one C++ method with 3 variations because of the default values:

```C++
else if (context->argumentCount() == 1 && context->argument(0).isString())
  obj = new QSettings(context->argument(0).toString());
else if (context->argumentCount() == 2 && context->argument(0).isString() && context->argument(1).isString())
  obj = new QSettings(context->argument(0).toString(), context->argument(1).toString());
else if (context->argumentCount() == 3 && context->argument(0).isString() && context->argument(1).isString() && context->argument(2).isObject())
  obj = new QSettings(context->argument(0).toString(), context->argument(1).toString(), qobject_cast<QObject*>(context->argument(3).toQObject()));
// or
switch (context->argumentCount())
{
  case 0: obj = new QSettings();
          break;
  case 1: if (context->argument(0).isString())
            obj = new QSettings(context->argument(0).toString());
          else if (context->argument(1).isObject())
            obj = new QSettings(qobject_cast<QObject*>(context->argument(1).toQObject()));
          break;
// or ...
```

The end of `setupQSettingsProto()`  can then read:
```C++
if (obj)
  return engine->newQObject(obj);
context->throwError(QScriptContext::UnknownError, "Could not find an appropriate QSettings constructor.");
```
to avoid multiple `throwError()`s and strange `else` or `default:` blocks.

scriptapi/qsettings.cpp
```diff
@@ -0,0 +1,295 @@
...
+    item->fileName();
+  return QString();
+}
+
+QSettings::Format QSettingsProto::format() const
+{
+  QSettings *item = qscriptvalue_cast<QSettings*>(thisObject());
+  if (item)
+    return item->format();
+  return QSettings::NativeFormat; // the default for each OS
```
gilmoskowitz: 👍

davidbeauchamp:
Wound up being 13 overloads for QSettings(). I implemented them all, but switched the check to `(!obj)`, compiler was complaining about possible lack of return in non-void function otherwise.

davidbeauchamp:
Ahh boooo, on 5.5.1 I have to add the explicit exposure of each enum value. The 5.11.x qmetaobject does it all for you, but on 5.5.1 none of the enums get exposed unless I register them against the constructor. Doing so now.

PR #2127, opened 2018-10-18
msvs has problems with bulk translation commands in makefiles

gilmoskowitz:


xtuple.pro
```diff
@@ -18,7 +18,6 @@ SUBDIRS = common \
 CONFIG += ordered
 
 TRANSLATIONS = share/dict/xTuple.ar_eg.ts \
-               share/dict/xTuple.base.ts \
```
ChristopherCotnoir: Removing this will prevent lupdate from updating this file. Is it possible to just remove the qm after lrelease instead?

PR #2132, opened 2018-10-23
Resolve 4.12 merge into 5.0 regression

andersdd:


andersdd:
@gilmoskowitz sorry to be difficult, but this issue is popping up in several ZENQ issues so we need to get this merged into 5.0

PR #2135, opened 2018-10-28
[Version 5.0] Resolve errant user messages in sales order item

andersdd:


andersdd:
Decided issues are also present in 4.12, so going to create a PR for that release (and upmerge to 5.0)

PR #2145, opened 2018-11-05
[33376] VoucherItem window mishandles distributions when user cancels Voucher Item

mmicr001:
-Added appropriate responses to user for most combinations of Tagged receipts and Distribution Amounts
 during **save** and **cancel**
-created `saveVoItemInformation()` function, because the 'tagged' status was only being stored when the `Save` button was being clicked. But we now have cases where we untag receipts on `Cancel`


ChristopherCotnoir:
There can be more than one distribution per voucher item, which is why there is a list, as long as they have different cost elements.

guiclient/voucherItem.h
```diff
@@ -14,6 +14,8 @@
 #include "guiclient.h"
 #include "xdialog.h"
 #include "ui_voucherItem.h"
+#include <vector>
```
ChristopherCotnoir: This can be removed.

PR #2146, opened 2018-11-06
[34100] Pricing Schedule Update initial list populate

andersdd:


ChristopherCotnoir:
This approach works but is somewhat confusing and unnecessarily complicated. The problem is simply that the window opens without the set function being called and populate is only called in the set function. A better fix would be to simply add a populate call to the constructor. Or, if we wish to avoid querying twice for list prices, we can simply call the set function and pass in a ParameterList with listpricesched set to false and check listpricesched's value rather than existence in the set function, which is more clear than adding an entirely new parameter.

andersdd:
Comments by @ChristopherCotnoir addressed.

guiclient/updatePrices.cpp
```diff
@@ -90,14 +90,15 @@ enum SetResponse updatePrices::set(const ParameterList &pParams)
...
   {
-    _listpricesched = true;
-    setWindowTitle(tr("List Pricing Schedule Assignments"));
+    if (param.toBool())
+    {
+      _listpricesched = true;
+      setWindowTitle(tr("List Pricing Schedule Assignments"));
+    }
+    else
+      _showCurrent->setChecked(true);
```
ChristopherCotnoir: This isn't necessary, it's already checked. Merely ending up in set so that populate is called is sufficient.

andersdd:
That's weird.  When I originally looked at this issue the checkbox was not being set to checked.

PR #2147, opened 2018-11-07
[33376]  VoucherItem window mishandles distributions when user cancels Voucher Item

mmicr001:
- Only rollback when the user has not saved a voItem yet (_voitemid == -1)
- delete all distributions when rolling back
- check if `Qty to Voucher == 0` or `no distribution items` when saving or canceling

Also added check to postVoucher function in xtuple repo.
-added translations to all exceptions in postVoucher function

### DIFF
--- a/guiclient/voucherItem.cpp
+++ b/guiclient/voucherItem.cpp
@@ -745,10 +745,9 @@ void voucherItem::reject()
     ;
     if (GuiErrorCheck::reportErrors(this, tr("Cannot Cancel Voucher Item"), errors))
     {
-      int ret =  QMessageBox::question(this, tr("Cancel Anyway?"),
+      if(QMessageBox::question(this, tr("Cancel Anyway?"),
                                           tr("Would you like to cancel anyway?"),
-                                          QMessageBox::Yes | QMessageBox::No);
-      if(ret == 65536) // 16384 = 0x00010000  = "No"
+                                          QMessageBox::Yes | QMessageBox::No)) == QMessageBox::No;
         return;
     }
   } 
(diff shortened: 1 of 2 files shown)


guiclient/voucherItem.cpp
```diff
@@ -693,22 +693,66 @@ void voucherItem::rollback()
...
+                          tr("You must enter a postive Quantity to Voucher."))
+          << GuiErrorCheck(!voucherSave.first(), _qtyToVoucher,
+                          tr("You must make at least one distribution for this Voucher Item."))
+    ;
+    if (GuiErrorCheck::reportErrors(this, tr("Cannot Cancel Voucher Item"), errors))
+    {
+      int ret =  QMessageBox::question(this, tr("Cancel Anyway?"),
+                                          tr("Would you like to cancel anyway?"),
+                                          QMessageBox::Yes | QMessageBox::No);
+      if(ret == 65536) // 16384 = 0x00010000  = "No"
```
ChristopherCotnoir: You don't need to use the integer equivalent, just use the enum same as above, replacing 65536 with QMessageBox::No. This is much easier to read. You can also combine this with the QMessageBox call, i.e. if (QMessageBox::question(...) = QMessageBox::No).

guiclient/voucherItem.cpp
```diff
@@ -745,10 +745,9 @@ void voucherItem::reject()
     ;
     if (GuiErrorCheck::reportErrors(this, tr("Cannot Cancel Voucher Item"), errors))
     {
-      int ret =  QMessageBox::question(this, tr("Cancel Anyway?"),
+      if(QMessageBox::question(this, tr("Cancel Anyway?"),
                                           tr("Would you like to cancel anyway?"),
-                                          QMessageBox::Yes | QMessageBox::No);
-      if(ret == 65536) // 16384 = 0x00010000  = "No"
+                                          QMessageBox::Yes | QMessageBox::No)) == QMessageBox::No;
```
ChristopherCotnoir: The parentheses before the == should be at the end.

PR #2155, opened 2018-11-09
[32051] Inactive item selectable on project task

mmicr001:
Made changes to type/search/list so that only active items are displayed by default.
-type: only active items are displayed (if the user wants to select an inactive item, they must list or search and check the 'Show Inactive' checkbox

This fix doesn't disallow the user from selecting an inactive item for the task


widgets/itemCluster.cpp
```diff
@@ -582,6 +582,7 @@ void ItemLineEdit::sHandleCompleter()
     numQ.prepare(QString("SELECT *"
                          "  FROM (%1) data"
                          " WHERE (POSITION(:number IN item_number)=1)"
+                         "   AND item_active"
```
gilmoskowitz: This should take `_itemType & ItemLineEdit::cActive` into account.

widgets/itemCluster.cpp
```diff
@@ -594,6 +595,7 @@ void ItemLineEdit::sHandleCompleter()
 
     QStringList clauses;
     clauses = _extraClauses;
+    clauses << " (AND item_active) ";
```
gilmoskowitz: This should take `_itemType & ItemLineEdit::cActive` into account.

widgets/itemCluster.cpp
```diff
@@ -1144,10 +1146,30 @@ void itemList::sSearch(const QString &pTarget)
...
   _listTab->clear();
+  
   if (_useQuery)
-  {
-    _listTab->populate(_sql, _itemid);
+  { 
+    QString whereClause = "";
+    if(clauses.size()>0)
+      whereClause = " WHERE" + clauses.join(" AND ");
+    QString sql = "SELECT * FROM (" + _sql + ") AS dummy " + whereClause + ";" ;
```
gilmoskowitz:
```suggestion
    QString sql = "SELECT * FROM (" + _sql.remove(";") + ") AS dummy " + whereClause + ";" ; 
```

gilmoskowitz:
The build is failing with
```
itemCluster.cpp:589:19: error: no match for ‘operator&’ (operand types are ‘QString’ and ‘ItemLineEdit::Type’)
     if( _itemType & ItemLineEdit::cActive)
                   ^
```
and again on line 604. Apparently `_itemType` is the wrong class member to use.

mmicr001:
The query used in the xtte extension did not filter on active_item. If we add that clause to the query then we would break the list and search functionality for the specific ItemCluster in xtte.

PR #2157, opened 2018-11-11
[33969] Resolve contact merge schema conflict

andersdd:
There is one other conflict in `contacts.cpp #375` but that will be resolved with some other dependent changes I need to make.


ChristopherCotnoir:
The changes in this PR are already in xtuple/qt-client#259, so this can be closed.

PR #2158, opened 2018-11-12
[33827] Attaching file from UNC path raises DB exception

mmicr001:
This bug only occurred when `Save to Database` was checked. The `setPath()` function would produce a null value for the url value causing the exception.


mmicr001:
jenkins - retest this please

PR #2159, opened 2018-11-13
CRM Marketing improvements and bugfixes.

andersdd:
-Improve UI for Contact-Account assignments
-Improve CRM Group Utility features
-Bug fixes with 4.12 upmerge

requires xtuple/xtuple#3489

@ChristopherCotnoir I had to remove your recent "Mass Replace" function on the Contacts page for CRM Account only as a mass replace of the account on contact did not make sense in a many-to-many relationship and also ignored the other attributes (role/default/active) that are part of that assignment.  It's possible using the screen I added in this PR we _might_ be able to reinstate it but it will take a bit more thought how to make it work.


guiclient/contacts.cpp
```diff
@@ -474,10 +479,27 @@ void contacts::sEdit()
...
 
-    contact* newdlg = new contact(0, "", false);
-    newdlg->set(params);
-    newdlg->setAttribute(Qt::WA_DeleteOnClose);
-    newdlg->show();
+    contact newdlg(0, "", false);
+    newdlg.set(params);
+    newdlg.setAttribute(Qt::WA_DeleteOnClose);
+    if (newdlg.exec() != XDialog::Rejected)
+      sFillList();
```
ChristopherCotnoir: This shouldn't be changed. We want the user to be able to use Edit to open multiple contacts at once, but this forces them to do it one at a time.

Also, if we do change this, we absolutely MUST remove Qt::WA_DeleteOnClose or else the entire app will crash when closing the window. That attribute is for preventing memory leaks with dynamic allocation.

guiclient/contactAccountAssign.ui
```diff
@@ -0,0 +1,223 @@
...
+       </item>
+      </layout>
+     </item>
+     <item row="1" column="1">
+      <layout class="QHBoxLayout" name="horizontalLayout">
+       <item>
+        <widget class="CRMAcctCluster" name="_accnt">
+         <property name="enabled">
+          <bool>false</bool>
+         </property>
```
ChristopherCotnoir: Having enabled false and readOnly false is inconsistent. It seems we want this widget enabled by default, so this should be removed because if readOnly is applied first it won't override the enabled setting.

guiclient/crmGroups.h
```diff
@@ -44,12 +46,12 @@ struct GroupMap
   }
 };  
 
-class crmGroups : public XWidget, public Ui::crmGroups
+class crmGroups : public XDialog, public Ui::crmGroups
```
ChristopherCotnoir: Why is this being changed to an XDIalog? Also, this is used in places other than buildCRMGroups above and so if this is going to be changed the calls there should be changed as well. Currently handleNewWindow warns if passed a dialog, I'm not sure why but there is probably a reason it shouldn't be done.

guiclient/crmGroups.cpp
```diff
@@ -107,6 +108,11 @@ enum SetResponse crmGroups::set(const ParameterList &pParams)
   return NoError;
 }
 
+void crmGroups::sClose()
+{
+  XDialog::done(1);
```
ChristopherCotnoir: Why is the close button being routed to an accepted state? This seems wrong.

andersdd: This is for creating new Groups.  The combobox is populated from existing groups at time of construction.  This is a "New" button that opens the dialog and allows a user to create new groups.  The parent function waits till the dialog is closed then refreshes the combobox.

ChristopherCotnoir: Ah, so the idea isn't to change the type but to simply refresh. OK, then using populate() instead of setType would make that more clear.

I think simply not checking the value of exec() is less likely to cause problems than overriding standard QDialog behavior. Additionally, it will cause the combobox to still refresh if the window is x'd out instead of using the close button. In this case, we always want to refresh, regardless of what happens to close the window.

Also, menuSales, menuCRM, menuSystem, and employee (and any other windows that open this one that I missed) should have their calls changed.

guiclient/contactAccountAssign.ui
andersdd: Removed ReadOnly and tidied up logic

guiclient/crmGroups.cpp
andersdd: This is an unusual scenario where we are opening the equivalent of a CRM groups list dialog.  This does not have a _save()_ function as this screen opens yet another child screen.  The intention here is to trigger the combobox to refresh regardless of outcome in the crmGroups screen, hence the `done()` signal when closing this list.  I will update the process in the menus however.

guiclient/crmGroups.h
andersdd: Done as suggested

PR #2160, opened 2018-11-14
Minor tweak to display to allow setupCharacteristics multiple times

andersdd:
If setupCharacteristics() is called more than once in a screen or script, duplicate column errors occur.  This is needed in the Inventory Days package where we rebuild the list columns before each fillList() due to variable columns based on filter selection.

Minor change that checks whether the arrays already contain the characteristic and appends if they do not.

Might need porting back to 4.12 depending on customer requirement.

### DIFF
--- a/guiclient/display.cpp
+++ b/guiclient/display.cpp
@@ -295,11 +295,14 @@ void displayPrivate::setupCharacteristics(QStringList uses)
     }
     else if (chartype == characteristic::Date)
     {
-      _charidsdate.append(chars.value("char_id").toInt());
-      QString start = QApplication::translate("display", "Start Date", 0);
-      QString end = QApplication::translate("display", "End Date", 0);
-      _parameterWidget->append(name + " " + start, column + "startDate", ParameterWidget::Date);
-      _parameterWidget->append(name + " " + end, column + "endDate", ParameterWidget::Date);
+      if (!_charidslist.contains(chars.value("char_id").toInt()))
+      {
+        _charidsdate.append(chars.value("char_id").toInt());
+        QString start = QApplication::translate("display", "Start Date", 0);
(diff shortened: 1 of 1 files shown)


guiclient/display.cpp
```diff
@@ -266,23 +266,32 @@ void displayPrivate::setupCharacteristics(QStringList uses)
...
+        QString sql = QString("SELECT charopt_value, charopt_value"
+                              "  FROM charopt"
+                              " WHERE charopt_char_id = %1"
+                              " ORDER BY charopt_order, charopt_value;")
+            .arg(chars.value("char_id").toInt());
+        _parameterWidget->append(name, column, ParameterWidget::Multiselect, QVariant(), false, sql);
+      }
     }
     else if (chartype == characteristic::Date)
     {
```
ChristopherCotnoir: Same change should also be done here.

andersdd: Well spotted. Fixed now.

guiclient/display.cpp
```diff
@@ -266,31 +266,43 @@ void displayPrivate::setupCharacteristics(QStringList uses)
...
-      _charidsdate.append(chars.value("char_id").toInt());
-      QString start = QApplication::translate("display", "Start Date", 0);
-      QString end = QApplication::translate("display", "End Date", 0);
-      _parameterWidget->append(name + " " + start, column + "startDate", ParameterWidget::Date);
-      _parameterWidget->append(name + " " + end, column + "endDate", ParameterWidget::Date);
+      if (!_charidslist.contains(chars.value("char_id").toInt()))
```
ChristopherCotnoir: Wrong parameter name, should be _charidsdate.

andersdd: Bah!  Thanks - fixed that now

PR #2162, opened 2018-11-16
[33126] Currency edits were not updating the currency combobox

andersdd:
Trigger notification was only refreshing the first XComboBox to use that notification

Added a secondary notification trigger that can be called by specific tables that need this (only applied to curr_symbol so far)

requires xtuple/xtuple#3492


andersdd:
Closing as an alternate (simpler) solution has been found

PR #2172, opened 2018-11-20
[33615] create sales order without Default Ship To 

mmicr001:
This bug was reported in the Distribution Edition, but the fix needs to be applied here as well.
The main fix is in `salesOrder.cpp`. 
The changes in `shipTo.cpp` make sure that if a ship-to is inactive then it can't be the default ship-to

### DIFF
--- a/guiclient/salesOrder.cpp
+++ b/guiclient/salesOrder.cpp
@@ -2078,8 +2078,7 @@ void salesOrder::populateShipto(int pShiptoid)
                     "       shipto_preferred_warehous_id, "
                     "       shipto_salesrep_id, shipto_commission AS commission"
                     "  FROM shiptoinfo"
-                    " WHERE shipto_active " 
-                    "   AND shipto_id = :shipto_id;" );
+                    " WHERE shipto_id = :shipto_id;" );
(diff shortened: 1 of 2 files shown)


guiclient/salesOrder.cpp
```diff
@@ -2078,7 +2078,8 @@ void salesOrder::populateShipto(int pShiptoid)
                     "       shipto_preferred_warehous_id, "
                     "       shipto_salesrep_id, shipto_commission AS commission"
                     "  FROM shiptoinfo"
-                    " WHERE shipto_id = :shipto_id;" );
+                    " WHERE shipto_active " 
+                    "   AND shipto_id = :shipto_id;" );
```
gilmoskowitz: This will do the wrong thing when viewing an old, completed sales order that used a now-inactive ship-to.

mmicr001: You are right. The change made to `shipto.cpp` that doesn't allow an inactive ship-to to be the default is the reason my testing seemed fine (I didn't try to open an existing SO). Previously, ship-tos that were inactive were allowed to still be the default (thus causing the bug)

PR #2173, opened 2018-11-21
Set new characteristic input mask to empty/null

andersdd:
as it was defaulting to first item in combobox


ChristopherCotnoir:
The null option should always be the top option of the list but here it is the bottom, which could cause problems. The coding standards say not to populate comboboxes in the ui files which is what is happening here. That is likely the cause of the issues. Populate the combobox in the constructor with append instead, that should fix the issue more reliably.

andersdd:
Updated code as suggested

PR #2174, opened 2018-11-21
[32188] Cancel not working on TE invoice date

mmicr001:
the `xdateinputdialog` didn't have slots to deal with the `ok` or `cancel` causing the bug.
Changes needed to be made to the corresponding file in `xtte` to rollback in case of `cancel`


ChristopherCotnoir:
It looks like these slots already exist in the ui file.

PR #2202, opened 2019-01-09
[28731] Ellipses is used improperly in desktop

mmicr001:
Second pass through, should cover all popup menus in core.


guiclient/salesOrder.cpp
```diff
@@ -1534,7 +1534,7 @@ void salesOrder::sPopulateMenu(QMenu *pMenu)
 
       menuItem = pMenu->addAction(tr("Return Stock"), this, SLOT(sReturnStock()));
       menuItem->setEnabled(_privileges->check("IssueStockToShipping"));
-      menuItem = pMenu->addAction(tr("Issue Stock..."), this, SLOT(sIssueStock()));
+      menuItem = pMenu->addAction(tr("Issue Stock"), this, SLOT(sIssueStock()));
```
gilmoskowitz: Ellipsis should remain here since we open the issueLineToShipping window here (compared with "Issue Line Balance" that comes next, which won't).

guiclient/unpostedCreditMemos.cpp
```diff
@@ -87,16 +87,16 @@ void unpostedCreditMemos::sPopulateMenu(QMenu *pMenu)
...
 
-  menuItem = pMenu->addAction(tr("View..."), this, SLOT(sView()));
+  menuItem = pMenu->addAction(tr("View"), this, SLOT(sView()));
   menuItem->setEnabled(_privileges->check("MaintainCreditMemos") ||
                        _privileges->check("ViewCreditMemos"));
 
   pMenu->addSeparator();
 
-  menuItem = pMenu->addAction(tr("Post..."), this, SLOT(sPost()));
+  menuItem = pMenu->addAction(tr("Post"), this, SLOT(sPost()));
```
gilmoskowitz: Ellipsis should remain here since we ask the user for the G/L distribution date. Consider making it optional depending on `_privileges->check("ChangeSOMemoPostDate")`.

guiclient/workOrder.cpp
```diff
@@ -2183,34 +2183,34 @@ void workOrder::sPopulateMenu(QMenu *pMenu,  QTreeWidgetItem *selected)
...
       {
-        menuItem = pMenu->addAction(tr("Issue..."), this, SLOT(sIssueMatl()));
+        menuItem = pMenu->addAction(tr("Issue"), this, SLOT(sIssueMatl()));
         if (!_privileges->check("IssueWoMaterials"))
           menuItem->setEnabled(false);
       }
       if (status == "I")
       {
-          menuItem = pMenu->addAction(tr("Return..."), this, SLOT(sReturnMatl()));
+          menuItem = pMenu->addAction(tr("Return"), this, SLOT(sReturnMatl()));
```
gilmoskowitz: Ellipsis should remain (compare sReturnMatl() with sReturnMatlBatch())

PR #2203, opened 2019-01-09
Add missing connections

ChristopherCotnoir:
Some of the signals with XTreeWidget instead of QTreeWidget as arguments that were supposed to be added way back by https://github.com/xtuple/qt-client/commit/2e8251335f06400d418223cb64fad21073c29618 never actually had their slots connected so they still can't be used in scripts. This finally fixes that.


davidbeauchamp:
jenkins, retest this please

davidbeauchamp:
jenkins, retest this please

PR #2205, opened 2019-01-11
[34370] Need xcombobox code to be country_abbr for new country system

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/qt-client#2204 and xtuple/custom#207.

PR #2208, opened 2019-01-17
[33967] product cost not updating to invoice

mmicr001:
After speaking with Jim and Mike, it seems that adjusting the item cost from a sales order creates too many accounting complications regardless of the costing method. The user should not be allowed to edit the cost item at this stage


guiclient/salesOrderItem.ui
```diff
@@ -2358,6 +2361,9 @@ to be bound by its terms.</comment>
               <property name="alignment">
                <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
               </property>
+              <property name="enabled">
+                <bool>false</bool>
+              </property>
```
ChristopherCotnoir: This field should still be enabled. It's the entire reason for the unit cost field's existence.

PR #2218, opened 2019-02-08
add script exposure to tax integration

gilmoskowitz:
and fix some compiler warnings

### DIFF
--- a/common/avalaraIntegration.h
+++ b/common/avalaraIntegration.h
@@ -25,6 +25,9 @@ class AvalaraIntegration : public TaxIntegration
     void wait();
 
   protected:
+    using TaxIntegration::error;
+    using TaxIntegration::handleResponse;
+
     virtual void        sendRequest(QString, QString, int, QString, QStringList, QString);
     virtual QString     error(QString, QNetworkReply*, QJsonObject);
     virtual void        done();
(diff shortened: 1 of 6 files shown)


ChristopherCotnoir:
TaxIntegration is a global object like _metrics. There should only be one running at a time (a couple exceptions, but that gets complicated). Scripts should use the existing global object rather than creating a new one in almost all circumstances, so the global object in guiclient needs to be exposed.

Everything else looks more or less fine, so far as I could tell.

andersdd:
Not sure if this is what Chris is talking about, but I am running tax calculation on a document (Voucher) by calling a NOTIFY.  The users currently running client has a LISTEN built in which triggers the appropriate events (use internal tax calculation, or call Avalara API).  This works fine, but I have some secondary events that have to occur _after_ the Voucher tax has been saved.  So my script needs access to the emit taxCalculated (or something else) to understand when the next steps are available to proceed.

gilmoskowitz:
With the second commit you should now be able to get the global `taxIntegration` object in a script and connect to it:
```javascript
taxIntegration['taxCalculated(double, QString'].connect(sMyScriptedSlotFunction);
```

andersdd:
@gilmoskowitz Thanks, the signal is working in scripts now.

FYI the script actually requires the following pattern:
`taxIntegration.taxCalculated.connect(sMyScriptedFunction);`

PR #2219, opened 2019-02-10
[34506] Resolve incorrect GL postings for manually distributed PO reject

andersdd:
PO Reject records were being updated before the Voucher Item was created.  This follows the same path as Received records and updates once the voitem is saved.

Should be able to directly backport to 4.12

### DIFF
--- a/guiclient/voucherItem.cpp
+++ b/guiclient/voucherItem.cpp
@@ -401,7 +401,6 @@ void voucherItem::sSave()
   }
 
   //Save 'tagged' status stored in rev_vohead_id to recv_voitem_id
-
   voucherSave.prepare( "UPDATE recv "
                        "SET recv_voitem_id=CASE WHEN (recv_vohead_id IS NULL) THEN NULL ELSE :voitem_id END "
                        "WHERE ( (NOT recv_invoiced) "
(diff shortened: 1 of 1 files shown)


guiclient/voucherItem.cpp
```diff
@@ -420,6 +419,23 @@ void voucherItem::sSave()
     return;
   }
 
+  //Save 'tagged' status stored in poreject_vohead_id to poreject_voitem_id
+  voucherSave.prepare( "UPDATE poreject "
+                       "   SET poreject_voitem_id=CASE WHEN (poreject_vohead_id IS NULL) THEN NULL ELSE :voitem_id END "
+                       "WHERE  poreject_poitem_id=:poitem_id "
+                       "  AND  poreject_posted "
+                       "  AND  poreject_vohead_id IS NULL OR poreject_vohead_id=:vohead_id;" );
```
gilmoskowitz: [AND has higher precedence than OR](https://www.postgresql.org/docs/9.5/sql-syntax-lexical.html#SQL-PRECEDENCE) so the `poreject_vohead_id` comparisons won't work quite as expected. Wrap the `OR` in parens.

PR #2220, opened 2019-02-11
[34405] Align version 4.12 with published compatibility matrix

andersdd:


ChristopherCotnoir:
Includes earliest, excludes latest, so should be 11.0 (or something like that, are we properly handling postgres' switch from MAJ.MIN.PAT to MIN.PAT? Had to make changes to the plv8 script for that).

andersdd:
@ChristopherCotnoir I went with the _published_ compatibility matrix.  Regarding the change to Postgres 10 versioning, I figured this issue related to 4.12, and we can make appropriate changes in the 5.0 series to the upper postgres versioning and versioning scheme.

gilmoskowitz:
Chris is right -- the compatibility matrix says 4.12.0-alpha works with 10.x but this code change excludes all of 10.x.

andersdd:
Amended `latest` version check value

PR #2228, opened 2019-02-18
Ssave signals

interhawk:
Adds signals to sSave in purchase order item and sales order item.
Enum for setting failed status is not yet working in scripts.

### DIFF
--- a/guiclient/salesOrderItem.cpp
+++ b/guiclient/salesOrderItem.cpp
@@ -1160,14 +1160,16 @@ void salesOrderItem::sSave(bool pPartial)
     salesSave.bindValue(":soitem_dropship", _supplyDropShip->isChecked());
     salesSave.exec();
 
-    if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Item Information"),
-                             salesSave, __FILE__, __LINE__))
+    if (salesSave.lastError().type() != QSqlError::NoError)
     {
       emit saveBeforeRollback(&salesSave);
       rollback.exec();
       emit saveAfterRollback(&salesSave);
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Item Information"),
+                           salesSave, __FILE__, __LINE__);
       return;
     }
+
   }
   else if ( (_mode == cEdit) || ((_mode == cNew) && _partialsaved) )
   {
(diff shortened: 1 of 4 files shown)


guiclient/salesOrderItem.cpp
```diff
@@ -1146,11 +1160,12 @@ void salesOrderItem::sSave(bool pPartial)
...
     salesSave.exec();
 
-    if (salesSave.lastError().type() != QSqlError::NoError)
+    if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Item Information"),
+                             salesSave, __FILE__, __LINE__))
     {
+      emit saveBeforeRollback(&salesSave);
       rollback.exec();
-      ErrorReporter::error(QtCriticalMsg, this, tr("Error Saving Item Information"),
-                           salesSave, __FILE__, __LINE__);
```
ChristopherCotnoir: Don't move this; it's after the rollback so that the transaction isn't held open while the user has the dialog open.

guiclient/salesOrderItem.cpp
```diff
@@ -1252,6 +1268,9 @@ void salesOrderItem::sSave(bool pPartial)
       else if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Retrieving Item Information"),
                                     salesSave, __FILE__, __LINE__))
       {
+        emit saveBeforeRollback(&salesSave);
+        rollback.exec();
+        emit saveAfterRollback(&salesSave);
```
ChristopherCotnoir: These also need to go above the ErrorReporter call above.

guiclient/salesOrderItem.cpp
interhawk: fixed

guiclient/salesOrderItem.cpp
interhawk: fixed

ChristopherCotnoir:
There's no way to interrupt a function being run, so I'm pretty sure the only thing that can be done is to set a variable from the script. We can use an isFailed() function in place of _saveStatus == Failed, but I don't think we can do much else.

I think the code could be made cleaner by using a rollback() function instead of having the rollback handling be several lines. We could also put checking for an error and the ErrorReporter call in there, although we'd presumably want to pass in __FILE__ and __LINE__ which is kind of awkward.

interhawk:
Agree with Chris. We could potentially create 2 slots: setScriptFailedStatus, setScriptOKStatus which just set a simple boolean and not have to deal with the enum if we wanted to, but I don't see a way around setting some kind of flag on the main object.
User interaction has been placed outside the transaction in Purchase Order Item.

PR #2263, opened 2019-03-12
[34164] Refund credit card when voiding cash receipt

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/xtuple#3592, associated with xtuple/payment-gateways#63.

PR #2265, opened 2019-03-12
[34489] Split out Employee deletion SQL

andersdd:
This code was masking an underlying SQL error preventing the deletion of employees.

I originally did not replicate the error in version 4.12 but this was due to my Db not having `xtte` installed.  This change can possibly be backported to 4.12.  Related change to `xtte` package possibly makes this change redundant, but updated the code to current standards regardless.

requires xtuple/xtuple#3595

### DIFF
--- a/guiclient/employees.cpp
+++ b/guiclient/employees.cpp
@@ -84,15 +84,39 @@ void employees::sDelete()
                             QMessageBox::No | QMessageBox::Default) == QMessageBox::No)
     return;
 
+  XSqlQuery rollbackq;
+  rollbackq.prepare("ROLLBACK;");
+  XSqlQuery begin("BEGIN;");
+
   XSqlQuery delq;
-  delq.prepare("DELETE FROM charass WHERE charass_target_type = 'EMP' AND charass_target_id = :emp_id;"
-               "DELETE FROM empgrpitem WHERE groupsitem_reference_id = :emp_id;"
-               "DELETE FROM emp WHERE emp_id = :emp_id;");
+  delq.prepare("DELETE FROM charass WHERE charass_target_type = 'EMP' AND charass_target_id = :emp_id;");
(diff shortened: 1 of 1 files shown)


guiclient/employees.cpp
```diff
@@ -84,15 +84,39 @@ void employees::sDelete()
...
 
+  XSqlQuery rollbackq;
+  rollbackq.prepare("ROLLBACK;");
+  XSqlQuery begin("BEGIN;");
+
   XSqlQuery delq;
-  delq.prepare("DELETE FROM charass WHERE charass_target_type = 'EMP' AND charass_target_id = :emp_id;"
-               "DELETE FROM empgrpitem WHERE groupsitem_reference_id = :emp_id;"
-               "DELETE FROM emp WHERE emp_id = :emp_id;");
+  delq.prepare("DELETE FROM charass WHERE charass_target_type = 'EMP' AND charass_target_id = :emp_id;");
```
gilmoskowitz: This is already [done in the trigger](https://github.com/xtuple/xtuple/blob/5_0_x/foundation-database/public/trigger_functions/emp.sql#L193-L196) so we don't need it here.

guiclient/employees.cpp
```diff
@@ -84,15 +84,39 @@ void employees::sDelete()
...
     return;
+  }
+  delq.prepare("DELETE FROM empgrpitem WHERE groupsitem_reference_id = :emp_id;");
+  delq.exec();
+  if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Deleting Employee"),
+                           delq, __FILE__, __LINE__))
+  {
+    rollbackq.exec();
+    return;
+  }
```
gilmoskowitz: Consider adding this to the trigger as well.

andersdd: Ancillary deletion logic moved to triggers as suggested.

PR #2268, opened 2019-03-14
Fix tax setup crash

ChristopherCotnoir:


common/avalaraIntegration.cpp
```diff
@@ -185,9 +185,9 @@ void AvalaraIntegration::handleResponse(QNetworkReply* reply)
     QJsonObject responseJson = QJsonDocument::fromJson(response).object();
     replies.removeOne(reply);
     TaxIntegration::handleResponse(type, orderType, orderId, QString::fromUtf8(response), error(type, reply, responseJson));
-  }
 
-  delete reply;
+    delete reply;
+  }
```
gilmoskowitz: The [QNetworkAccessManager docs](https://doc.qt.io/qt-5/qnetworkaccessmanager.html) say this (emphasis added):

> Note: After the request has finished, it is the responsibility of the user to delete the QNetworkReply object at an appropriate time. **Do not directly delete it inside the slot connected to finished()**. You can use the deleteLater() function.

PR #2270, opened 2019-03-15
Backport 'Remove MDI interface option' to 4_12_x

ChristopherCotnoir:


gilmoskowitz:
CI shows a missed case in todoItem.cpp, which no longer exists in the 5_0_x branch.

PR #2272, opened 2019-03-18
[34302] Release Task number when cancelling the screen

andersdd:
requires xtuple/xtuple#3599

### DIFF
--- a/guiclient/task.cpp
+++ b/guiclient/task.cpp
@@ -520,6 +520,17 @@ void task::sClose()
     if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Deleting Task Information"),
                              query, __FILE__, __LINE__))
         return;
+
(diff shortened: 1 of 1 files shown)


guiclient/task.cpp
```diff
@@ -520,6 +520,17 @@ void task::sClose()
     if (ErrorReporter::error(QtCriticalMsg, this, tr("Error Deleting Task Information"),
                              query, __FILE__, __LINE__))
         return;
+
+    if (_metrics->value("TaskNumberGeneration") != "A" ||
+        _metrics->value("TaskNumberGeneration") != "O")
```
gilmoskowitz: These metric checks are either reversed or unnecessary. We want to release the task number if it was or could have been auto-generated ("A" or "O"). The conditional as written always evaluates to `true`.

PR #2273, opened 2019-03-18
Fix bugs found during demo

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3600.

PR #2277, opened 2019-03-21
[30314] Handle postgres version check separately for each client

ChristopherCotnoir:


ChristopherCotnoir:
Requires xtuple/updater#70, xtuple/connect#146.

PR #2280, opened 2019-03-22
Selection Widget initial commit

interhawk:
Initial commit of the selection widget. Gil has already suggested a rewrite of the sql generating portion in metasql for the next iteration.

### DIFF
--- a/widgets/selectionwidget.cpp
+++ b/widgets/selectionwidget.cpp
@@ -392,7 +392,7 @@ int SelectionWidget::sCommitChanges()
       return -1;
     }
   }
-  qry.exec("COMMIT;");
+  if (!_parentInTrans) { qry.exec("COMMIT;"); }
   _added.clear();
   _removed.clear();
   return 0;
--- a/widgets/selectionwidget.cpp
+++ b/widgets/selectionwidget.cpp
@@ -13,6 +13,86 @@
 
 #define DEBUG false
 
+/**
+* @class SelectionWidget
+
+* @brief Widget used to pick items from one list and add them to another.
+
+* The selection widget in corporates 2 XTreeWidgets, one as a source (or available)
+* and one as a destination (or selected). Each tree will be populated from a different query
+* but the data should be compatible with the fields on both trees otherwise moving items back
+* and forth doesn't make sense.
+*
+* The widget was designed to be adaptible and thus requires some configuration first.
+* The following must be done befor the widget will function properly:
+*
+* <UL>
+* <LI>Tree Setup:<BR>
+* The internal trees are exposed via the getAvail() and getSel() functions. Once you
+* have pointers to the trees, they must be configured as is usual for XTreeWidgets.
+*
+* <LI>Equality Columns:<BR>
+* Since each of the trees may present different atributes of the underlying data, there needs
+* to be a way to determine equality between items on different trees. A list of name-value 
+* pairs must be passed to the setEqualityColumns() function. Each pair will contain the XTreeWidget
+* column names to be considered equivalent. When the trees first load, any items in the "available" 
+* tree that are already present in the "selected" tree (based on the equality columns) will be
+* filtered out of the available list.
+*
+* <LI>Select and Cancel Buttons:<BR>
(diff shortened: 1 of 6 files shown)


widgets/selectionwidget.cpp
```diff
@@ -0,0 +1,292 @@
...
+  }
+  destTree->addTopLevelItem(pXtitem);
+}
+
+bool SelectionWidget::isSameItem(XTreeWidgetItem *pXtitem1, XTreeWidgetItem *pXtitem2) const
+{
+  int index = 0;
+  while (index < _equalityColumns.count())
+  {
+    if (pXtitem1->rawValue(_equalityColumns.name(index)) !=
```
ChristopherCotnoir: Should this be value instead of name?

interhawk: Looking at this with a fresh pair of eyes, it really needs a comment in the code. I am using a parameter list to create pairs of columns that should match up. Since the 2 trees may not have the same columns, the person setting up the widget needs to configure what "equality" means for purposes of filtering out things from the available list. So for example I would add a parameter like: {subscriptcustdtl_detail: cntcteml_email} and then check:
 pXitem1->rawValue(subscriptcustdtl_detail) against
pXitem2->rawValue(cntcteml_email)
this is why it uses .name for the first and .value for the second.

ChristopherCotnoir: OK. Not sure a ParameterList is the best option here as it differs pretty substantially from what name and value usually mean. A QPair would probably be clearer, but at least it functions.

widgets/selectionwidget.cpp
```diff
@@ -0,0 +1,292 @@
...
+  {
+    qDebug("Remove query string: %1", qryString);
+    QMapIterator<QString, QVariant> bindings(qry.boundValues());
+    while (bindings.hasNext())
+    {
+      bindings.next();
+      qDebug("%1:\t%2", bindings.key(), bindings.value().toString());
+    }
+  }
+  qry.exec();
```
gilmoskowitz: Consider trying [execBatch](https://doc.qt.io/qt-5/qsqlquery.html#execBatch). We've not used it before but it might be a good fit here. Alternatively try MetaSQL.

As above the method name is misleading — we're running a query, not just building it to be run later.

widgets/selectionwidget.cpp
```diff
@@ -0,0 +1,292 @@
...
+    }
+  }
+}
+
+int SelectionWidget::buildAddQuery()
+{
+  XSqlQuery qry;
+  XSqlQuery rollback;
+  rollback.prepare("ROLLBACK;");
+  qry.exec("BEGIN;");
```
gilmoskowitz: This is potentially dangerous. If the caller is already in a transaction then we might inadvertently roll back or commit that transaction prematurely.

Also the method name is misleading. We're running the query, not just building it.

widgets/selectionwidget.cpp
```diff
@@ -0,0 +1,292 @@
...
+    }
+  }
+  if ((!_modifyTableName.isEmpty()) && (!_removeByIdTableColName.isEmpty()) && (!_removed.isEmpty()))
+  {
+    if (buildRemoveQuery() == 0)
+    {
+      _removed.clear();
+    }
+  }
+}
```
gilmoskowitz: Ah... even worse... Not only are we creating a transaction in `buildAddQuery()` that might conflict with the world outside this widget, we're not even committing changes within the widget atomically.

widgets/selectionwidget.cpp
```diff
@@ -0,0 +1,421 @@
...
+  {
+    if (execRemoveQuery(outQry) == -1)
+    {
+      if (!_parentInTrans) { rollback.exec(); }
+      ErrorReporter::error(QtCriticalMsg, this, tr("Error Removing Selected items."),
+                           outQry, __FILE__, __LINE__);
+      return -1;
+    }
+  }
+  qry.exec("COMMIT;");
```
gilmoskowitz:
```suggestion
  if (! _parentInTrans) { qry.exec("COMMIT;"); }
```

PR #2292, opened 2019-03-31
[34761] Revert closing of job costed W/O to shipping transaction.  

andersdd:
Resolve COGS bug where a job costed item was not generating COGS entries due to not finding an appropriate item cost from the W/O.

requires xtuple/xtuple#3611


davidbeauchamp:
jenkins, retest this please

andersdd:
I thought it better to clarify the comment so devs know what is being done here rather than remove.

PR #2293, opened 2019-04-01
consolidate qt-client, openrpt and csvimp builds

mmicr001:
Added openrpt and csvimp as subdirectories to qt-clien


davidbeauchamp:
jenkins retest this please

PR #2306, opened 2019-04-08
Selection widget

interhawk:
Addresses remaining comments from code review.

### DIFF
--- a/widgets/selectionwidget.cpp
+++ b/widgets/selectionwidget.cpp
@@ -315,11 +315,12 @@ int SelectionWidget::execAddQuery(XSqlQuery &outQry)
   
   //build lists of values for each constraint
   QList<QVariantList> valueLists;
+  QVariant valToInsert;
   for (int i = 0; i < _addConstraints.count(); i++) { valueLists.append(QVariantList()); }
-  foreach(XTreeWidgetItem *xtitem, _added)
+  
+  for (int conIndex = 0; conIndex < _addConstraints.count(); conIndex++)
   {
-    QVariant valToInsert;
-    for (int conIndex = 0; conIndex < _addConstraints.count(); conIndex++)
+    foreach(XTreeWidgetItem *xtitem, _added)
     {
       valToInsert = xtitem->rawValue(_addConstraints.value(conIndex).toString());
       if (valToInsert.isNull())
@@ -328,11 +329,8 @@ int SelectionWidget::execAddQuery(XSqlQuery &outQry)
       }
       valueLists[conIndex].append(valToInsert);
     }
-  }
-  //bind value lists for batch execution
-  for (int i = 0; i < valueLists.count(); i++)
-  {
-    qry.bindValue(QString(":parameter%1").arg(i), valueLists[i]);
+    //bind value list for batch execution
+    qry.bindValue(QString(":parameter%1").arg(conIndex), valueLists[conIndex]);
   }
 
   if (DEBUG)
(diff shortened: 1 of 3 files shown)


widgets/selectionwidget.cpp
```diff
@@ -293,40 +311,60 @@ int SelectionWidget::execAddQuery(XSqlQuery &outQry)
...
-      qry.bindValue(QString(":parameter%1").arg(i), bindVal);
+      valueLists[conIndex].append(valToInsert);
     }
-    if (DEBUG)
+  }
+  //bind value lists for batch execution
+  for (int i = 0; i < valueLists.count(); i++)
+  {
+    qry.bindValue(QString(":parameter%1").arg(i), valueLists[i]);
+  }
```
ChristopherCotnoir: These loops would be easier to follow if the _addConstraints loop was the outer loop and the xitem loop was the inner loop. You could then combine this last loop with the previous section, building the full list for each parameter and binding them immediately. Something like:

for (int i = 0; i < _addConstraints.count(); i++)
{
  QList<QVariantList> values;
  foreach(XTreeWidgetItem *xtitem, _added)
  {
    valToInsert = xtitem->rawValue(_addConstraints.value(conIndex).toString());
    if (valToInsert.isNull())
    {
      bindVal = _addConstraints.value(i);
      valToInsert = _addConstraints.value(conIndex);
    }
    values.append(valToInsert)
  }
  qry.bindValue(QString(":parameter%1").arg(i), values);
}

widgets/selectionwidget.h
```diff
@@ -80,14 +87,14 @@ protected slots:
...
-  QList<XTreeWidgetItem*> _added;
-  QList<XTreeWidgetItem*> _removed;
-  QString _modifyTableName;
-  QString _removeByIdTableColName;
-  ParameterList _addConstraints;
-  ParameterList _equalityColumns;
-  bool _removeByAltId;
-  bool _parentInTrans;
+  QList<XTreeWidgetItem*> _added; /**<@brief List of items added to "available" tree since loading*/
+  QList<XTreeWidgetItem*> _removed; /**<@brief List of items removed from the "selected" tree sine loading*/
```
ChristopherCotnoir:
```suggestion
  QList<XTreeWidgetItem*> _removed; /**<@brief List of items removed from the "selected" tree since loading*/
```

widgets/selectionwidget.h
```diff
@@ -80,14 +87,14 @@ protected slots:
...
+  QString _modifyTableName; /**<@brief Name of table where items will be added or removed*/
+  QString _removeByIdTableColName; /**<@brief Column name used as a key by the delete statement to remove records*/
+  ParameterList _addConstraints; /**<@brief Name value pairs indicating column names and values to insert */
+  ParameterList _equalityColumns; /**<@brief Name Value pairs indicating pairs of columns to be compared for equality*/
+  bool _removeByAltId; /**<@brief When true, delete statements us the item's altId instead of id.*/
```
ChristopherCotnoir:
```suggestion
  bool _removeByAltId; /**<@brief When true, delete statements use the item's altId instead of id.*/
```

widgets/selectionwidget.cpp
ChristopherCotnoir: My bad, I forgot to change values to QList<QVariant> above. We don't need a list of lists with the loops in this order. The rest of what I had should be fine.

PR #2307, opened 2019-04-11
[34877] Don't allow saving of sales credits with no item or misc information

ChristopherCotnoir:


ChristopherCotnoir:
Associated with xtuple/xtuple#3621.

PR #2313, opened 2019-04-18
Exposes savestatus enum to scripting and enhances sSave signals

interhawk:


guiclient/salesOrderItem.cpp
```diff
@@ -994,6 +994,18 @@ void salesOrderItem::clear()
...
+void salesOrderItem::setSaveStatus(SaveStatus status, QString msg)
+{
+  _saveStatus = status;
+  _scriptErrorMsg = msg;
+}
+
+void salesOrderItem::showScriptError()
+{
+  QMessageBox failure(QMessageBox::Critical, tr("Script Error"), _scriptErrorMsg);
+  failure.exec();
```
ChristopherCotnoir: Can be simplified to QMessageBox::critical(this, tr("Script Error"), _scriptErrorMsg);

guiclient/salesOrderItem.cpp
```diff
@@ -5190,3 +5207,25 @@ void salesOrderItem::sHandleScheduleDate()
...
+
+//script exposure
+
+QScriptValue constructSalesOrderItem(QScriptContext *context, QScriptEngine  *engine)
+{
+#if QT_VERSION >= 0x050000
+  return engine->toScriptValue(new salesOrderItem());
+#else
+  Q_UNUSED(context); Q_UNUSED(engine); return QScriptValue();
+#endif
```
ChristopherCotnoir: We don't need this version check, we don't compile at all with Qt4 or earlier anymore.

PR #2317, opened 2019-04-24
[34342] backport initial fix and add more

gilmoskowitz:
The additional problems were found when using the NEXT and PREVIOUS
buttons on the Sales Order Item window through kit and enhanced pricing
sublines. Once the window had been put into view mode for the sublines,
the `_mode` variable would get set properly but the various fields
would not get re-enabled.

### DIFF
--- a/guiclient/salesOrderItem.cpp
+++ b/guiclient/salesOrderItem.cpp
@@ -37,6 +37,9 @@
 
 #define ISQUOTE(mode) (((mode) & 0x20) == 0x20)
 #define ISORDER(mode) (!ISQUOTE(mode))
+#define ISNEW(mode)   (((mode) & (~ 0x20)) == cNew)
+#define ISEDIT(mode)  (((mode) & (~ 0x20)) == cEdit)
+#define ISVIEW(mode)  (((mode) & (~ 0x20)) == cView)
 
 #define iDontUpdate   1
 #define iAskToUpdate  2
@@ -113,6 +116,8 @@ salesOrderItem::salesOrderItem(QWidget *parent, const char *name, Qt::WindowFlag
   _authLineNumberLit->hide();
 
   _charVars << -1 << -1 << -1 << 0 << -1 << omfgThis->dbDate();
+
+  _initialMode = -1;
 
   //  Configure some Widgets
   _item->setType(ItemLineEdit::cSold | ItemLineEdit::cActive);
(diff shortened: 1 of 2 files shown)


guiclient/salesOrderItem.cpp
```diff
@@ -114,6 +117,8 @@ salesOrderItem::salesOrderItem(QWidget *parent, const char *name, Qt::WindowFlag
 
   _charVars << -1 << -1 << -1 << 0 << -1 << omfgThis->dbDate();
 
+  _initialMode = -1;
```
ChristopherCotnoir: Not needed because of xtuple/qt-client#2318.

guiclient/salesOrderItem.cpp
```diff
@@ -609,62 +583,22 @@ enum SetResponse salesOrderItem::set(const ParameterList &pParams)
...
-    _promisedDate->setEnabled(!viewMode);
-    _qtyUOM->setEnabled(!viewMode);
-    _priceUOM->setEnabled(!viewMode);
-    _warranty->setEnabled(!viewMode);
-    _listPrices->setEnabled(!viewMode);
-    _altCosAccnt->setEnabled(!viewMode);
-    _altRevAccnt->setEnabled(!viewMode);
-
-    _subItemList->setVisible(!viewMode);
-    _save->setVisible(!viewMode);
```
ChristopherCotnoir: This is missing from handleFieldsOnModeChange.

guiclient/salesOrderItem.cpp
```diff
@@ -5208,6 +5143,69 @@ void salesOrderItem::sHandleScheduleDate()
...
+  }
+  else
+  {
+    _comments->setType(Comments::SalesOrderItem);
+  }
+
+  _warehouse->setEnabled(true);
+  _save->setEnabled(false);             // _save will be enabled when _item.isValid()
+
+  _comments->setReadOnly(pMode == cNewQuote);   // prior code really is this specific
```
ChristopherCotnoir: Add || ISVIEW(pMode)

guiclient/salesOrderItem.cpp
```diff
@@ -5208,6 +5143,69 @@ void salesOrderItem::sHandleScheduleDate()
...
+  if (ISQUOTE(pMode))
+  {
+    setWindowTitle(tr("Quote Item"));
+    _comments->setType(Comments::QuoteItem);
+    _tabs->removeTab(_tabs->indexOf(_costofsalesTab));
+
+    _cancel->hide();
+    _sub->hide();
+    _subItem->hide();
+    _subItemList->hide();
```
ChristopherCotnoir: Should also be hidden in cView mode.

guiclient/salesOrderItem.cpp
```diff
@@ -4386,6 +4319,8 @@ void salesOrderItem::populate()
       _warehouse->setEnabled(false);
     }
   }
+
+  _save->setEnabled(ISEDIT(_mode));
```
ChristopherCotnoir: This shouldn't be needed because _save should be hidden in view mode.

guiclient/salesOrderItem.cpp
gilmoskowitz: See the comment below about jumping buttons.

guiclient/salesOrderItem.cpp
gilmoskowitz: See the comment below about jumping buttons.

PR #2323, opened 2019-05-01
Exposes QInputDialog to scripting

interhawk:


scriptapi/qinputdialogproto.h
```diff
@@ -0,0 +1,90 @@
...
+void setupQInputDialogProto(QScriptEngine *engine);
+QScriptValue constructQInputDialog(QScriptContext *context, QScriptEngine *engine);
+
+class QInputDialogProto : public QObject, public QScriptable
+{
+  Q_OBJECT
+
+public:
+  QInputDialogProto(QObject *parent);
+  Q_INVOKABLE QString        cancelButtonText() const;
```
gilmoskowitz: See if the QInputDialog's [properties](https://doc.qt.io/qt-5/qinputdialog.html#properties) shine through instead of exposing the methods (at least for the getters). This should eliminate either 18 or 36 `Q_INVOKABLE`s and correspondingly simplify the .cpp.

scriptapi/qinputdialogproto.cpp
```diff
@@ -0,0 +1,829 @@
...
+  {
+    return
+    QInputDialog::getDouble(qscriptvalue_cast<QWidget*>(context->argument(0)),
+                            qscriptvalue_cast<QString>(context->argument(1)),
+                            qscriptvalue_cast<QString>(context->argument(2)),
+                            qscriptvalue_cast<double>(context->argument(3)),
+                            qscriptvalue_cast<double>(context->argument(4)),
+                            qscriptvalue_cast<double>(context->argument(5)),
+                            qscriptvalue_cast<int>(context->argument(6)),
+                            (bool*)context->argument(7).toQObject());
```
gilmoskowitz: Double-check that this properly modifies the value of the JS variable that was passed in. Same with `getInt`, `getItem`, `getMultiLineText`, and `getText` below.

scriptapi/qinputdialogproto.cpp
```diff
@@ -0,0 +1,829 @@
...
+    qScriptRegisterMetaType(engine, InputDialogOptionsToScriptValue,
+                            InputDialogOptionsFromScriptValue);
+    metaObject.setProperty("NoButtons",
+                           QScriptValue(engine, QInputDialog::NoButtons), ENUMPROPFLAGS);
+    metaObject.setProperty("UseListViewForComboBoxItems",
+                           QScriptValue(engine, QInputDialog::UseListViewForComboBoxItems),
+                           ENUMPROPFLAGS);
+    metaObject.setProperty("UsePlainTextEditForTextInput",
+                           QScriptValue(engine, QInputDialog::UsePlainTextEditForTextInput),
+                           ENUMPROPFLAGS);
```
gilmoskowitz: :-( I thought the call to `engine->newQMetaObject()` above would take care of these.

scriptapi/qinputdialogproto.h
gilmoskowitz: They don't :-( Having these properties seems to require adding even more code.

PR #2324, opened 2019-05-01
[33304] *Project not saved when releasing planned orders to purchase request

mmicr001:
The query for updating the `pr_prj_id` column was only present in the `cNew` code but not the `cRelease` code.
Also added an error check if createPr() does not return a valid value

### DIFF
--- a/guiclient/purchaseRequest.cpp
+++ b/guiclient/purchaseRequest.cpp
@@ -271,11 +271,10 @@ void purchaseRequest::sSave()
         purchaseCreate.bindValue(":pr_id",  _prid);
         purchaseCreate.bindValue(":prj_id", _project->id());
         purchaseCreate.exec();
-
-        purchaseCreate.prepare("SELECT releasePlannedOrder(:planord_id, false) AS result;");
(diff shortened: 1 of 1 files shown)


guiclient/purchaseRequest.cpp
```diff
@@ -265,9 +265,23 @@ void purchaseRequest::sSave()
...
+      if (_prid != -1)
+      {
+        purchaseCreate.prepare("UPDATE pr SET pr_prj_id=:prj_id WHERE (pr_id=:pr_id);");
+        purchaseCreate.bindValue(":pr_id",  _prid);
+        purchaseCreate.bindValue(":prj_id", _project->id());
+        purchaseCreate.exec();
+
+        purchaseCreate.prepare("SELECT releasePlannedOrder(:planord_id, false) AS result;");
+        purchaseCreate.bindValue(":planord_id", _planordid);
+        purchaseCreate.exec();
```
gilmoskowitz: The `releasePlannedOrder()` should be called regardless of the value of `_prid`. The reported bug is that the project association is missing, not that the planned order was released improperly.

PR #2332, opened 2019-05-09
[35024] Cleanup properly on reject()

ChristopherCotnoir:


guiclient/lotSerialRegistration.ui
```diff
@@ -628,7 +628,7 @@ to be bound by its terms.</comment>
    <sender>_buttonBox</sender>
    <signal>rejected()</signal>
    <receiver>lotSerialRegistration</receiver>
-   <slot>reject()</slot>
+   <slot>close()</slot>
```
gilmoskowitz: The [Qt docs](https://doc.qt.io/qt-5/qdialog.html) are unclear about what the result of `exec()` is if we call `close()` instead of `reject()`. This matters [when the caller makes decisions based on the result](https://github.com/xtuple/qt-client/blob/5_0_x/guiclient/contact.cpp#L1105).

PR #2345, opened 2019-05-20
[35104] Revert refresh behavior

ChristopherCotnoir:


ChristopherCotnoir:
Since the behavior is already unlike a dialog (it doesn't close on save), I figured it would be more clear to always use done() instead or accept() or reject(). Also hooked it up to closeEvent while I was at it so it does this when clicking the x.

Since we're making changes, instead of reverting the behavior, I changed the salesOrderItem to trigger updates in edit mode, not just new mode, instead of manually calling the refresh after the dialog is closed. This causes it to only be called when needed.

PR #2353, opened 2019-05-23
[35136] Modify itemsite and sales order item to honour Site shipping …

andersdd:
…setting

### DIFF
--- a/guiclient/itemSite.cpp
+++ b/guiclient/itemSite.cpp
@@ -938,10 +938,7 @@ void itemSite::sCheckItemsite()
   {
     _siteShipping = query.value("warehous_shipping").toBool();
     if (!_siteShipping)
-    {
       _sold->setChecked(_siteShipping);
-      _sold->setEnabled(_siteShipping);
-    }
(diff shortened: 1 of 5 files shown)


andersdd:
I can appreciate a company might have a Site where sales are made (a customer service office) that is separate from the warehouse(s), however it seems clear to me that the site referenced on a sales order item is the site/warehouse the order is to be shipped from.  That is _why_ you can specify the site on the S/O line.  
If I were to follow this logic though , I would revert the Item Site changes to allow selling from a non-Shipping Site, but we would then have to change the S/O line Site to a Shipping type.  And possibly the other widgets you have mentioned.  In that case, the S/O _header_ site should be changed from Shipping type to Sold type.  

Does that make more sense?

gilmoskowitz:
@andersdd Yes, putting the handling of sales vs. shipping sites in the sales order processing makes sense. It should then be possible to add a backwards-compatibility script in case someone dislikes the change.

andersdd:
Revised logic to default Item Sites based on shipping site setting, but do not prevent user override.  Altered S/O logic to allow "Sold" sites on the S/O header, but have "Shipping" sites only on the S/O to prevent the user trying to ship items from sites that are not allowed to ship from.

PR #2354, opened 2019-05-28
[35167] Fully clear PO Return screen after posting (RMA and list)

andersdd:


gilmoskowitz:
@andersdd Is there a bug number associated with this fix?

andersdd:
Issue raised and applied to PR

PR #2356, opened 2019-05-30
[30460] properly calculate WO start date whenever due date is changed

mmicr001:
when changing wo dates, the startDate would be set to `wo_startdate + (:dueDate-wo_duedate)` which could be a non work day or an exception. but we never check to see if the new startDate is actually a work date

### DIFF
--- a/guiclient/salesOrderItem.cpp
+++ b/guiclient/salesOrderItem.cpp
@@ -3082,11 +3082,11 @@ void salesOrderItem::sHandleSupplyOrder()
               applychange = true;
             if (applychange)
             {
-              ordq.prepare("SELECT changeWoDates(:wo_id, wo_startdate + (:dueDate-wo_duedate), :dueDate, true) AS result "
-                           "FROM wo "
-                           "WHERE (wo_id=:wo_id);");
+              ordq.prepare("SELECT changeWoDates(:wo_id, calculatenextworkingdate(:whs_id, :dueDate, :leadTime), :dueDate, true) AS result; ");
               ordq.bindValue(":wo_id", _supplyOrderId);
               ordq.bindValue(":dueDate", _scheduledDate->date());
+              ordq.bindValue(":leadTime", -1*_leadTime);
+              ordq.bindValue(":whs_id", _itemsiteLastWarehousid);
               ordq.exec();
               if (ordq.first())
               {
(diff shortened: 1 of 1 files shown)


guiclient/salesOrderItem.cpp
```diff
@@ -3082,11 +3082,11 @@ void salesOrderItem::sHandleSupplyOrder()
               applychange = true;
             if (applychange)
             {
-              ordq.prepare("SELECT changeWoDates(:wo_id, wo_startdate + (:dueDate-wo_duedate), :dueDate, true) AS result "
-                           "FROM wo "
-                           "WHERE (wo_id=:wo_id);");
+              ordq.prepare("SELECT changeWoDates(:wo_id, calculatenextworkingdate(:whs_id, :dueDate, :leadTime), :dueDate, true) AS result; ");
```
gilmoskowitz: `calculatenextworkingdate()` only exists if the inventory extension is installed. This won't work in a PostBooks database. The two options for handling this are MetaSQL and having two different versions of the query string.

```C++
// either
QString sql;
if (_metrics->boolean("UseSiteCalendar"))
  sql = "SELECT changeWoDates(:wo_id, calculatenextworkingdate(:whs_id, :dueDate, :leadTime), :dueDate, true) AS result;";
else
  sql = "SELECT changeWoDates(:wo_id, wo_startdate + (:dueDate - wo_duedate), :dueDate, true) AS result;";
// or
QString mql = "SELECT changeWoDates(<? value('wo_id') ?>,"
              "  <? if exists('useSiteCalendar') ?>calculatenextworkingdate(<? value('whs_id') ?>, <? value('dueDate') ?>, <? value('leadTime') ?>)"
              "  <? else ?> wo_startdate + (<? value('dueDate') ?> - wo_duedate)"
              "  <? endif ?>,"
              "  <? value('dueDate') ?>, true) AS result;";
```
followed by the appropriate binding calls (QSqlQuery::bindValue() for the first, ParameterList population for the second).

guiclient/salesOrderItem.cpp
```diff
@@ -3108,11 +3108,11 @@ void salesOrderItem::sHandleSupplyOrder()
           } // end scheduled date changed
           else if (_supplyOrderDueDate->date() != _supplyOrderDueDateCache)
           { // supply ord due date changed
-            ordq.prepare("SELECT changeWoDates(:wo_id, wo_startdate + (:dueDate-wo_duedate), :dueDate, true) AS result "
-                         "FROM wo "
-                         "WHERE (wo_id=:wo_id);");
+            ordq.prepare("SELECT changeWoDates(:wo_id, calculatenextworkingdate(:whs_id, :dueDate, :leadTime), :dueDate, true) AS result; ");
```
gilmoskowitz: ditto

PR #2362, opened 2019-06-04
[35189] prevent itemList crashes when called out of usual context

gilmoskowitz:


ChristopherCotnoir:
Should make this change in itemSearch::sFillList() as well to be safe.

PR #2378, opened 2019-06-06
[35034] parameter widget multiselect filters take up too much space

mmicr001:
After some [digging ](https://bugreports.qt.io/browse/QTBUG-68503), I found that in Qt versions=>5.11: 

> the minimum section size is properly checked

forcing the row size (after `tab->resizeRowsToContents();`) to be quite bigger than we would like.

Therefore, we must manually adjust the `MinimumSectionSize` as well as the height of every row as it is added to the table.


widgets/parameterwidget.cpp
```diff
@@ -1296,8 +1298,8 @@ void ParameterWidget::changeFilterObject(int index)
           qDebug() << objectName() << __FUNCTION__ << "Multiselect added"
                    << item->text() << item->data(Qt::UserRole) << "at" << i;
         tab->setItem(i, 0, item);
+        tab->setRowHeight(i,fm.capHeight()+4); //+4 adds a bit of padding
```
ChristopherCotnoir: height() would probably be more appropriate than capHeight(). From the description in the docs, it seems capHeight() does not take into account fonts extending below the baseline (ie characters like j).

PR #2386, opened 2019-06-11
Add Jenkinsfile, linux first

davidbeauchamp:


Jenkinsfile
```diff
@@ -0,0 +1,17 @@
+#!groovy
+
+pipeline {
+    agent none
+    stages {
+        stage('Build on Linux') {
+            agent { 
+                label 'xt-master-02'
```
gilmoskowitz: It would be better to use other properties of the desired agent, such as `linux pr 5_0_x qt5.12`. This would allow for testing multiple submissions in parallel and easier migration later.

davidbeauchamp: set it to the generic `docker` label so it can run on either of the master builders, the actual docker image is specified in the command.

PR #2390, opened 2019-06-11
make SelectionWidget constructor available to scripts

gilmoskowitz:


gilmoskowitz:
[Postel's Law](https://www.oreilly.com/library/view/software-architects-handbook/9781788624060/69fa0b38-9eb5-4640-8b17-c48f33d0a07a.xhtml)? I'm inclined to allow the extra args since JS is such a loose language. I could go either way.

interhawk:
Seems like a reasonable approach, definitely in the spirit of JS for better or for worse ;-) I'm good with the merge

PR #2392, opened 2019-06-14
[19253] * Omnibus : xTuple application crashes on selecting empty filter

mmicr001:
When the user selects an empty row from the combobox of a parameter widget filter value, we return from the `void ParameterWidget::storeFilterValue ` function before removing the value that was there before.
Therefore, when going from a selection in the combobox (i.e. 'Site') to an empty selection, the _filterValue for 'Site' still exists and was not removed. 
This would later cause the application to crash in the `filter()` function when the user would press `Query`


ChristopherCotnoir:
I don't think we should merge this or xtuple/qt-client#2398 for this release. Both of them are in volatile, messy, not well-understood parts of the code that are extremely prone to regressions and unexpected behavior, while on the flipside both bugs have been around for a long time and are either unlikely to be encountered (this one), or have little negative effect and are arguably not even a bug (2398). It seems relatively little gain while having huge potential for side effects in critical parts of the application.

PR #2398, opened 2019-06-20
[35194] disable item after user selects qty ordered for any item

mmicr001:
The discrepancy between sales orders and quotes was caused by a partial save being triggered only for sales order (and certain types of items) after selecting a quantity.
This change ensures that the itemCluster is set to `viewOnly`(by triggering a partial save) for all items in a sales order item or quote item.


guiclient/salesOrderItem.cpp
```diff
@@ -3317,7 +3317,6 @@ void salesOrderItem::sHandleSupplyOrder()
               // save the sales order item again to capture the supply order id
               _supplyOrderDropShipCache = _supplyDropShip->isChecked();
               sCalcUnitCost();
-              sSave(true);
```
ChristopherCotnoir: This save is needed to tie the new supply order id to the coitem record.

mmicr001: sSave is called at the end of sCalcUnitCost()
Though i would avoid calling the function twice

PR #2404, opened 2019-06-29
[35250] ‘Terms’ value is not displayed for Quote in ‘View’ mode

mmicr001:
The line being removed was introduced as part of the fix for [6912]
Does not seem to affect anything else aside from the disabled comboBox selection.


ChristopherCotnoir:
That line is there because it is supposed to change the combobox from type ARTerms, which only show AR Terms, to Terms, which shows all Terms, that way it will still be populated if Terms are changed to not be AR Terms any longer after using them on a sales order. The actual issue is that the XComboBox Terms query is broken. https://github.com/xtuple/qt-client/blob/5_0_x/widgets/xcombobox.cpp#L205-L207 If neither type AR or AP it should show all Terms, instead it shows none.

gilmoskowitz:
jenkins - retest this please

gilmoskowitz:
jenkins - retest this please

